PR #846, opened 2014-09-21
Replace password hashing with the new PHP Password Hashing API

rros:
Passwords are stored as Bcrypt-hashes instead of SHA-1-hashes. This is completely backwards-compatible, old password-hashes will be replaced with new hashes when the user logs in.

Read more about why Bcrypt is safer than (salted) SHA-1 here: http://codahale.com/how-to-safely-store-a-password/

PHP 5.5 Password Hashing API: http://php.net/manual/en/book.password.php (compatibility library: https://packagist.org/packages/ircmaxell/password-compat)


tcitworld:
We won't provide this in wallabag v1.x, but with v2 (build with SF2) we can implement this much more easier.

tcitworld:
Thanks for the work anyway, sorry we didn't consider it before.

j0k3r:
Hello @rros it could be a good idea for the v2.

Can you take a look at the [v2 branch](https://github.com/wallabag/wallabag/tree/v2) and see how you can replace the hashing method? I've added authentication and encoder classes in this commit https://github.com/wallabag/wallabag/commit/d91691573f108422cc2080462af35ebd62dc93fb

PR #859, opened 2014-10-07
Add STORAGE_PREFIX

kraoc:
Add STORAGE_PREFIX to enable multiple instances of Wallabag on shared hosting with only one database.


tcitworld:
We won't make that in v1.

tcitworld:
Thanks for proposing anyway, but with v2 it will be much easier to handle all this.

PR #960, opened 2014-12-01
fix bug on regex components

Camille31:
The trim and preg_match function were using $link[1] as argument, which, according to the previous regex, always contains a string with value "img". I changed it to $link[2] which actually contains the image url and I guess was the original author intent.
(same pull request as #959, but this one targets the right branch)


tcitworld:
I have no idea why, but pictures are not downloaded with your patch whereas there are correctly with the current dev branch.

nicosomb:
What can we do with this -so old- PR?

tcitworld:
Delete it. Too bad, but I don't have the time right now.

PR #986, opened 2015-01-02
Update fr_FR.utf8.po

goofy-bz:
formulation (trivial)


tcitworld:
The .mo file should have been updated too.

goofy-bz:
> The .mo file should have been updated too.
.po and .mo just suck :P — we translator want .properties! (but JSON will do)
http://www.lafermeduweb.net/billet/l20n-un-framework-de-traduction-avance-pour-le-web-par-mozilla-1791.html

tcitworld:
Ohhh. :D  Thanks for the link.

tcitworld:
And that went directly into master, @nicosomb ! I guess we're really tired. ^^

nicosomb:
:+1: thank you for the link ... And yes, we are tired :-D

PR #988, opened 2015-01-03
Extendemailshare

tcitworld:
Wait until icomoon is done, i mixed up a bit here.


nicosomb:
You created your branch from mine, right? There are my commits inside. 

This PR is to add an excerpt into the mail?

tcitworld:
Right, that's why we better wait until the other one is merged.

Yup, that's right. Do you want to make this an option ? It works on all mail clients I've tested (Thunderbird, Gmail, Geary)

nicosomb:
Nice feature :-)

I don't understand why we need the other PR, btw.

tcitworld:
Because I made too much merge mistakes and I'm afraid of making more. So I want to take my time. :D

nicosomb:
OK.

Issue #1000, opened 2015-01-14
RSS Enhancements


uytko:
A few suggestions:
- As stated in issue #553, I'd like a limitation (A number of articles or a date range).
- The pubDate of the item could be the date the article was added by user to Wallabag. I don't know if it's stored.

nicosomb:
Hello @uytko ! 

Thanks to @vpmalley, first suggestion is already done, see #1003.

tcitworld:
> The pubDate of the item could be the date the article was added by user to Wallabag. I don't know if it's stored.

Not stored, going to v2.

uytko:
(Very late) Thanks to both of you. I am able to find a workaround to my problem with the limitation.

nicosomb:
> The pubDate of the item could be the date the article was added by user to Wallabag. I don't know if it's stored.

Done by @j0k3r in v2.

Issue #553, opened 2014-03-13
Set number of articles provided in RSS files


tcitworld:
When the number of articles goes really big, the RSS files are becoming to take much time to fetch and also very relatively heavy.

A parameter like `?limit=xx` would be nice.

tcitworld:
Done.

PR #1003, opened 2015-01-16
[RSS] introducing query param 'limit' for max items in RSS feed

vpmalley:
As asked in issue #553 I introduced a 'limit' query parameter to control how many items should be displayed in the feed at most. It should also address a part of issue #1000.

So far the limit should be set manually, by default all items are displayed.


nicosomb:
Tested and approved. Nice! Thank you for your contribution, and welcome here!

Issue #1006, opened 2015-01-18
(Semantic-)Scuttle import?


kaffeeringe:
Hi all!

I am running Semantic Scuttle for my Bookmarks: http://bookmarks.kaffeeringe.de/ And as the project is dead, I am looking for an alternative. Wallabag seems to be more way I use Scuttle. But I wouldn't want to lose all my bookmarks. I tried to import them, but the format seems to be different and I wasn't able to find out, what the format should look like. 

I'd donate 50€ to Wallabag, if someone helps me with my migration.

Scuttle's export file looks like this:
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<posts update="2015-01-18T12:19:44Z" user="kaffeeringe">
	<post href="http://www.deutschlandfunk.de/militaertechnik-der-neue-drohnenkrieg.724.de.html?dram:article_id=308528" description="Militärtechnik - Der neue Drohnenkrieg, Teil 2" extended="Nahezu weltweit wird an &quot;smarten&quot; Kampfdrohnen getüftelt, die selbstständig ihr Ziel finden und vernichten sollen. An der Spitze dieses Wettrüstens stehen die USA. Die Bundesregierung beteuert hingegen stets, sie werde niemals eine Tötungsentscheidung irgendwelchen Automaten übertragen. Doch geforscht wird auch in Deutschland." hash="f1bbe0c2a6641e9af3cd51fc8ade5bb8" tag="krieg" time="2015-01-17T12:37:14Z"/>
	<post href="http://www.deutschlandfunk.de/militaertechnik-der-neue-drohnenkrieg.724.de.html?dram:article_id=308106" description="Militärtechnik - Der neue Drohnenkrieg, Teil 1" extended="Sie klären lautlos auf und töten bei Bedarf: Unbemannte militärische Flugzeuge sollen Soldaten schützen und zivile Opfer vermeiden. Vom sauberen Krieg sprechen Militärs, von Mordmaschinen die Gegner. Kampfdrohnen sind ein Milliardenmarkt - auch Deutschland interessiert sich für die Technologie." hash="eb10428b10df4df523ff3c8a5ecadde9" tag="krieg" time="2015-01-17T12:36:57Z"/>
[...]

PR #1008, opened 2015-01-18
Add Tags on Import

kaffeeringe:
I fixed it! And it works for me. 

But: I am not too good with programming. So better check the code ;-)


tcitworld:
I close since it's merged in dev (we work on dev branch).
I changed the `explode` from spaces to commas, since both wallabag and Pocket are using it, but I guess you can change the files produced with Semantic Scuttle by editing your own script at #1006.

Thanks a lot.

FYI, Pocket is the only Read-It-Later service providing tag export.

PR #1015, opened 2015-01-21
Savewithtags

tcitworld:
The add tags label will be replaced with an icon from Icomoon when all will be fixed with that (cf #964 #968)


tcitworld:
I'm kind of happy it works so easily.

PR #1016, opened 2015-01-21
URL encode 'via @wallabagapp' email body

shtrom:
Signed-off-by: Olivier Mehani <olivier.mehani@nicta.com.au>

PR #1018, opened 2015-01-21
URL encode 'via @wallabagapp' email body

shtrom:
#1016 on the dev branch (sorry for the duplicate, I'm clicking too fast...).

Signed-off-by: Olivier Mehani <olivier.mehani@nicta.com.au>


shtrom:
Oh no! It's actually a mistake. By doing this, the entire message is included in the URL for the message. I didn't realise this when checking the behaviour, as the mailto URL looked like what I expected.

This commit should be reverted.

I'm very sorry about that...

PR #1021, opened 2015-01-24
working on registration support

tcitworld:
Enable registration.
Work in progress, but please have a go.

WARNING : As this uses exceptions to the login system **it might bring security flaws**. Please double check after me.


tcitworld:
Needs (ok, really needs) CSS improvements, but it should be ok.

PR #1035, opened 2015-01-31
Tests are working again

j0k3r:
I've fixed travis & phpunit.
I also introduce `ant` which will be very helpful when the tests suites will became bigger and we will need to clear database, load fixtures, etc ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,18 @@
 language: php
+
 php:
+  - 5.4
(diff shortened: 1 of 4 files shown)


tcitworld:
Scrutinizer seems to be quite helpful.

nicosomb:
Yeah, back of @j0k3r !

Thank you.

PR #1036, opened 2015-01-31
Update composer

j0k3r:
_I'l rebase this branch when this PR #1035 will be merged_

It's better to use tagged version in composer (at least from my POV), it ease the download since it can download a zip file instead of cloning the project.

Do you really need the master of `simplepie/simplepie` ? Or the 1.3.1 branch is enough ?

You should tag your own libraries / fork:

 - wallabag/PHP-Flash-Messages
 - wallabag/kriss_php5
 - wallabag/pagination
 - wallabag/PHPePub
 - wallabag/php-readability
 - wallabag/phpMobi
 - wallabag/Fivefilters_Libraries

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+; top-most EditorConfig file
+root = true
+
+; Unix-style newlines
+[*]
+end_of_line = LF
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/bin/phpunit
+++ b/bin/phpunit
@@ -1,1 +1,1 @@
-../vendor/phpunit/phpunit/composer/bin/phpunit
+../vendor/phpunit/phpunit/phpunit
--- a/composer.json
+++ b/composer.json
@@ -68,11 +68,11 @@
         "friendsofsymfony/rest-bundle": "~1.4",
         "jms/serializer-bundle": "~0.13",
         "nelmio/api-doc-bundle": "~2.7",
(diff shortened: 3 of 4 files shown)


nicosomb:
About simplepie, 1.3.1 seems to be enough.
I plan to tag our libraries. Thank you for your advices !

j0k3r:
@nicosomb good to merge :)

nicosomb:
For my own libraries, I just need to create branch, like I did here: https://github.com/wallabag/Fivefilters_Libraries/ ?

tcitworld:
Tags is better than branches, right ?

j0k3r:
A tag is different than a branch :)
For your own librairies, you just need to create a tag on the HEAD of the master branch. It's really easy to do that using Github!

nicosomb:
I merge this PR and will open a new one with changes on our own libraries.

tcitworld:
It may have been better to keep the version number of the original versions, what do you think ?

j0k3r:
I agree with @tcitworld, could be better for every one if you keep the original version instead of tagging 1.0.0.

nicosomb:
:+1:

nicosomb:
But many of libraries don't have original version. That's why I tagged 1.0.0.

nicosomb:
I'm sure I don't do like I have to. What is the right process?

tcitworld:
> But many of libraries don't have original version. That's why I tagged 1.0.0.
Indeed. Let's leave it this way then (or just call master branch).

PR #1037, opened 2015-01-31
CS

j0k3r:
Since wallabag is using Symfony 2 it shoud also follow what is done in a Symfony2 project :)

`php-cs-fixer` is useful to solve all coding standard mistake done by developers.

For every updated code, it's a good practice to run it, simply `php-cs-fixer fix src/`


nicosomb:
Thanks for that!

PR #1044, opened 2015-02-03
v2: base API implementations

kevmeyer:
- changed `GET /api/entries` (as discussed)
- added `PATCH /api/entries/{entryID}`
- added `DELETE /api/entries/{entryID}`


nicosomb:
Hi @kevmeyer !

* There are tmp files ;-) 
* about delete method, I'll change our model to have a delete flag on each entry (a "deleted" entry will stay in database, and we'll have a purge option). I'm working on model. 
* I'm not fan of these parts `if ($isStarred == '0' || $isStarred == '1') {`... I think we can have a better way.

nicosomb:
Here is how I implemented the GET method as we discussed: https://github.com/wallabag/wallabag/commit/6e334aba689f891a229c8a57f908f10dad5b3759
star and archive status are no more necessary. 
By default, delete status is 0 if you don't set it. 

Here is the delete method: https://github.com/wallabag/wallabag/commit/42a9064620eb73eaa42928a22eeec86299d4a883

I'll work on others methods (like PATCH one) next.

tcitworld:
@nicosomb  tmp files have been deleted, right ?

nicosomb:
You're right

nicosomb:
I close this PR to rename branch.

PR #1050, opened 2015-02-05
Tags import + fix retrieving unfetched localized entries from database

erixtekila:
The filter was containing not the same text while localized.

I've added too tag support while importing.

### DIFF
--- a/inc/poche/Database.class.php
+++ b/inc/poche/Database.class.php
@@ -293,7 +293,7 @@ class Database {
             $sql_limit = "LIMIT ".$limit." OFFSET 0";
         }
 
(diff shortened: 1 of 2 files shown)


j0k3r:
Could you target the `dev` branch instead of the `master` ? Thanks.

tcitworld:
Tags are already imported in dev branch, if I remember correctly.

erixtekila:
Hm… how do I do that ? Should I checkout dev branch to commit in later ?

erixtekila:
@tcitworld then just fix the unfetched entires sql query which doesn't work on localized version.

j0k3r:
@erixtekila just re-create a PR and at the top, you will see the _target_ branch defined on the master, just change it to the dev branch

tcitworld:
@erixteckila I will take care of it if you don't want to, no worries. ;-)

erixtekila:
@tcitworld Did it ;)

PR #1053, opened 2015-02-05
Fix fetched entries when localized

erixtekila:
The stored message in DB used a not localized text.

### DIFF
--- a/inc/poche/Database.class.php
+++ b/inc/poche/Database.class.php
@@ -293,7 +293,7 @@ class Database {
             $sql_limit = "LIMIT ".$limit." OFFSET 0";
         }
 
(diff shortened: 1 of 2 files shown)


tcitworld:
The use of `$url->isCorrect()` is a great idea. I wonder why we haven't used before.
Thanks for that.

erixtekila:
Thanks. While inserting bookmarks from firefox for example, there could be some "about:something" URL which mess the whole import process. 
Anyhow, you'd also add a mean to bypass certain unfetchable articles that may block the whole process.

tcitworld:
You manage to import bookmarks from Firefox as JSON, you mean, right ?
I thought unfetchable article were ignored, I'll have to check that again.

WIll test you PR today.

erixtekila:
@tcitworld Either json or html.

tcitworld:
>  Will try on 2 if I get some time ?!

v2 is quite WIP, I don't think import has been implemented yet

PR #1054, opened 2015-02-05
we can now set a title to the article when we use POST /api/entries

nicosomb:


src/Wallabag/CoreBundle/Controller/WallabagRestController.php
```diff
@@ -80,17 +80,14 @@ public function getEntryAction(Entries $entry)
...
         //TODO gérer si on passe les tags
-        //TODO ne pas avoir du code comme ça qui doit se trouver dans le Repository
         $url = $request->request->get('url');
 
         $content = Extractor::extract($url);
         $entry = new Entries();
         $entry->setUserId(1);
         $entry->setUrl($url);
-        $entry->setTitle($content->getTitle());
+        $entry->setTitle(!is_null($request->request->get('title')) ? $request->request->get('title') : $content->getTitle());
```
j0k3r: Won't it be more simple like that:

```php
$entry->setTitle($request->request->get('title') ?: $content->getTitle());
```

PR #1061, opened 2015-02-06
Remove temporary entities

j0k3r:
We should avoid temporary file to be commited.

The `--no-backup` parameter avoid these tmp files to be generated

```
By default, the unmodified version of each entity is backed up and saved
 (e.g. Product.php~). To prevent this task from creating the backup file,
 pass the --no-backup option:

   php app/console doctrine:generate:entities Blog/Entity --no-backup
```


nicosomb:
Wow, I commited these files? So sorry ...

PR #1062, opened 2015-02-06
add a real relation between user and entry

nicosomb:


nicosomb:
Build failded because we've got a 500 with test database (schema is not created). 

As I saw with @j0k3r, we have to look for `ant` for executing creating scripts for test database.

j0k3r:
This isn't related to the database test:

> Argument 1 passed to Wallabag\CoreBundle\Entity\Entry::__construct() must be an instance of Wallabag\CoreBundle\Entity\User, none given, called in /Users/j0k/Sites/github/wallabag/src/Wallabag/CoreBundle/Controller/EntryController.php on line 22 and defined (500 Internal Server Error)

To run locally, you should do:

```
ant -f app/build.xml
phpunit -c app/
```

nicosomb:
I don't know why this build failed https://travis-ci.org/wallabag/wallabag/jobs/49892046

```
PHP Fatal error:  Call to a member function getId() on a non-object in /home/travis/build/wallabag/wallabag/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php on line 89
```

But this file has only 17 lines...

j0k3r:
Could you rebase your branch against the v2 one? And try again

j0k3r:
Well, I've checked it.

The problem is that there is no user created in your database. So the test to create a new entry fail because it can find't a user to associate to the new entry.
So, I checked how to add a new user, especially the `sha1` issue I didn't fixed :)

Regarding how Symfony security check the password for the user, it's different from what is done in wallabag v1. Symfony encode password like that 'algo(password{salt})' and wallabag v1 like that 'algo(password.username.salt)'. So, I need to update few security parameter to fix that.

I'll try to do that this evening.

nicosomb:
We also need to change queries in the repository, do add join between entry table and user one.

j0k3r:
I've fixed 

- the `sha1` issue
- missing `leftJoin`
- re-added login system
- update test according to login system

nicosomb:
Tested and approved. Good job. Thank you.

PR #1072, opened 2015-02-10
Add more tests to Entry controller + security

j0k3r:


nicosomb:
Nice :) Good tests. All is green ... and EntryController is covered :clap:

PR #1080, opened 2015-02-13
[WIP] 1st draft for new template

nicosomb:
Based on template found here http://www.blacktie.co/2014/07/dashgum-free-dashboard/

You can try this branch here: http://v2.wallabag.org/


j0k3r:
I think we should try to find a solution to manage assets, like gulp, via node or bower maybe instead of pushing all these js and css file to the repo.

tcitworld:
I think bower's the simplest system. Do we need more ?

j0k3r:
Yep. We then need to list all assets dependencies.
Then a short `gulpfile.js` to uglify and concat them :)

nicosomb:
Is that a problem for assets and bower / gulp if I use an existing template? 
I never used these tools.

tcitworld:
No, it shouldn't be a problem, we just need to register everything the theme uses. With this system, it will be much easier to share dependencies (assets, js) between themes.

nicosomb:
OK, I'll have a look. 
So Bower and Gulp.

j0k3r:
:+1:

nicosomb:
Moved to trash.

PR #1106, opened 2015-02-19
locale es_ES.utf8 files fully translated → Translation to Spanish updated

icaroperseo:
Please consider include the translations to Spanish (es_ES.utf8 files) which are fully translated into that language.

Regards!!!


tcitworld:
I you are motivated, I think there's even more strings to be translated. The only language having all the translations are en_US and fr_FR. Did you base yourself on one of those ?

icaroperseo:
No, I use the existing file. I'll look it later (thanks for the tip)!

icaroperseo:
Do you could close this PR and allow me to make a new, on this occasion using the *fr_FR.utf8* file (apparently it is the one that is more complete) as a basis?

Thanks in advance...

tcitworld:
If you need to, no problem.

icaroperseo:
Well, I finished finally! :smile: 

Regards!

**Note:** Apparently there are still missing strings to translate, which are not in the *fr_FR.utf8* file.

j0k3r:
@icaroperseo which strings are missing?

icaroperseo:
@j0k3r for example the *about* section...

PR #1115, opened 2015-02-21
Docker

jrottenberg:
Trying to get #220 resurrected ;-)

I've tried to follow the instructions from the Vagrantfile + as few tweaks as possible to get the application up and running.

It should be trivial form there to start populating official images of the application on the registry.

Also it is as small as possible (granted I've added all the nice to have php modules) :
```
REPOSITORY                  TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
wallabag                    latest              d0ec9aecc4b1        4 minutes ago       305.5 MB
```

Here is the build log
```
$ docker build  -t wallabag . && docker run -it -p 8080:80 wallabag
Sending build context to Docker daemon 36.93 MB
Sending build context to Docker daemon
Step 0 : FROM ubuntu:trusty
 ---> 2d24f826cb16
Step 1 : RUN apt-get -qq update &&         DEBIAN_FRONTEND=noninteractive apt-get install -y curl apache2 php5 php5-xdebug php5-sqlite php5-pgsql php5-mysql php5-gd php5-curl php5-tidy&&         apt-get clean -y &&         echo "ServerName localhost" >> /etc/apache2/apache2.conf &&         rm -rf /var/www/html
 ---> Running in 4117baf48a8f
Reading package lists...
Building dependency tree...
Reading state information...
The following extra packages will be installed:
  apache2-bin apache2-data ca-certificates fontconfig-config fonts-dejavu-core
  krb5-locales libapache2-mod-php5 libapr1 libaprutil1 libaprutil1-dbd-sqlite3
  libaprutil1-ldap libasn1-8-heimdal libcurl3 libedit2 libfontconfig1
  libfreetype6 libgd3 libgssapi-krb5-2 libgssapi3-heimdal libhcrypto4-heimdal
  libheimbase1-heimdal libheimntlm0-heimdal libhx509-5-heimdal libidn11
  libjbig0 libjpeg-turbo8 libjpeg8 libk5crypto3 libkeyutils1
  libkrb5-26-heimdal libkrb5-3 libkrb5support0 libldap-2.4-2 libmysqlclient18
  libpq5 libroken18-heimdal librtmp0 libsasl2-2 libsasl2-modules
  libsasl2-modules-db libtidy-0.99-0 libtiff5 libvpx1 libwind0-heimdal
  libx11-6 libx11-data libxau6 libxcb1 libxdmcp6 libxml2 libxpm4 lsof
  mysql-common openssl php5-cli php5-common php5-json php5-readline psmisc
[...]

PR #1120, opened 2015-02-22
Docker php ext

jrottenberg:
Follow up from #1115 

Using a more dockerized install of php instead of the ubuntu php packages.


nicosomb:
Hello, 

Why did you open 2 PR ? 
We work on dev branch before merging on master one: can you send your PR on dev branch please.? 

Thanks for your help with Docker!

PR #1132, opened 2015-02-23
V2 theme

j0k3r:
So, to sum up what I did:

 - remove all themes except *baggy* (from `app/Resources/views/` and `src/Wallabag/CoreBundle/Resources/views/themes/`)
 - install LiipTheme to handle many themes
  - and introduce a new service `wallabag_core.helper.detect_active_theme` which help the bundle to define which theme is active by looking a the Config entry.
  - config form is now a service, so I can inject the list of available theme that are also used by the bundle
 - simplify the way to make a template (just need to put a `layout.html.twig` in a new theme folder and but your own js / css, see the [baggy one](https://github.com/wallabag/wallabag/blob/495aecfe7417918c45d8ea32b1ee8903018449cd/src/Wallabag/CoreBundle/Resources/views/themes/baggy/layout.html.twig))

I will later write a quick help on to write your own theme. Maybe when we'll have chosen the official v2 theme :)


app/config/config.yml
```diff
@@ -104,3 +104,15 @@ fos_rest:
         default_format: json
 
 nelmio_api_doc: ~
+
+liip_theme:
+    load_controllers: false
+    themes:
+        - baggy
+        - courgette
```
nicosomb: you can really remove courgette one :)

j0k3r:
I've updated the description with what I did.

nicosomb:
Great job!

PR #1144, opened 2015-03-03
Simplify build system

j0k3r:
Instead of launching the tests suite using:

```
ant -f app/build.xml prepare && phpunit -c app
```

It will be more simple:
```
ant prepare && phpunit
```


nicosomb:
:+1:

PR #1165, opened 2015-03-28
Prefix table

### LINKED ISSUE #799: Prefix for database
Contrary to v1, we must give availability to wallabag users to create tables with prefix. 

This issue concerns only mysql and postgres storage.

j0k3r:
Will fix #799 

Of course, after merging this, you will need to clear your caches and drop then recreate your database schema.

I finally went with naming strategy solution from Doctrine instead of using a listener.
See http://doctrine-orm.readthedocs.org/en/latest/reference/namingstrategy.html

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -75,7 +75,10 @@ doctrine:
 
     orm:
         auto_generate_proxy_classes: "%kernel.debug%"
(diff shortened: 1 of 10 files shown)


nicosomb:
Need to try it, but this feature is nice ! Thank you.

PR #1166, opened 2015-03-28
Add RSS feeds

### LINKED ISSUE #1000: RSS Enhancements

j0k3r:
By default, items in each feeds are limited to 50. User can change this value in the config page. This will fix #1000.

I've re-added the token system (I put it in a `Tools/Utils` class), which can also be regenerated (todo: make an ajax call).
Url are like that: `http://wallabag.io/j0k3r/0jlmUxAVZqmc26P/unread.xml`

I've added a ParamConverter to check if the username and the token in the url are good.

I've also added an `@Assert\Range` on `itemPerPage` & `rssLimit` to be sure it's a number. So the validation works on the form too. I've put a kind of random number as max value: 100.000.

Regarding the `InstallCommand` tests, they are god damn too long to execute because it launch external commands (mostly related to doctrine).

So I've added a PHPUnit @group (`command-doctrine`) so that we can avoid launching them on a regular basis, like that:

```
phpunit --exclude-group command-doctrine
```

And I've update travis to run PHPUnit like that for the moment. Launching `phpunit` will still work, but it's longer. The downside is that the `InstallCommand` is less tested using `--exclude-group`

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/Entry/entries.xml.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/Entry/entries.xml.twig
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="utf-8"?>
+<rss version="2.0" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:media="http://search.yahoo.com/mrss/">
+    <channel>
+        <title>wallabag — {{type}} feed</title>
+        <link>{{ url('unread') }}</link>
+        <pubDate>{{ "now"|date('D, d M Y H:i:s') }}</pubDate>
+        <generator>wallabag</generator>
+        <description>wallabag {{type}} elements</description>
+
+        {% for entry in entries %}
+
+            <item>
+                <title><![CDATA[{{ entry.title }}]]></title>
+                <source url="{{ url('view', { 'id': entry.id }) }}">wallabag</source>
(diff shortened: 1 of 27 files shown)


src/Wallabag/CoreBundle/Resources/views/Entry/entries.xml.twig
```diff
@@ -0,0 +1,34 @@
...
+                <pubDate>{{ entry.createdAt|date('D, d M Y H:i:s') }}</pubDate>
+                <description>
+                    <![CDATA[
+                    {%- if entry.content| readingTime > 0 -%}
+                        {% trans %}estimated reading time :{% endtrans %} {{ entry.content| readingTime }} min
+                    {%- else -%}
+                        {% trans %}estimated reading time :{% endtrans %} &lt; 1 min
+                    {%- endif %}
+
+                    {{ entry.content -}}
```
nicosomb: `entry.content|raw` maybe?

j0k3r: To keep html content? Well I've encaplusated the content inside a CDATA to avoid problem.
But yes, I need to test with raw.

nicosomb: I tryied, and I think it doesn't work.

PR #1174, opened 2015-04-01
Docker

jrottenberg:
Docker support :
Run in sqlite mode by default
Support for mysql, at run time mount your config if you have one

See https://github.com/wallabag/wallabag/pull/1115 for full history.

Test with :
    
       docker build -t wallabag .
       docker run -it -p 8080:80 wallabag
       open http://docker_host:8080

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM    debian:jessie
+
+
+RUN     apt-get -qq update && \
+        DEBIAN_FRONTEND=noninteractive apt-get install -y curl apache2 php5 php5-xdebug php5-sqlite php5-pgsql php5-mysql php5-gd php5-curl php5-tidy&& \
+        apt-get clean -y && \
+        echo "ServerName localhost" >> /etc/apache2/apache2.conf && \
+        rm -rf /var/www/html
+
+
+WORKDIR /var/www/html
+
+# Making sure we run through the composer installation only if our requirements
+# have changed
+ADD     composer.* /var/www/html/
+
+RUN     curl -s http://getcomposer.org/installer | php && \
+        php composer.phar install
+
+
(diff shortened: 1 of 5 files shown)


nicosomb:
ping @bobmaerten

bobmaerten:
I have to admit I run wallabag on docker for my needs, and sqlite as a single user is quite enough.
There is [docker-wallabag](https://github.com/bobmaerten/docker-wallabag) running 1.9 and a new image should be made for 2.x. Docker has changed since then, docker-compose is the new way to go, and I think that should be included directly in wallabag like so many other oss-projects nowadays (discourse, gitlab, etc.)

jrottenberg:
Good point @bobmaerten , I've added docker-compose integration.

Not sure what standard you'd go with, but basically you define a mysql dbname/user and pass in the docker-compose file, and enter those values on the installer : 

<a href="http://imgur.com/51qdbOU"></a>


Of course you can stick to standalone + sqlite solution as well.

nicosomb:
@bobmaerten @jrottenberg do you think this PR is good to be merge?

bobmaerten:
Don't really know about the php/sql part, but a basic Dockerfile/docker-compose.yml is better than nothing. Maybe add something about credentials in a .md file, as I regulary have question about it on my docker-wallabag repo.

jrottenberg:
Oh did I miss something ? I can add a comment on a new markdown file if you need regarding credentials, as of now it was using simple value defined in docker-compose not too sure if we even needed that (could have use root + no password as well)

nicosomb:
You missed nothing, it's just because we won't work on 1.9.x anymore. You can help us to create a Dockerfile for wallabag v2 :-)

jrottenberg:
Deal !

PR #1180, opened 2015-04-15
changed email fields with the email type and added required parameter on...

tcitworld:
... some stuff

Just tiny changes.

### DIFF
--- a/src/Wallabag/CoreBundle/Form/Type/EntryType.php
+++ b/src/Wallabag/CoreBundle/Form/Type/EntryType.php
@@ -10,7 +10,7 @@ class EntryType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('url', 'url')
+            ->add('url', 'url', array('required' => true))
             ->add('save', 'submit')
         ;
     }
--- a/src/Wallabag/CoreBundle/Form/Type/ForgotPasswordType.php
+++ b/src/Wallabag/CoreBundle/Form/Type/ForgotPasswordType.php
@@ -20,6 +20,7 @@ class ForgotPasswordType extends AbstractType
     {
         $builder
             ->add('email', 'email', array(
+                'required' => true,
                 'constraints' => array(
                     new Constraints\Email(),
                     new Constraints\NotBlank(),
--- a/src/Wallabag/CoreBundle/Form/Type/NewUserType.php
+++ b/src/Wallabag/CoreBundle/Form/Type/NewUserType.php
@@ -11,7 +11,7 @@ class NewUserType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('username', 'text')
+            ->add('username', 'text', array('required' => true))
             ->add('password', 'password', array(
                 'constraints' => array(
                     new Constraints\Length(array(
@@ -21,7 +21,7 @@ class NewUserType extends AbstractType
                     new Constraints\NotBlank(),
                 ),
             ))
-            ->add('email', 'text')
+            ->add('email', 'email')
             ->add('save', 'submit')
         ;
     }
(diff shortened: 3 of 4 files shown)


src/Wallabag/CoreBundle/Form/Type/UserInformationType.php
```diff
@@ -10,8 +10,8 @@ class UserInformationType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('name', 'text')
-            ->add('email', 'text')
+            ->add('name', 'text', array('required' => true))
+            ->add('email', 'email')
```
tcitworld: Maybe email can also be required now. cc @j0k3r @nicosomb

nicosomb: You're right.

j0k3r:
As far as I can remember, I used `text` because the css doesn't handle properly the `email` type :expressionless:  (_yeah not a really good reason_)

src/Wallabag/CoreBundle/Form/Type/UserInformationType.php
j0k3r: The `required` isn't needed because, since the form is mapped to the User entity, it will apply annotation from the entity to the form. In that case, it will add these ones for the email:

```
 * @Assert\Email()
 * @Assert\NotBlank()
```

So, it will be required in some way :smiley: 

And about the name, why should it be required?

src/Wallabag/CoreBundle/Form/Type/ForgotPasswordType.php
```diff
@@ -20,6 +20,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
             ->add('email', 'email', array(
+                'required' => true,
```
j0k3r: We already declare constraints as `new Constraints\NotBlank(),` so this is not required to _extra_ add the `required` attribute.

src/Wallabag/CoreBundle/Form/Type/NewUserType.php
```diff
@@ -11,7 +11,7 @@ class NewUserType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('username', 'text')
+            ->add('username', 'text', array('required' => true))
```
j0k3r: Same as below, the entity already define username as `@Assert\NotBlank()`

src/Wallabag/CoreBundle/Form/Type/EntryType.php
```diff
@@ -10,7 +10,7 @@ class EntryType extends AbstractType
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
-            ->add('url', 'url')
+            ->add('url', 'url', array('required' => true))
```
j0k3r: And same here, the url field in the entity can not be blank `@Assert\NotBlank()`

j0k3r:
I was wondering if you got any problems while testing the v2 and then feel the need to add this requirement.
I'm just curious :smile:

src/Wallabag/CoreBundle/Form/Type/UserInformationType.php
tcitworld: > The required isn't needed because, since the form is mapped to the User entity, it will apply annotation from the entity to the form.

It's just more pratical to use with mobile keyboards when entering your email (@ shows up). Nothing to do with server checking. Right ?

tcitworld: > And about the name, why should it be required?

My bad.

tcitworld:
> I was wondering if you got any problems while testing the v2 and then feel the need to add this requirement.
I'm just curious :smile:

I just like my browser starting to tell me the informations I entered aren't correct before sending the page. Of course it doesn't replace server-side checking which stays mandatory but I appreciate it.

Issue #1203, opened 2015-05-25
Tags not being exported in Wallabag 1.9.0


th3bl4cksmith:
Moving from Framabag account to my server (both 1.9.0). Exported and imported JSON file but all tags were missing. Checked Framabag JSON, but nothing to indicate tags were exported. Added tags to local server instance, did an export and still found no tags. 

Seems like Wallabag doesn't currently export its own tags. Tried looking through docs and searching, but found nothing to confirm this.

vpmalley:
Tags were indeed not exported at all. There is a pull request (#1207) for this, I hope you will be able to use the feature soon.

tcitworld:
Fix merged, just a little duplication of tags problem due to #1209 to fix and it's good to go.

Issue #1005, opened 2015-01-17
Deleting article in archive return to "unread"


Alwaysin:
Hi,

Why are we redirected to UNREAD when deleting an article in ARCHIVE?
It is time-consuming when you want to delete several articles at once :(

tcitworld:
It's a bug, not intended at all.

PR #1207, opened 2015-06-02
issue #1203: fixing tag export

vpmalley:
Tags were not exported with articles as written in issue #1203. They now are and can be imported back. Please review my code.


vpmalley:
also fixing issue #1005 on wrong redirect when deleting an archived article

tcitworld:
Will try very soon, thanks a lot ! :)

tcitworld:
Export seems fine, but it showed another problem (#1209).

PR #1225, opened 2015-06-28
Send postMessage to opener before autoclose

freddyb:
This patch comes in handy when using another webapp that uses window.open to delegate adding to the wallabag site. I'm needing this for the upcoming Firefox OS app that I'm working on: https://github.com/freddyb/wallabag-fxos/

The patch apply cleanly on the dev branch as well as 1.9 (where I'm using it).

cc @tcitworld

### DIFF
--- a/themes/_global/js/autoClose.js
+++ b/themes/_global/js/autoClose.js
@@ -1,6 +1,26 @@
 $(document).ready(function() {
 	current_url = window.location.href
 	if (current_url.match("&closewin=true")) {
+		if (opener) {
+			var msgDiv = $("div.messages");
+			var msg = msgDiv.children("p").text();
+			var status = msgDiv.hasClass("success") ?
+				'success' : 'unknown';
+			/* TODO
+			 * I would also like to send the link back
+			 * but the first link e.g. $(".tool.link")[0]
+			 * may not be the right way to find it:
+			 * We also get a "success" when the item is
+			 * already stored. In this case the item
+			 * will be sorted to the top.
+			 * Are there other cases where we get a
+			 * success but the url is not on top?!
+			*/
+			var url = $(".tool.link")[0].href;
+			opener.postMessage({"wallabag-status": status,
+				"wallabag-msg": msg,
+				"wallabag-url": url }, "*");
+		}
 		window.close();
 	}
 });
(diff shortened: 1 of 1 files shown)


themes/_global/js/autoClose.js
```diff
@@ -1,6 +1,26 @@
 $(document).ready(function() {
 	current_url = window.location.href
 	if (current_url.match("&closewin=true")) {
+		if (opener) {
```
j0k3r: What is `opener` and where is it defined ?

tcitworld: I guess it's window.opener in fact.

tcitworld:
> I would also like to send the link back but the first link e.g. $(".tool.link")[0] may not be the right way to find it

Currently it's the only way I see. Maybe a better way would be to have the server send back this information somehow, for example in the success message.

> We also get a "success" when the item is already stored. In this case the item will be sorted to the top. Are there other cases where we get a success but the url is not on top?!

When a article is added, it should be [the only time](https://github.com/wallabag/wallabag/blob/master/inc/poche/Poche.class.php#L234).

themes/_global/js/autoClose.js
freddyb: Yeah, I can rewrite this to `window.opener` if you want

And if you allow me, I think I could make the surrounding JS a bit nicer:
```diff
-	current_url = window.location.href
-	if (current_url.match("&closewin=true")) {
+	if (location.search.match("&closewin=true")) {
```

freddyb:
@tcitworld: I will remove the comment and go on using `$(".tool.link")[0]` for the URL then. Thanks!

tcitworld:
Thanks to you. 
Btw, I got someone to show me the app running on Firefox OS yesterday, but the wifi connexion was bad and I couldn't complete adding a link from the app. Seems to do what it needs to do nicely though.

freddyb:
I have added another commit to address your review comments. Looking forward for additional feedback!

Once the review is granted, I will meld those commits into a single one using git rebase.

tcitworld:
Looks fine to me. I wonder if we could benefit from this feature in other apps (cc @kevmeyer @jlnostr)

tcitworld:
@freddyb, can you resend your PR on dev branch. I will merge it just after this.

PR #1227, opened 2015-07-03
Readremaining

tcitworld:
Add readremaining : https://github.com/Aerolab/readremaining.js

Needs some work on css to be fine.


nicosomb:
* When I display an article, the remaining time is wrong. For example, for "Framabag, un nouveau service libre et gratuit", I have 18 min. When I scroll, the time becomes right. 
* Is this time related to the estimated reading time displayed on the homepage?

tcitworld:
Will wait v2.

PR #1294, opened 2015-08-07
fix #1282: add title on wallabag logo

### LINKED ISSUE #1282: Some bugs with new Materiel theme
Hi,
I've just check some pages of wallabag with this new theme and I noticed several tiny things.
Maybe some of them are not related to Materiel theme. I use Firefox and a Mac.

## 1
On [Save new entry page](http://v2.wallabag.org/new), the Save button has no specific CSS styles. Should it be displayed as "Save" when editing tags?
wallabag_issue_1

## 2
Has you can see is this next screenshot, search form is activated but I don't see any cross on right.
Cursor is changed when hovering this area but there is no icon. My cache has been cleared with no effect.
wallabag_issue_2

## 3
If I click on the Hamburger menu icon, i get a strange display. By the way, this icon is displayed only after closing search form (See #1280)!
Next clicks are good I think: it hide/show menu on left. And I always get an overlay (or sort of) when main menu is also there.
wallabag_issue_3

## 4
In an article page, we can edit title but the title attribute as a wrong value: "Edit tags".

## 5
- I think it should be better to display current category as title in main area: Unread, Favorites, ...
- And a title could be added on wallabag's logo to know this links goes to unread articles.

nicosomb:


modos189:
I don't line it

nicosomb:
what?

modos189:
not clear that the logo - it is a list of unread articles

nicosomb:
still don't understand, sorry. What do you suggest to fix that

modos189:
Sorry, I did not understand correctly the change. Now understood. I like =)

Issue #393, opened 2014-01-12
Find articles with reading time = X


nicosomb:
it should be possible to sort articles by reading time: all articles with a reading time < 3 minutes, < 5 minutes, etc.

PR #1297, opened 2015-08-07
store estimated reading time in database (#393)

nicosomb:
see #393

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -9,6 +9,9 @@ use Wallabag\CoreBundle\Entity\Entry;
 use Wallabag\CoreBundle\Service\Extractor;
 use Wallabag\CoreBundle\Form\Type\NewEntryType;
 use Wallabag\CoreBundle\Form\Type\EditEntryType;
+use Wallabag\CoreBundle\Filter\EntryFilterType;
+use Pagerfanta\Adapter\DoctrineORMAdapter;
+use Pagerfanta\Pagerfanta;
 
 class EntryController extends Controller
 {
@@ -89,15 +92,31 @@ class EntryController extends Controller
     /**
      * Shows unread entries for current user.
      *
+     * @param Request $request
+     * @param int     $page
+     *
      * @Route("/unread/list/{page}", name="unread", defaults={"page" = "1"})
      *
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showUnreadAction($page)
+    public function showUnreadAction(Request $request, $page)
     {
-        $entries = $this->getDoctrine()
+        $form = $this->get('form.factory')->create(new EntryFilterType());
+
+        $filterBuilder = $this->getDoctrine()
             ->getRepository('WallabagCoreBundle:Entry')
             ->findUnreadByUser($this->getUser()->getId());
+
+        if ($request->query->has($form->getName())) {
+            // manually bind values from the request
+            $form->submit($request->query->get($form->getName()));
+
+            // build the query from the given form object
+            $this->get('lexik_form_filter.query_builder_updater')->addFilterConditions($form, $filterBuilder);
+        }
+
+        $pagerAdapter = new DoctrineORMAdapter($filterBuilder->getQuery());
+        $entries = new Pagerfanta($pagerAdapter);
 
         $entries->setMaxPerPage($this->getUser()->getConfig()->getItemsPerPage());
         $entries->setCurrentPage($page);
@@ -105,6 +124,7 @@ class EntryController extends Controller
         return $this->render(
             'WallabagCoreBundle:Entry:entries.html.twig',
             array(
+                'form'          => $form->createView(),
                 'entries'       => $entries,
                 'currentPage'   => $page
             )
@@ -114,15 +134,31 @@ class EntryController extends Controller
     /**
      * Shows read entries for current user.
      *
+     * @param Request $request
+     * @param int     $page
+     *
      * @Route("/archive/list/{page}", name="archive", defaults={"page" = "1"})
      *
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showArchiveAction($page)
+    public function showArchiveAction(Request $request, $page)
     {
-        $entries = $this->getDoctrine()
+        $form = $this->get('form.factory')->create(new EntryFilterType());
+
+        $filterBuilder = $this->getDoctrine()
             ->getRepository('WallabagCoreBundle:Entry')
             ->findArchiveByUser($this->getUser()->getId());
+
+        if ($request->query->has($form->getName())) {
+            // manually bind values from the request
+            $form->submit($request->query->get($form->getName()));
+
+            // build the query from the given form object
+            $this->get('lexik_form_filter.query_builder_updater')->addFilterConditions($form, $filterBuilder);
+        }
+
+        $pagerAdapter = new DoctrineORMAdapter($filterBuilder->getQuery());
+        $entries = new Pagerfanta($pagerAdapter);
 
         $entries->setMaxPerPage($this->getUser()->getConfig()->getItemsPerPage());
         $entries->setCurrentPage($page);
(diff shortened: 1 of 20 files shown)


tcitworld:
Any idea what's happening with Travis CI ?

tcitworld:
I wonder if we should try looking for another solution. This method to estimate reading time is kind of *really* bad with non-Latin characters and/or some languages.

nicosomb:
What do you suggest?

tcitworld:
> try looking for another solution

;-)

nicosomb:
@j0k3r I really don't understand why tests failed. Maybe I'm so tired. Go to bed. If you can help me ... ;-)

tcitworld:
Didn't find any real alternative, but it seems to me we could take in consideration pictures, with a time related to their size.

nicosomb:
Yeah, I found my error. Fixed.

nicosomb:
> but it seems to me we could take in consideration pictures

Seriously? It's an estimation of reading time, I don't think we should take in consideration pictures.

tcitworld:
Well, I thought it more like the time spent on the whole article. Weren't we considering counting the length of videos too ?

src/Wallabag/CoreBundle/Helper/Tools.php
```diff
@@ -118,4 +118,16 @@ public static function generateToken()
...
+    /**
+     * For a given text, we calculate reading time for an article.
+     *
+     * @param $text
+     *
+     * @return float
+     */
+    public static function getReadingTime($text)
+    {
+        return floor(str_word_count(strip_tags($text)) / 200);
```
j0k3r: WP plugin seems to use 250 words per minutes ([see here](https://plugins.trac.wordpress.org/browser/estimated-post-reading-time/trunk/plugin.php#L73)), maybe we can adjust it ?
Also, it could be interesting to explain the calculation, at least what is this number `200`.

j0k3r:
I don't this a lot of tests in this PR :)

tcitworld:
Wikipedia shows some data about the number of words reads per minute : https://en.m.wikipedia.org/wiki/Reading_(process)#Reading_rate

nicosomb:
@modos189 I'll add filters to filter articles by estimated reading view. I'll add a select list (with 3 items "< 5min", "between 5 and 10" and "> 10 min") and 2 text fields "min" and "max" to allow the user to filter with personal filters. 
If my explanations are clear, can you imagine how we can implement that on Unread / starred / archive screen?

Next, we'll have a new filter, by domain name.

modos189:
screen6

nicosomb:
I'm not sure that the filters icon is obvious for everyone.
And do you think popin is good for users? 
Do you think we can consider a right sidebar? 

Thank you for your reactivity.

tcitworld:
It's normal when using material design, but on an desktop view it's taking time to access.

tcitworld:
Also, maybe buttons instead of a list to choose from (one click instead of two).

nicosomb:
> It's normal when using material design
You talk about which point?

tcitworld:
The filter icon. You find it on any android app using material design.

nicosomb:
> I don't this a lot of tests in this PR :) 
This PR is still in progress, don't worry ;-)

modos189:
screen7
I do not want the icon to change

tcitworld:
I find this nice.

nicosomb:
:+1:

tcitworld:
And we can add more filter options in the future.

tcitworld:
As for more filters (I write it here, I know it's dirty), there should be the date when the entry was created and the type (MIME type in database).

nicosomb:
@j0k3r how do you think we can implement tests for this feature? Just by adding some articles (and setting estimating time by hand) and filling filter form?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -32,6 +36,7 @@ public function addEntryAction(Request $request)
 
             $entry->setTitle($content->getTitle());
             $entry->setContent($content->getBody());
+            $entry->setReadingTime(Tools::getReadingTime($content->getBody()));
```
j0k3r: Why don't you do that directly in the EntryEntity ?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -89,22 +94,39 @@ public function editEntryAction(Request $request, Entry $entry)
...
      */
-    public function showUnreadAction($page)
+    public function showUnreadAction(Request $request, $page)
     {
-        $entries = $this->getDoctrine()
+        $form = $this->get('form.factory')->create(new EntryFilterType());
+
+        $filterBuilder = $this->getDoctrine()
             ->getRepository('WallabagCoreBundle:Entry')
             ->findUnreadByUser($this->getUser()->getId());
```
j0k3r: Did you really need to call it `find`? `getFilterForUnreadByUser` doesn't work?

j0k3r:
Just defined some fixtures (and/or updated existing one) and they will have estimated time. Then create a test for a filter by filling the form and check number of results.

src/Wallabag/CoreBundle/Controller/EntryController.php
nicosomb: It must start by findBy or findOne...

src/Wallabag/CoreBundle/Controller/EntryController.php
nicosomb: fixed.

nicosomb:
I pushed my draft for test. I need help for this one, if you can have a look...

nicosomb:
Ready for review.

src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -17,16 +17,11 @@ class EntryRepository extends EntityRepository
...
     {
-        $qb = $this->createQueryBuilder('e')
+        return $this->createQueryBuilder('e')
             ->leftJoin('e.user', 'u')
             ->where('e.isArchived = false')
             ->andWhere('u.id =:userId')->setParameter('userId', $userId)
-            ->orderBy('e.id', 'desc')
-            ->getQuery();
-
-        $pagerAdapter = new DoctrineORMAdapter($qb);
```
j0k3r: Sine you removed lines where `DoctrineORMAdapter` is used, you can remove the `use` at the beginning of the file

src/Wallabag/CoreBundle/Helper/Tools.php
j0k3r: Could you just add a comment about the 200 number? Telling that it is referencing to _words per minute_

src/Wallabag/CoreBundle/Resources/views/Entry/entries.html.twig
```diff
@@ -21,11 +21,12 @@
     {% if entries is empty %}
         <div class="messages warning"><p>{% trans %}No articles found.{% endtrans %}</p></div>
     {% else %}
+        <div><form>{{ form_rest(form) }}<button class="btn waves-effect waves-light" type="submit" id="submit-filter" value="filter">Filter</button></form></div>
         {% for entry in entries %}
             <div id="entry-{{ entry.id|e }}" class="entrie">
```
j0k3r: `entrie`, really? It's not english at all, it should be entry.
Maybe it should be done in a different PR, but you are using this class in your test, some maybe you can fix it

nicosomb: it's baggy theme ... I don't know if I have to fix this theme here...

tcitworld: I started migration work on the baggy theme this afternoon, but you can fix this here, no worries.

src/Wallabag/CoreBundle/Repository/EntryRepository.php
nicosomb: still used at the end of the file.

nicosomb:
finally good to merge?

src/Wallabag/CoreBundle/Repository/EntryRepository.php
j0k3r: Woops, sorry

PR #1298, opened 2015-08-07
fix material

modos189:


nicosomb:
yeah, good job guy. I like the notifications :+1:

src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
```diff
@@ -18,6 +18,67 @@
...
+                </ul>
+            </div>
+            <form method="get" action="index.php">
+                <div class="input-field nav-panel-search" style="display: none">
+                    <input name="search" id="searchfield" type="search" required placeholder="{% trans %}Enter your search here{% endtrans %}">
+                    <label for="search"><i class="mdi-action-search"></i></label>
+                    <i class="mdi-navigation-close"></i>
+                </div>
+            </form>
+            <!--<form name="entry" method="post" action="{{ path('new_entry') }}">
```
nicosomb: "add a new link" form will be displayed in the header? Or is that dead code?

src/Wallabag/CoreBundle/Resources/views/themes/material/public/js/init.js
```diff
@@ -10,19 +10,26 @@ $(document).ready(function(){
     //    $(".nav-panel-buttom").hide(100);
     //    $(".nav-panel-add").show(100);
     //    $(".nav-panel-menu").hide(100);
+    //    $(".nav-panels .action").hide(100);
```
nicosomb: dead code?

nicosomb:
just 2 comments, we can merge after that.

modos189:
When you are able to add material I uncomment.

If I understand correctly, it is now possible to add articles from the page ONLY / Neuve, because you want to pass the token

nicosomb:
> If I understand correctly, it is now possible to add articles from the page ONLY http://site.com/new,

yes, you can add new articles from this page only. So you can implement form in the header I think.

modos189:
Please tell me how to do it

nicosomb:
OK, I'll do that after my current work.

j0k3r:
Also, could you rebase your branch against the v2 to fix conflicts?

modos189:
Up

nicosomb:
I merged your PR. About new link form, my PR #1307 is not yet finished, so I think it's better for you to work on a new PR.

Issue #1305, opened 2015-08-09
[v2] Maintain baggy for wallabag v2?


nicosomb:
For wallabag v2, we'll have material theme, thanks to @modos189. 
Today, baggy theme is finished in v2. 

@tcitworld @j0k3r and others too, do you think we have to work on this theme? 
I love it, but I think we'll loose time with it.

tcitworld:
I've got used to it and I think most users are used to it too.

What do you expect to take time ?

nicosomb:
We need to create all the twig files for v2.

tcitworld:
Sure, but I don't see much problems we could encounter. Will have a look.
Who worked on *baggy* in the first place ?

nicosomb:
It was @Courgette :-)

tcitworld:
Well, we'll always be welcomed for help. ;-)

j0k3r:
I think it's cool to have a new theme for the v2 but there are always people who don't really like change. So if we let them come back to the _old_ baggy theme for the v2, they will be really happy :)

So :+1:  to keep Baggy.

courgette:
Hi, personnaly, i don't have time to maintain baggy now. But in October, i can get a few time to participate to projects Open source, so i can create a new theme for the v2 or if baggy already here, redesign it.

tcitworld:
OK. Can I just ask you for help for JS or css stuff sometime ?

courgette:
Yeah, of course

nicosomb:
PR from @j0k3r #1401

nicosomb:
Merged ... So we can close?

PR #1307, opened 2015-08-10
[WIP] V2 material new link form header

nicosomb:
I know, tests are broken.
It's because tests are made on baggy theme. If we want to keep compatibility between some new features (like this PR), we really discuss about #1305 ... 
For example, if I remove /new page on material, what about baggy theme?


tcitworld:
We should do the things for *material* and then rewrite *baggy* nearly from scratch.

modos189:
What if use API for this?

PR #1316, opened 2015-08-11
fix #1310 add focus to search form

### LINKED ISSUE #1310: [v2] [material theme] Autofocus when clicking on search button
When clicking on the search button, the cursor should already be inside the field so that I can start typing.

nicosomb:


tcitworld:
Seems good.

PR #1318, opened 2015-08-12
Update Useragent to Win8.1 and FF36

ticoombs:
I recently tried to save a link from exploit-db.com who use Sucuri.net Cloud Firewall.

As part of their IDS they block all unknown/old useragents.


j0k3r:
Looks good to me :+1:

nicosomb:
@tcitworld good to merge?

tcitworld:
Yup. In the future, maybe this should be a constant.

PR #1319, opened 2015-08-12
add new post form in material header

nicosomb:
replaces #1307 (I had problems with merge, some files were deleted, I made so many mistakes I think)

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
@@ -58,7 +58,7 @@
             </div>
             <div class="input-field nav-panel-buttom">
                 <ul>
-                    <li class="bold"><a class="waves-effect" href="{{ path('new_entry') }}" id="nav-btn-add"><i class="mdi-content-add"></i></a></li>
+                    <li class="bold"><a class="waves-effect" href="{{ path('new') }}" id="nav-btn-add"><i class="mdi-content-add"></i></a></li>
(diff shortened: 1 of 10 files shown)


src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
```diff
@@ -58,7 +58,7 @@
             </div>
             <div class="input-field nav-panel-buttom">
                 <ul>
-                    <li class="bold"><a class="waves-effect" href="{{ path('new_entry') }}" id="nav-btn-add"><i class="mdi-content-add"></i></a></li>
+                    <li class="bold"><a class="waves-effect" href="javascript: void(null);" id="nav-btn-add"><i class="mdi-content-add"></i></a></li>
```
j0k3r: `href="#"` isn't enough ?

j0k3r: I think you can let the url inside the href `{{ path('new') }}`. In that case, if every thing goes well, user will have js fully load and will have the js form poped. In any other case, bad network connection, js error, etc ... it will still be able to add an entry because it will be redirected to the `new` form page.

nicosomb: I will try it and will change it if it's good.

nicosomb:
@j0k3r I changed the href (your method works well :+1: )

j0k3r:
You don't need to push force every time you udpate a PR :)

PR #1335, opened 2015-08-12
store Content-Type in database

nicosomb:
As you can see below, mimetype is stored, but sometimes, we've got charset. 

What can I do? 

mimetype

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/Entry.php
+++ b/src/Wallabag/CoreBundle/Entity/Entry.php
@@ -265,6 +265,8 @@ class Entry
     public function setContent($content)
     {
         $this->content = $content;
+        $headers = get_headers($this->url, 1);
+        $this->mimetype = $headers['Content-Type'];
         $this->readingTime = Tools::getReadingTime($content);
 
         return $this;


tcitworld:
Well, parse the mimetype entry and compare it to a list before it gets stored.

nicosomb:
wow, tests are green on my computer.

tcitworld:
It's also strange that it passes with hhvm.

nicosomb:
and tests fail on a part I didn"t change ...

j0k3r:
I don't think the entity is the right place to define the content type.
Also, using `get_header` PHP will perform an extra HTTP request to get header every time you will save the content.

I think this information should come from FullTextRss (or graby) which has already performed the request to get the content.

nicosomb:
OK. Do I need to open a new issue on graby project?

j0k3r:
Yeah do that, because graby doesn't return header at the moment, only `html`, `title`, `url` and `summary`.
And I think it may return only content type instead of all headers.

tcitworld:
Full Text Rss has the mime type information, yup (it is returned with `dc:format`)

nicosomb:
Close this PR?

tcitworld:
Yup. Follow on https://github.com/j0k3r/graby/issues/5

j0k3r:
Well, we'll still need to get this information from graby to save it later

PR #1337, opened 2015-08-12
fixes material

modos189:


modos189:
I replaced the {{form_rest (form)}} manual entry of each field, so the system does not assign values to them. As it can improve?

j0k3r:
You shouldn't do that.
Check the documentation http://symfony.com/fr/doc/current/reference/forms/twig_reference.html

modos189:
Up

tcitworld:
Saying « up » doesn't make things go faster.

modos189:
Hmm. I am not receiving notification of new commits in other pull-request by mail, only the comments, so I wrote it.
Ok, I will not)

nicosomb:
that's a lack on github, you're right...

nicosomb:
When I click on "new article" icon, and on the cross after, the filter icon is no more visible.

nicosomb:
When I change the size of my browser window, the filter appears but I didn't click on the icon

nicosomb:
Good job :+1: 

* When I click on "new article" icon, and on the cross after, the filter icon is no more visible.
* When I change the size of my browser window, the filter appears but I didn't click on the icon
* Add a cross at the top right of the filter bar
* Bug: can't click on pagination > 1
* display bug: when we load homepage, we see quickly the filter form 
* new entry form: when we click on the field, the history of the field is at the left of the screen, not below the field (see screenshot)

nicosomb:
Yeah, calendars :-)

little bugs:
* on mobile view, if I click on sandwich icon, I can't close the sidebar.
* on chromium, I can't display calendars.
* on firefox, I can display calendar. But I can't see buttons to close the calendar. 
* Is that possible to reduce the calendar size?

modos189:
>on mobile view, if I click on sandwich icon, I can't close the sidebar.
As it turned out, the framework materializecss not support simultaneous operation of two sidenav. I fixed a bug in the framework and send a separate PR

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
```diff
@@ -62,4 +61,50 @@
...
+
+                <div class="col s12">
+                    <label>{% trans %}Create at{% endtrans %}</label>
+                </div>
+                <div class="input-field col s6">
+                    {{ form_widget(form.createdAt.left_date, {'type': 'date', 'attr': {'class': 'datepicker', 'data-value': form.createdAt.left_date.vars.value} }) }}
+                    <label for="entry_filter_createdAt_left_date" class="active">{% trans %}from{% endtrans %}</label>
+                </div>
+                <div class="input-field col s6">
+                    {{ form_widget(form.createdAt.right_date, {'type': 'date', 'attr': {'class': 'datepicker', 'data-value': form.createdAt.left_date.vars.value} }) }}
```
nicosomb: `form.createdAt.left_date.vars.value` => `form.createdAt.right_date.vars.value`

nicosomb:
\o/

modos189:
=)

PR #1341, opened 2015-08-14
capistrano config

nicosomb:
Used for v2 deployment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,8 @@
 # Data for wallabag
 data/assets/*
 data/db/wallabag*.sqlite
+
+# Capistrano
+/config/deploy.rb
+/config/deploy/production.rb
+/config/deploy/staging.rb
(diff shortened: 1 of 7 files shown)


j0k3r:
You don't need other specific configuration? Just the default one is ok?

nicosomb:
I also have a config directory (with deploy.rb, etc.) ... but I forgot to add them in gitignore. Will do.

nicosomb:
Fixed .gitignore.

j0k3r:
And you still didn't add the config files :)

j0k3r:
So, with this config, you just do `cap deploy`, right?

nicosomb:
`bundle exec cap staging deploy`

j0k3r:
This should need some documentation for the v2 (ping #1359)

.gitignore
```diff
@@ -36,3 +36,8 @@
 # Data for wallabag
 data/assets/*
 data/db/wallabag*.sqlite
+
+# Capistrano
+/config/deploy.rb
+/config/deploy/production.rb
+/config/deploy/staging.rb
```
j0k3r: Is that necessary since you finally add them?

PR #1342, opened 2015-08-14
fix #981: add filter on domain name

### LINKED ISSUE #981: Search by same domain name

nicosomb:
I put a field text which looks for the domain names who contain the term (%term% in fact). I disabled the availibity to have a select list with these options: starts with, ends with, contains, is exactly. 
@j0k3r, @tcitworld Tell me if it's good for you.

After this PR is merged, @modos189 needs to enhance it with some JS and CSS :-)

### DIFF
--- a/src/Wallabag/CoreBundle/Filter/EntryFilterType.php
+++ b/src/Wallabag/CoreBundle/Filter/EntryFilterType.php
@@ -5,6 +5,8 @@ namespace Wallabag\CoreBundle\Filter;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
+use Lexik\Bundle\FormFilterBundle\Filter\FilterOperands;
+use Lexik\Bundle\FormFilterBundle\Filter\Query\QueryInterface;
 
 class EntryFilterType extends AbstractType
 {
@@ -24,7 +26,18 @@ class EntryFilterType extends AbstractType
                             'placeholder' => 'dd/mm/yyyy'),
                         'format' => 'dd/MM/yyyy',
                         'widget' => 'single_text'
-                    )));
+                )))
+            ->add('domainName', 'filter_text', array(
+                'apply_filter' => function (QueryInterface $filterQuery, $field, $values)
+                    {
+                        $value = $values['value'];
+                        if (strlen($value) <= 3 || empty($value)) {
+                            return null;
+                        }
+                        $expression = $filterQuery->getExpr()->like($field, $filterQuery->getExpr()->literal('%'.$value.'%'));
+                        return $filterQuery->createCondition($expression);
+                    }
+            ));
     }
 
     public function getName()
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Filter/EntryFilterType.php
```diff
@@ -5,12 +5,15 @@
...
+use Lexik\Bundle\FormFilterBundle\Filter\FilterOperands;
 
 class EntryFilterType extends AbstractType
 {
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('readingTime', 'filter_number_range');
+        $builder
+            ->add('readingTime', 'filter_number_range')
+            ->add('domainName', 'filter_text', array('condition_pattern' => FilterOperands::STRING_BOTH));
```
j0k3r: Could you add a validator to avoid people to search for `a` and kill the database ?I guess a minimum value of 3 characters should be ok.

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -266,7 +266,7 @@ public function setContent($content)
     {
         $this->content = $content;
         $this->readingTime = Tools::getReadingTime($content);
-        $this->domainName = parse_url($this->url, PHP_URL_HOST);
+        $this->domainName = Tools::getHostname(parse_url($this->url, PHP_URL_HOST));
```
j0k3r: Regarding how you will search in the database, I don't think you need to remove the subdomain or the `www.`. Maybe you just need to remove the _too common_ `www.` but some subdomain are important. Like for Tumblr, you won't search for `tumblr` but using the subdomain of the tumblr, which is more accurate.

nicosomb:
ready to review / ready to squash / ready to merge

src/Wallabag/CoreBundle/Filter/EntryFilterType.php
```diff
@@ -24,7 +26,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                             'placeholder' => 'dd/mm/yyyy'),
                         'format' => 'dd/MM/yyyy',
                         'widget' => 'single_text'
-                    )));
+                )))
+            ->add('domainName', 'filter_text', array(
+                'apply_filter' => function (QueryInterface $filterQuery, $field, $values)
+                    {
+                        $value = $values['value'];
+                        if (strlen($value) <= 3 || empty($value)) {
```
j0k3r: I don't really like this kind of _homemade_ validation but it doesn't seems to be more simplier

nicosomb: You're right ... but with LexikFormFilter, I don't believe we can make in another way.

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
```diff
@@ -77,8 +77,16 @@
                     <label for="entry_filter_readingTime_right_number">{% trans %}to{% endtrans %}</label>
                 </div>
 
+
                 <div class="col s12">
+
                     <label>{% trans %}Create at{% endtrans %}</label>
```
j0k3r: Where is the filter on created at ?

nicosomb: https://github.com/wallabag/wallabag/pull/1363

j0k3r:
Could you squash your commits? :)

nicosomb:
Done.

PR #1343, opened 2015-08-14
[WIP] filter by creation date

nicosomb:
TODO:
* ~~test~~
* if one field is not filled with the good format => error (in fact, it's the same thing for the reading time filter)
* ~~smart calendar (hi @modos189)~~ done in an other PR.

PR #1353, opened 2015-08-17
check if one entry or not

nicosomb:


modos189:
Filters by create date are delete =(

nicosomb:
oh yes, because we have to merge #1343 before.

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
```diff
@@ -17,7 +17,7 @@
     {% block pager %}
     {% if entries is not empty %}
         <div class="results clearfix">
-            <div class="nb-results left">{{ entries.count }} {% trans %}entries{% endtrans %}</div>
+            <div class="nb-results left">{{ entries.count }} {% trans %}entrie{% endtrans %}{% if entries.nbPages > 1 %}s{% endif %}</div>
```
j0k3r: Use [`transchoice `](http://symfony.com/doc/current/book/translation.html#twig-templates) instead

nicosomb: wow <3

PR #1363, opened 2015-08-18
add creation date filter view

nicosomb:


j0k3r:
What about testing that filter actually works? :)

nicosomb:
This one seems good to be reviewed too.

PR #1366, opened 2015-08-18
fix tag view in material theme

nicosomb:


j0k3r:
We need test on different themes .. so this kind of error shouldn't appear.

nicosomb:
test on different themes ... phpunit tests?

j0k3r:
I mean, we don't change theme in tests.
We should have test that cover all different theme to avoid this kind of problem.

I mean, run the test suite on baggy theme, then on material theme.

I said that, but I've no idea how to achieve that. I think we should define some ENV variable that contain the current theme and launch test using a different ENV each time.

I'll create an issue about that.

PR #1367, opened 2015-08-18
action labels depend on status

nicosomb:
I don't know if it's the good way (about the translation for example).


j0k3r:
Good catch !

PR #1369, opened 2015-08-18
fix #1332: bottom bar to display message

### LINKED ISSUE #1332: Add a bar on each page with alpha message
It's not a priority. But I want to have a bottom bar with a message like "be careful, it's an alpha version with bugs, etc. please open a new issue etc."

nicosomb:
Just a first version, no need to have a beautiful bar.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/base.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/base.html.twig
@@ -113,6 +113,10 @@
                 <p>{% trans %}powered by{% endtrans %} <a href="http://wallabag.org">wallabag</a></p>
             </footer>
         {% endblock %}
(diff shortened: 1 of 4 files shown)


src/Wallabag/CoreBundle/Resources/views/base.html.twig
```diff
@@ -113,5 +113,6 @@
                 <p>{% trans %}powered by{% endtrans %} <a href="http://wallabag.org">wallabag</a></p>
             </footer>
         {% endblock %}
+        <div id="alphabar">You're trying wallabag v2, which is in alpha version. If you find a bug, please have a look to <a href="https://github.com/wallabag/wallabag/issues">our issues list</a> and <a href="https://github.com/wallabag/wallabag/issues/new">open a new if necessary</a>.</div>
```
j0k3r: Hum I don't really like this hard coded way, can we defined that in the parameters.yml file? So when there is something in to display, we just have to fill the parameter and the alpha message will be displayed

nicosomb:
Fixed.

PR #1370, opened 2015-08-18
fix 1356: diet for composer

nicosomb:
I tried locally (composer clear cache, rm vendor -rf, composer install): no more "cloning ...". Only downloading.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,27 +29,7 @@
     "repositories": [
         {
             "type": "vcs",
-            "url": "https://github.com/wallabag/PHP-Flash-Messages"
-        },
-        {
-            "type": "vcs",
-            "url": "https://github.com/wallabag/kriss_php5"
-        },
-        {
-            "type": "vcs",
-            "url": "https://github.com/wallabag/pagination"
-        },
-        {
-            "type": "vcs",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -69,7 +49,6 @@
         "sensio/framework-extra-bundle": "~3.0",
         "incenteev/composer-parameter-handler": "~2.0",
         "nelmio/cors-bundle": "~1.4.0",
-        "umpirsky/twig-gettext-extractor": "~1.1.5",
```
nicosomb: already embedded with an other dependancy

composer.json
```diff
@@ -29,30 +29,10 @@
...
-            "type": "vcs",
-            "url": "https://github.com/wallabag/pagination"
-        },
-        {
-            "type": "vcs",
-            "url": "https://github.com/wallabag/PHPePub"
-        },
-        {
-            "type": "vcs",
             "url": "https://github.com/wallabag/php-readability"
```
j0k3r: This one is still required ?

nicosomb: yes, we still use Readability class.

j0k3r: Oh sorry I was thinking that removing all lines and just add tag to require will be enough, my bad.

PR #1372, opened 2015-08-19
assign tags to an entry

nicosomb:
TODO 
- [x] if tag already exists, don't create it
- [ ] ~~split tags "tag1, tag2" to create 2 tags~~ see #1376
- [x] update the unit test

### DIFF
--- a/src/Wallabag/CoreBundle/Repository/TagRepository.php
+++ b/src/Wallabag/CoreBundle/Repository/TagRepository.php
@@ -24,4 +24,21 @@ class TagRepository extends EntityRepository
 
         return new Pagerfanta($pagerAdapter);
     }
+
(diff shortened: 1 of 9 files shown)


src/Wallabag/CoreBundle/Repository/TagRepository.php
```diff
@@ -24,4 +24,21 @@ public function findTags($userId)
...
+
+    /**
+     * Find a tag by its label and its owner
+     *
+     * @param string $label
+     * @param int    $userId
+     *
+     * @return Tag|null
+     */
+    public function findOneByLabelAndUser($label, $userId)
```
j0k3r: Better to `findOneByLabelAndUserId`

PR #1375, opened 2015-08-19
fix #1368 config: redirect on correct tab in material theme

### LINKED ISSUE #1368: [v2] back to good section while updating configuration
Today, when we update user information, we are redirected to first tab, which can be frustrating because we don't see error message.

nicosomb:
It fixes the problem. 
But the anchor is between the tabs and the form. So the page scrolls down when we validate the form. But it fixes the problem.

(I know, the branch name is wrong)


j0k3r:
This does the job!

PR #1383, opened 2015-08-20
CS

j0k3r:
We shouldn't forget to run `php-cs-fixer fix src/` time to time


nicosomb:
We can add it in travis job.

PR #1384, opened 2015-08-20
fix #1371 config screen: display bug in RSS tab

### LINKED ISSUE #1371: Display issues in config screen
There is a strange display in configuration view: 
- "Rss token VS. token string
- "Rss links" VS. "Unread"

capture d ecran 2015-08-19 a 01 34 47

Moreover "Rss" should be written in caps.
And why not using "favourites" instead of "fav"?
Why mandatory fields are not indicated with a specific character (i.e. *)?

Tested on v2.wallabag.org with Firefox.

nicosomb:
It's not the better display (because I use h5 ..) but it's better than before.


j0k3r:
Shouldn't we try to fix that by using some CSS instead of using `h5`?

nicosomb:
I was in my train, and h5 was the easiest solution.

j0k3r:
So now, since you aren't in the train, you can find a better solution :)

nicosomb:
@modos189 If you can help me here, I prefer let you do it.

modos189:
I can not offer anything but to use css

j0k3r:
Fixed:

image

modos189:
I like)

PR #1385, opened 2015-08-20
filters: implement status filter and a new view (to display all entries)

### LINKED ISSUE #1382: [v2] Filter on status entry (unread/archive/starred)
see #1380. 
We need to filter on status entry.

nicosomb:
fix #1382

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -3466,16 +3466,16 @@
         },
         {
             "name": "phpunit/phpunit",
-            "version": "4.8.5",
+            "version": "4.8.4",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sebastianbergmann/phpunit.git",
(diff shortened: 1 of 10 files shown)


composer.lock
```diff
@@ -3466,16 +3466,16 @@
         },
         {
             "name": "phpunit/phpunit",
-            "version": "4.8.4",
+            "version": "4.8.5",
```
j0k3r: Try to avoid committing composer.lock if you didn't touch the composer.json :)

nicosomb: Do you want I remove composer.lock?

j0k3r: Yep

j0k3r:
Could you rebase your branch against the v2 branch and use the `showEntries` instead ?

nicosomb:
Done this morning, still in my train, but not pushed after ;-)

PR #1392, opened 2015-08-22
Fix redirect after deletion

### LINKED ISSUE #1391: [v2] Redirection expected after deleting an entry currently displayed
Hi,

On http://v2.wallabag.org, if we display an article and click on Delete link, this entry is successfully deleted but we get a 404.
I think users should be redirected to homepage.

j0k3r:
Fix #1391

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -292,16 +292,15 @@ class EntryController extends Controller
     }
 
     /**
-     * Deletes entry.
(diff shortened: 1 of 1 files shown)


nicosomb:
Is that possible to redirect to the last status view? 
For example, I'm on Starred view, I click on an article, I delete it, can wallabag redirect me to the starred view?

j0k3r:
It will require to store the _status view_ in the session and then, if it exists redirected to the right status page.
What was the behavior in the 1.x version?

nicosomb:
We redirect to homepage.

Issue #972, opened 2014-12-18
Preview picture for each article


nicosomb:
As @moafred (from Bookeen team) told me yesterday, we can use og:image meta to get a preview picture for each article. 
This tag is used by twitter and facebook to generate preview in posts. 

For example, with this article http://www.lemonde.fr/pixels/article/2014/12/18/affaire-sony-pictures-la-strategie-anti-piratage-d-hollywood-revelee_4542100_4408996.html, we have this meta :

```
<meta property="og:image" content="http://s1.lemde.fr/image/2014/12/17/600x315/4542146_3_8afb_le-legendaire-signe-surplombant-hollywood-en_443f1e6556ca1af14830f9d266203bcf.jpg" />
```

It can be useful for generating nice ePUB files (instead of wallabag logo). 

@tcitworld Are you interested to have a look to this feature?

tcitworld:
I was looking if FTRSS already has something like that, but it doesn't seem so (I only looked the 3.2 version). The picture they produce comes only from pure RSS feeds (`<itunes:image>`, `<atom:logo>`, `<atom:icon>`, `<image><title>` or `<image><dc:title>) via Simple Pie, not regular html pages.
I could build the detection right in the makefulltextfeed.php file, or in wallabag itself. Are you interested, @Fivefilters ?

tcitworld:
Wondering if it's better to extend the Readability class or the ContentExtrator one.

tcitworld:
Thanks a lot for the link !
I'll start the work with the ContentExtractor class.

nicosomb:
for your information, @fivefilters implemented this feature in last version http://blog.fivefilters.org/post/121443274427/full-text-rss-35

j0k3r:
This mean I need to add this feature to graby :)

j0k3r:
Picture is now attached to each entry. We need to display it now.

nicosomb:
what do you think of this?

![](https://lut.im/5FDiZ4PY/Ym3XuecZ)

j0k3r:
Great !

PR #1397, opened 2015-08-24
Integrate graby

### LINKED ISSUE #1262: Implement graby into wallabag v2
I think @j0k3r knows what he has to do with https://github.com/j0k3r/graby. Good luck! 

graby will replace Full-Text RSS for parsing articles.

j0k3r:
This should fix #1262.
It stills needs to fix how graby will read site_config directory, this is why tests are failing atm.

I also implement #972. It just save the Open Graph image in the database. I didn't display it.

 - [x] Fix graby-site-config
 - [x] Store image preview
 - [x] Grab open graph information when content failed to be fetched (like title, summary, etc ...)


j0k3r:
Graby returns the _final_ url, like if someone bag a `http://t.co` it will return the url the final url. Should we override this irk in wallabag or keep the original submitted url?

nicosomb:
Oh, nice! We have to avoid URL like t.co I think, so wallabag should store the final URL.

j0k3r:
Job done.
I've introduced a class called `ContentProxy` (_yeah it's not the best class name, feel free to share a good one if you have_) that take care of getting the content from an url and update the entry (using graby). So the way we grab content is in one place (instead of in the Core and in the Api).

graby will surely evolve in the future (1.0.0 isn't ready yet) but it'll be as a dependency, so it's not a problem.

nicosomb:
Nice job. The title still contains "WIP", is that still true?

j0k3r:
Nope :-)

PR #1403, opened 2015-08-25
Use full size of the panel for domain-filtering

tcitworld:
For longer domains, use all the space.

domain-filter-fullsize


nicosomb:
:+1:

PR #1418, opened 2015-09-12
filter for entries with previewPicture

nicosomb:


tcitworld:
I didn't find this filter much useful, at first, but then I thought you could remember the picture instead of the title.

src/Wallabag/CoreBundle/Filter/EntryFilterType.php
```diff
@@ -39,10 +39,23 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
                         return $filterQuery->createCondition($expression);
-                    },
+                },
             ))
             ->add('isArchived', 'filter_checkbox')
-            ->add('isStarred', 'filter_checkbox');
+            ->add('isStarred', 'filter_checkbox')
+            ->add('previewPicture', 'filter_checkbox', array(
+                'apply_filter' => function (QueryInterface $filterQuery, $field, $values) {
+                    $value = $values['value'];
```
j0k3r: Since you won't use `$value` after, you can directly use `$values['value']` in your `if` case.

nicosomb: removed (and already squashed)

PR #1419, opened 2015-09-13
[WIP] fosoauth

nicosomb:


j0k3r:
Why did you come back on fosuser 1.3 when you jumped to 2.0 because of a (I can't remember which one) bug ?

nicosomb:
I can't remember the bug. 
Why 1.3? If we want to use sonata admin and sonata user (why not?), we need fosuser 1.3 (I lost my wednesday afternoon at work due to this problem)...

j0k3r:
On Gitter, you told me about that issue: https://github.com/FriendsOfSymfony/FOSUserBundle/issues/1800

nicosomb:
You're right.. But I can't reproduce it at the moment ...

PR #1421, opened 2015-09-15
V2 comments

tcitworld:
comments

As for now, the way to add a new comment is in the sidebar, maybe the textarea can be shown all the time (especially for mobile view ?)


src/Wallabag/CoreBundle/Entity/Comment.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Wallabag\CoreBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Doctrine\Common\Collections\ArrayCollection;
+
+/**
+ * Tag.
```
j0k3r: This should be `Comment` :)

src/Wallabag/CoreBundle/Entity/Comment.php
```diff
@@ -0,0 +1,109 @@
...
+     * @ORM\JoinColumn(nullable=false)
+     */
+    private $entry;
+
+	/**
+	 * @var datetime $date
+	 *
+	 * @ORM\Column(name="date", type="datetime")
+	 */
+	private $date;
```
j0k3r: You missed one space on each line.

tcitworld: For the indentation ?

j0k3r: Yep

tcitworld: I count four of them.

j0k3r: It seems to be tab instead of spaces.
Don't hesitate to install php-cs-fixer and run it against your change (or the whole project)

tcitworld: Thanks for the tip.

marienfressinaud: I was looking yesterday for such a tool and found PHP_CodeSniffer but wasn't totally satisfied. Thanks @j0k3r, I will have a look too :)

tcitworld:
Can't see why the created date filter test fails here. Why doesn't this happens on v2 branch ?

tcitworld:
@modos189 Do you have ideas to make the comment section more "sexy" ?

j0k3r:
It fails because you've added a new entry `$entry7`. So, before that we were waiting for 5 items and we now get 6. You should update the test to set 6 instead of 5 and it'll be ok.

tcitworld:
Thanks, that was obvious.

modos189:
I can not understand the essence of the new feature.

Comments are not parsed from the page, and added manually by the user? If so, it does not comment, it notes. And the name of the user (in this case, "admin") is not needed.

nicosomb:
It's not comments, it's note in fact. 
You read the article, then you can add a note on this part of the text.

modos189:
So far, there are no ideas.
In the full version, you can use libreoffice-like view:

screenshot4

but it is nice to add to the mobile version of do not know

tcitworld:
I can try to save a position in the DOM where the comment is saved to do this aside view.

modos189:
Only need to remember that the width of the text depends on the size of the screen

tcitworld:
@j0k3r Can I ask your help for the two last errors ?

j0k3r:
Wow so much commits here :)
I think you should wait for #1434 to be merged and then rebase your branch. I've updated some tests to avoid _regression_ on them.

tcitworld:
Yeah, it's getting quite ugly (and for some unknown merge reason some of them were made two times), I'll squash all this when it's done.

src/Wallabag/CoreBundle/Form/Type/NewCommentType.php
```diff
@@ -0,0 +1,34 @@
...
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class NewCommentType extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('content', 'textarea', array('required' => true,
+                'attr' => array('class' => 'materialize-textarea'),
```
j0k3r: Be careful, it should also works for baggy. Does baggy know that class?

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -47,10 +48,25 @@ public function updateEntry(Entry $entry, $url)
...
+                $paragraph->appendChild($domAttribute);
+                ++$paragraphId;
+            }
+            $html = $html5->saveHTML($crawler);
+        }
+
         $entry->setUrl($content['url'] ?: $url);
         $entry->setTitle($title);
         $entry->setContent($html);
-        $entry->setLanguage($content['language']);
```
j0k3r: I guess this line shouldn't be remove (and that's why tests are failing).

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
```diff
@@ -156,13 +164,36 @@ main {
...
+                    </div>
+                {% endfor %}
+                </div>
+            {% endif %}
+        </div>
 {% endblock %}
 
 {% block footer %}
-{% endblock %}
+{% endblock %}
```
j0k3r: Do not forget to install a plugin on your IDE to handle `.editorconfig` file to avoid empty line at the end of file.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -47,10 +48,25 @@ public function updateEntry(Entry $entry, $url)
             }
         }
 
+        /*
+         * Set unique ids on paragraphs to set comments on the right ones
+         */
+        if (!strcmp($html, strip_tags($html)) == 0) {
+            $html5 = new HTML5();
```
j0k3r: What if the html isn't html5?
Why don't you use DomDocument instead?

tcitworld: Well, it raises many errors when it encounters html5 elements.

src/Wallabag/CoreBundle/Form/Type/NewCommentType.php
tcitworld: I have to call it if I want my textarea stylised in material theme, so I guess I'll have to implement it too in baggy.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
tcitworld: > What if the html isn't html5?

It uses DomDocument all the same, I think. I didn't encounter compatibility issues on the few non-html5 websites I tested.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
tcitworld: That's the one, thanks.

tcitworld:
Having issues, but only with pgsql. Strange.

j0k3r:
I'll try to have a look tonight.

j0k3r:
> Foreign key violation: 7 ERROR:  update or delete on table "wallabag_entry" violates foreign key constraint "fk_3ff60356ba364942" on table "wallabag_comment"

You should update the schema to delete on cascade comment when you remove an entry.

j0k3r:
@tcitworld Did you manage to fix the problem with cascade deletion?

tcitworld:
Nope, didn't touch it. Will do tonight maybe.

j0k3r:
Do you need extra help here?

tcitworld:
It still needs a new Form Type to edit comments, and adaptation to baggy. The big deal is the front-end, it fact, quite tricky to adapt.

nicosomb:
@tcitworld Do you need @modos189 help?

tcitworld:
I'll use Annotator instead. Everything in this branch is to be thrown away.

nicosomb:
So you can close this one, maybe?

tcitworld:
Yup. But keep the branch for now.

PR #1422, opened 2015-09-15
V2 – Export entries

### LINKED ISSUE #79: export epub

### LINKED ISSUE #701: Add .mobi export
Hi everybody,
In addition to the epub export it would be great to support also mobi export. I did some research and found this interesting project https://github.com/raiju/phpMobi. Maybe you can have a look at it and consider using it for wallabag.
Thanks

tcitworld:
What routes would you like ?

For now I just have `/category/ebook/format` for groups of articles and `/ebook/id/format` for one entry.

Fix #79
Fix #701

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/ExportController.php
+++ b/src/Wallabag/CoreBundle/Controller/ExportController.php
@@ -4,62 +4,55 @@ namespace Wallabag\CoreBundle\Controller;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Wallabag\CoreBundle\Entity\Entry;
-use Wallabag\CoreBundle\Helper\EntriesExport;
 
 class ExportController extends Controller
 {
-    /**
-     * Gets all entries for current user.
-     *
-     * @Route("/export/{category}.{format}", name="ebook", requirements={
-     *     "_format": "epub|mobi|pdf|json|xml|txt|csv"
-     * })
-     */
-    public function getEntriesAction($format, $category)
-    {
-        $repository = $this->getDoctrine()->getRepository('WallabagCoreBundle:Entry');
-        switch ($category) {
-            case 'all':
-                $method = 'All';
-                break;
-
-            case 'unread':
-                $method = 'Unread';
-                break;
-
-            case 'starred':
-                $method = 'Starred';
-                break;
-
-            case 'archive':
-                $method = 'Archive';
-                break;
-
-            default:
-                break;
-        }
-
-        $methodBuilder = 'getBuilderFor'.$method.'ByUser';
-        $qb = $repository->$methodBuilder($this->getUser()->getId());
-        $entries = $qb->getQuery()->getResult();
-
-        $export = new EntriesExport($entries);
-        $export->setMethod($method);
-        $export->exportAs($format);
-    }
-
     /**
      * Gets one entry content.
      *
      * @param Entry $entry
      *
-     * @Route("/export/id/{id}.{format}", requirements={"id" = "\d+"}, name="ebook_entry")
+     * @Route("/export/{id}.{format}", requirements={"id" = "\d+"}, name="export_entry")
      */
-    public function getEntryAction(Entry $entry, $format)
+    public function downloadEntryAction(Entry $entry, $format)
     {
-        $export = new EntriesExport(array($entry));
-        $export->setMethod('entry');
-        $export->exportAs($format);
+        try {
+            return $this->get('wallabag_core.helper.entries_export')
+                ->setEntries($entry)
+                ->updateTitle('entry')
+                ->exportAs($format);
+        } catch (\InvalidArgumentException $e) {
+            throw new NotFoundHttpException($e->getMessage());
+        }
+    }
+
+    /**
+     * Export all entries for current user.
+     *
+     * @Route("/export/{category}.{format}", name="export_entries", requirements={
+     *     "_format": "epub|mobi|pdf|json|xml|txt|csv",
+     *     "category": "all|unread|starred|archive"
+     * })
+     */
+    public function downloadEntriesAction($format, $category)
+    {
+        $method = ucfirst($category);
+        $methodBuilder = 'getBuilderFor'.$method.'ByUser';
+        $entries = $this->getDoctrine()
+            ->getRepository('WallabagCoreBundle:Entry')
+            ->$methodBuilder($this->getUser()->getId())
+            ->getQuery()
+            ->getResult();
+
+        try {
+            return $this->get('wallabag_core.helper.entries_export')
+                ->setEntries($entries)
+                ->updateTitle($method)
+                ->exportAs($format);
+        } catch (\InvalidArgumentException $e) {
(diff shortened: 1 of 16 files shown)


j0k3r:
What kind of routes do we need? Do we really need to export all articles in an ebook format?

Maybe we can have `/export/TYPE/CATEGORY.FORMAT`, for example `/export/ebook/unread.json` and `/export/ebook/123.json`.

tcitworld:
Is TYPE necessary then ? Why not directly `/export/unread.json`

j0k3r:
Oh I see, sorry. Yes we can have :

- `/export/unread.json`
- `/export/all.xml`
- `/export/123.epub`
- `/export/read.pdf`
- `/export/archive.mobi`
- `/export/unread.csv`
- etc ...

tcitworld:
Right. I'll just have `/export/tag/tagname.format` too.

I also had the idea to use the existing routes and just add `.FORMAT` at the end, but it seems I have to use FOSRestBundle instead of just a controller.

tcitworld:
Do you think a side panel (as for filters) is necessary for export on categories ? Of course, there will be more options to export to (xml, json, csv, txt, ...) and they will be stylised.

export1
export2

j0k3r:
It could be a good solution. But all these icons without a single text seems weird to me. First user impression will be hard to understand all the icons. But it's an other subject ...

tcitworld:
You mean on the first screenshot ? This one's may be better. But yes, I see your point.
export3

nicosomb:
@j0k3r you're right. What can we do?

j0k3r:
Well, can we replace them with plain text? :kissing_smiling_eyes:

nicosomb:
4 words instead of icons? It will be longer ...

tcitworld:
> First user impression will be hard to understand all the icons

There's also title attributes on all the icons.

j0k3r:
You should better do a `git rebase origin/v2` instead of merging the v2 branch in your own branch.

tcitworld:
Yeah, really sorry for the mess. I had also issues with rebase in the past so I don't really what to use each time... even though the documentation seems clear about this.

tcitworld:
Tests are failing on Travis, but not on my side. Hope I didn't blacklist github from LeMonde with all my tests. :(

nicosomb:
It's a graby's bug ;-) I found it this afternoon.

j0k3r:
It's in fact related to safecurl. I'm fixing it atm.

j0k3r:
:arrow_right: https://github.com/j0k3r/graby/releases/tag/1.0.1

tcitworld:
Thanks, I'll wait a little for it to come through Packagist.

j0k3r:
You can easily refactor `getEntriesAction` with a smaller switch case.
What I'll do:

- `Ebook` should take only one param: entries
- add a method `setMethod()` which take one parameter: the type (unread, archived, etc ..)
- add a method called `exportAs()` which take one parameter: the format
- move Ebook declaration outside the switch.
- defined the repository to use inside the switch (`$method = 'getBuilderForAllByUser';`) and then call it outside it, like `$repository->$method($this->getUser()->getId());`

Also Ebook doesn't have a meaning since you can now export as csv, zip, etc .. Maybe `EntriesExport` is a better name ?

tcitworld:
Thanks, will do that.

src/Wallabag/CoreBundle/Controller/ExportController.php
```diff
@@ -0,0 +1,65 @@
...
+     * Gets all entries for current user.
+     *
+     * @Route("/export/{category}.{format}", name="ebook", requirements={
+     *     "_format": "epub|mobi|pdf|json|xml|txt|csv"
+     * })
+     */
+    public function getEntriesAction($format, $category)
+    {
+        $repository = $this->getDoctrine()->getRepository('WallabagCoreBundle:Entry');
+        switch ($category) {
```
j0k3r: Instead of this big switch / case, you can define what `category` can be in the url (the same way you did for `_format` and then you just need to apply a `ucfirst()` to have the `$method`.

src/Wallabag/CoreBundle/Helper/EntriesExport.php
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * Sets the output format.
+     *
+     * @param string $format
+     */
+    public function exportAs($format)
+    {
+        $this->format = $format;
```
j0k3r: You don't need to re-set the format, you don't use `$this->format` outside of this function

src/Wallabag/CoreBundle/Helper/EntriesExport.php
```diff
@@ -0,0 +1,263 @@
...
+
+        // set image scale factor
+        $pdf->setImageScale(PDF_IMAGE_SCALE_RATIO);
+
+        $pdf->Output($this->title.'.pdf', 'D');
+    }
+
+    private function produceCSV()
+    {
+        header('Content-type: application/csv');
```
j0k3r: You should better create your own Response (from the HttpFoundation) instead of using php internal way.

src/Wallabag/CoreBundle/Helper/EntriesExport.php
```diff
@@ -0,0 +1,263 @@
...
+                $this->title = 'Starred articles';
+                break;
+            case 'Archive':
+                $this->title = 'Archived articles';
+                break;
+            case 'entry':
+                $this->title = $this->entries[0]->getTitle();
+                break;
+            default:
+                break;
```
tcitworld: If I directly take the value from the route, how do I deal with wrong categories here ?

j0k3r: Since won't affect this part.
Category from the route will only limit the categories you can get. If someone provide a wrong category in the route, the Router won't match the route and will lead to a 404.

You'll still have the `$method` here.

What I said above is that you don't need the switch / case,  just a simple:

```php
$method = ucfirst($category);
```

j0k3r:
Do you need extra help here?

tcitworld:
As of today it works and I made the changes you suggested. I just need to make things clean (return a proper Response for instance), avoiding dirty things [like this](https://github.com/wallabag/wallabag/pull/1422/files#diff-dc5f6d7a8b7bcf372eabb30fb2045c3eR261) and a little help would be welcomed here. I'll have to add JSON and TXT then.

j0k3r:
I can work on that tomorrow!

tcitworld:
For some reason when the serializer is used content is outputted twice in the file. Can somebody have a look at that ?

j0k3r:
Did JMS fix your duplicate output ?

tcitworld:
Nope, it didn't.

j0k3r:
I've checked the PR and didn't get duplicate output. Did you check that the duplicate output you get didn't come from different entries WITH same data? Since loaded fixtures are quite the same.

Also, I've added better test on json & xml export.

j0k3r:
Maybe we can review this PR and if it's ok to @wallabag/owners we can publish alpha1 with that feature :rocket:

tcitworld:
Hmm, I think to remember it was happening too with any new entry fetched, exported as XML or JSON. Will check again then.

nicosomb:
sorry, some conflicts on your branch

j0k3r:
@nicosomb rebased

nicosomb:
It works nice :+1: Good job @tcitworld

PR #1434, opened 2015-09-26
Add multiple database tests on Travis

j0k3r:
Support:

- MySQL
- SQLite
- PostgreSQL

To avoid bad behavior when testing archived, starred, deleted, etc .. I try to use the same content every time:
- I create the content (using a custom `$url`)
- I re-use this `$url` to retrieve the content on every test

I also moved setup of readingTime & domainName in ContentProxy to centralize all the work performed when saving a new content.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,10 +34,9 @@ branches:
     only:
         - v2
 
-install:
-    - composer self-update
-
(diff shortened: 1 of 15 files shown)


nicosomb:
:heart:

PR #1464, opened 2015-10-06
fix type hint for User

nicosomb:


j0k3r:
Well, if there is this error, this means there is no test that cover that line :warning:

nicosomb:
OK, but we never wrote tests for methods like that. What do you suggest?

j0k3r:
How did you see the error?

nicosomb:
Not locally, when I deployed and when I ... launch a command maybe, I can't remember, it was yesterday.

PR #1471, opened 2015-10-07
implement bookmarklet

nicosomb:
Do we need a test for this PR?

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -15,6 +15,23 @@ use Pagerfanta\Pagerfanta;
 class EntryController extends Controller
 {
     /**
+     * @param Entry $entry
+     */
+    private function updateEntry(Entry $entry)
+    {
+        try {
+            $entry = $this->get('wallabag_core.content_proxy')->updateEntry($entry, $entry->getUrl());
+            $em = $this->getDoctrine()->getManager();
+            $em->persist($entry);
+            $em->flush();
+        } catch (\Exception $e) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
      * @param Request $request
      *
      * @Route("/new-entry", name="new_entry")
@@ -30,12 +47,7 @@ class EntryController extends Controller
         $form->handleRequest($request);
 
         if ($form->isValid()) {
-            $entry = $this->get('wallabag_core.content_proxy')->updateEntry($entry, $entry->getUrl());
-
-            $em = $this->getDoctrine()->getManager();
-            $em->persist($entry);
-            $em->flush();
-
+            $this->updateEntry($entry);
             $this->get('session')->getFlashBag()->add(
                 'notice',
                 'Entry saved'
@@ -47,6 +59,22 @@ class EntryController extends Controller
         return $this->render('WallabagCoreBundle:Entry:new_form.html.twig', array(
             'form' => $form->createView(),
         ));
+    }
+
+    /**
+     * @param Request $request
+     *
+     * @Route("/bookmarklet", name="bookmarklet")
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function addEntryViaBookmarklet(Request $request)
+    {
+        $entry = new Entry($this->getUser());
+        $entry->setUrl($request->get('url'));
+        $this->updateEntry($entry);
+
+        return $this->redirect($this->generateUrl('homepage'));
     }
 
     /**
--- a/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
+++ b/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
@@ -31,6 +31,31 @@ class EntryControllerTest extends WallabagCoreTestCase
         $this->assertCount(1, $crawler->filter('button[type=submit]'));
     }
 
+    public function testPostNewViaBookmarklet()
+    {
+        $this->logInAs('admin');
+        $client = $this->getClient();
+
+        $crawler = $client->request('GET', '/');
+
+        $this->assertCount(4, $crawler->filter('div[class=entry]'));
+
+        // Good URL
+        $crawler = $client->request('GET', '/bookmarklet', array('url' => $this->url));
+        $this->assertEquals(302, $client->getResponse()->getStatusCode());
+        $crawler = $client->followRedirect();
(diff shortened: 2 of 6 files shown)


src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -52,6 +58,22 @@ public function addEntryFormAction(Request $request)
...
      *
+     * @Route("/w", name="new_bookmarklet")
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function addEntryViaBookmarklet(Request $request)
+    {
+        $entry = new Entry($this->getUser());
+        $entry->setUrl($request->get('url'));
+        $this->updateEntry($entry);
```
j0k3r: You can simplify to :

```php
$this->updateEntry(new Entry($this->getUser()), $request->get('url'));
```

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -15,6 +15,17 @@
...
+     */
+    private function updateEntry(Entry $entry)
+    {
+        $entry = $this->get('wallabag_core.content_proxy')->updateEntry($entry, $entry->getUrl());
+
+        $em = $this->getDoctrine()->getManager();
+        $em->persist($entry);
+        $em->flush();
+    }
+    /**
```
j0k3r: You should add one empty line before

src/Wallabag/CoreBundle/Resources/views/base.html.twig
```diff
@@ -73,5 +73,9 @@
             {{ warning_message | raw }}
         </div>
         {% endif %}
+        <script src="{{ asset('themes/_global/js/hideShowPassword.min.js') }}"></script>
```
j0k3r: Shouldn't this be included only in the registration page ?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -52,6 +58,22 @@ public function addEntryFormAction(Request $request)
     /**
      * @param Request $request
      *
+     * @Route("/w", name="new_bookmarklet")
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function addEntryViaBookmarklet(Request $request)
```
j0k3r: You might need a test for that action.
And shouldn't this method be `addEntryViaBookmarkletAction` instead?

And why `/w` ?

nicosomb: Because I had to define a route.

Envoyé depuis mon appareil Android.

-----Original Message-----

j0k3r: Why not `/bookmarklet` then?

nicosomb: I wanted a shorter route, but why not bookmarklet :-)

src/Wallabag/CoreBundle/Resources/views/base.html.twig
nicosomb: nothing to do in this PR. sorry.

nicosomb:
Route changed & test added.

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -52,6 +59,25 @@ public function addEntryFormAction(Request $request)
...
      * @param Request $request
      *
+     * @Route("/bookmarklet", name="bookmarklet")
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function addEntryViaBookmarklet(Request $request)
+    {
+        $url = $request->get('url');
+        if (filter_var($url, FILTER_VALIDATE_URL)) {
```
j0k3r: I'm curious, what happend if you don't do that? You got an exception or sth?
This kind of verification is already done in graby.

src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
```diff
@@ -31,6 +31,30 @@ public function testGetNew()
...
+        $this->assertCount(5, $crawler->filter('div[class=entry]'));
+
+        // Bad URL
+        $crawler = $client->request('GET', '/bookmarklet', array('url' => 'fakeURL'));
+        $this->assertEquals(302, $client->getResponse()->getStatusCode());
+        $crawler = $client->followRedirect();
+        $crawler = $client->request('GET', '/');
+        $this->assertCount(5, $crawler->filter('div[class=entry]'));
+    }
+
```
j0k3r: Instead of updating all count below, you should better remove the entry after added it.
You can remove it using the url.

src/Wallabag/CoreBundle/Controller/EntryController.php
nicosomb: It's an exception launched by graby. I have to catch it in fact.

src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
nicosomb: ok

nicosomb:
I removed the test about  "Bad URL". In fact, the behaviour if we have a bad URL is the same as before: we create a new entry with no content and "fakeURL" in the URL field. 
I think it has to be fixed in graby maybe.

nicosomb:
Tests are green but GitHub lost the relation with travis https://travis-ci.org/wallabag/wallabag

j0k3r:
Maybe there is something to fix in graby but ContentProxy also need to handle Exception in case of graby fail.
We'll do that later

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -5,6 +5,7 @@
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Acl\Exception\Exception;
```
j0k3r: Is it used somewhere?

nicosomb: thanks for your vigilence.

nicosomb:
ready to be rebased.

j0k3r:
Good to be rebased and then merged :+1:

Issue #1477, opened 2015-10-11
[v2] Rule-based tags


K-Phoen:
Would you be interested by a way to tag entries based on rules written by the users?

For instance, I'd like to be able to define the following rules:
* `domainName = "github.com"`, and tag the matching entries as `github` and `to check`
* `readingTime >= 15`, and tag the matching entries as `long reading`
* `readingTime <= 3`, and tag the matching entries as `short reading`
* …

These rules can be combined, so an entry would receive the tags `github` and `long reading` if it matches both rules.

More complex rules can be imagined and the attributes exposed in the rules would be the same as the ones defined in entries (title, url, domainName, …).

If you think that this kind of feature would be useful in wallabag, I could draft a PoC and submit a PR :)

j0k3r:
It could be a very interesting feature. I guess you are going to use Hoa\Ruler ? :)

K-Phoen:
I was going to use [RulerZ](https://github.com/K-Phoen/RulerZ) (which is based on Hoa\Ruler). It's faster and it could be useful if we want to make these rules retroactive :)

nicosomb:
Oh nice idea.
You can send us a PR :-)



Sur 11 oct. 2015 à 13:08, à 13:08, "Kévin Gomez" <notifications@github.com> a écrit:

tcitworld:
Do you think we should extend the filtering panel to create these rules, or create a new dedicated page ?

j0k3r:
@tcitworld a new dedicated page seems better to me (according to options and the little FAQ)

PR #1478, opened 2015-10-11
Rule based tags

K-Phoen:
As described in #1477, here is my PoC for a rule-based automatic tagging.

It rules engine is [RulerZ](https://github.com/K-Phoen/rulerz) and I created a new tab in the configuration panel to manage the tagging rules. It's a quick draft but at least it allows us to create rules and test them:

capture d ecran de 2015-10-11 18-41-05

It's really a WIP so there are no tests, validation for the rules or help about the DSL used for the rules.

If you like the overall direction of this feature, I'll finish my PR :)

Note: there are a lot of *additions* but it's mainly the `composer.lock file`. Also, reading the PR commit-by-commit should be easier.

**Updated screenshot**:

image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "scheb/two-factor-bundle": "~1.4",
         "grandt/phpepub": "~4.0",
         "wallabag/php-mobi": "~1.0.0",
-        "kphoen/rulerz-bundle": "~0.7,>=0.7.3"
+        "kphoen/rulerz-bundle": "~0.9,>=0.9.1"
     },
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "~2.2.0",
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "scheb/two-factor-bundle": "~1.4",
         "grandt/phpepub": "~4.0",
         "wallabag/php-mobi": "~1.0.0",
-        "kphoen/rulerz-bundle": "dev-master"
+        "kphoen/rulerz-bundle": "~0.7,>=0.7.2"
     },
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "~2.2.0",
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "scheb/two-factor-bundle": "~1.4",
         "grandt/phpepub": "~4.0",
         "wallabag/php-mobi": "~1.0.0",
-        "kphoen/rulerz-bundle": "~0.7,>=0.7.2"
+        "kphoen/rulerz-bundle": "~0.7,>=0.7.3"
     },
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "~2.2.0",
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,8 @@
         "friendsofsymfony/oauth-server-bundle": "^1.4@dev",
         "scheb/two-factor-bundle": "~1.4",
         "grandt/phpepub": "~4.0",
-        "wallabag/php-mobi": "~1.0.0"
+        "wallabag/php-mobi": "~1.0.0",
+        "kphoen/rulerz-bundle": "dev-master"
     },
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "~2.2.0",
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "scheb/two-factor-bundle": "~1.4",
         "grandt/phpepub": "~4.0",
         "wallabag/php-mobi": "~1.0.0",
-        "kphoen/rulerz-bundle": "~0.9,>=0.9.1"
+        "kphoen/rulerz-bundle": "~0.10"
     },
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "~2.2.0",
--- a/src/Wallabag/CoreBundle/Controller/ConfigController.php
+++ b/src/Wallabag/CoreBundle/Controller/ConfigController.php
@@ -102,7 +102,7 @@ class ConfigController extends Controller
 
         // handle tagging rule
         $taggingRule = new TaggingRule();
-        $newTaggingRule = $this->createForm(new TaggingRuleType(), $taggingRule);
+        $newTaggingRule = $this->createForm(new TaggingRuleType(), $taggingRule, array('action' => $this->generateUrl('config').'#set5'));
         $newTaggingRule->handleRequest($request);
 
         if ($newTaggingRule->isValid()) {
--- a/src/Wallabag/CoreBundle/Controller/ConfigController.php
+++ b/src/Wallabag/CoreBundle/Controller/ConfigController.php
@@ -7,9 +7,11 @@ use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Wallabag\CoreBundle\Entity\Config;
+use Wallabag\CoreBundle\Entity\TaggingRule;
 use Wallabag\UserBundle\Entity\User;
 use Wallabag\CoreBundle\Form\Type\ChangePasswordType;
 use Wallabag\CoreBundle\Form\Type\UserInformationType;
+use Wallabag\CoreBundle\Form\Type\TaggingRuleType;
 use Wallabag\CoreBundle\Form\Type\NewUserType;
 use Wallabag\CoreBundle\Form\Type\RssType;
 use Wallabag\CoreBundle\Tools\Utils;
@@ -98,6 +100,24 @@ class ConfigController extends Controller
             return $this->redirect($this->generateUrl('config'));
         }
 
+        // handle tagging rule
+        $taggingRule = new TaggingRule();
+        $newTaggingRule = $this->createForm(new TaggingRuleType(), $taggingRule);
+        $newTaggingRule->handleRequest($request);
+
+        if ($newTaggingRule->isValid()) {
+            $taggingRule->setConfig($config);
+            $em->persist($taggingRule);
+            $em->flush();
+
+            $this->get('session')->getFlashBag()->add(
+                'notice',
+                'Tagging rules updated'
+            );
+
+            return $this->redirect($this->generateUrl('config'));
+        }
+
         // handle adding new user
         $newUser = $userManager->createUser();
         // enable created user by default
@@ -136,6 +156,7 @@ class ConfigController extends Controller
                 'pwd' => $pwdForm->createView(),
                 'user' => $userForm->createView(),
                 'new_user' => $newUserForm->createView(),
+                'new_tagging_rule' => $newTaggingRule->createView(),
             ),
             'rss' => array(
                 'username' => $user->getUsername(),
--- a/src/Wallabag/CoreBundle/Controller/ConfigController.php
+++ b/src/Wallabag/CoreBundle/Controller/ConfigController.php
@@ -189,6 +189,33 @@ class ConfigController extends Controller
     }
 
     /**
+     * Deletes a tagging rule and redirect to the config homepage.
+     *
+     * @param TaggingRule $rule
+     *
+     * @Route("/tagging-rule/delete/{id}", requirements={"id" = "\d+"}, name="delete_tagging_rule")
+     *
+     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     */
+    public function deleteTaggingRule(TaggingRule $rule)
+    {
+        if ($this->getUser()->getId() != $rule->getConfig()->getUser()->getId()) {
+            throw $this->createAccessDeniedException('You can not access this tagging ryle.');
+        }
+
+        $em = $this->getDoctrine()->getManager();
+        $em->remove($rule);
+        $em->flush();
+
+        $this->get('session')->getFlashBag()->add(
+            'notice',
+            'Tagging rule deleted'
+        );
+
+        return $this->redirect($this->generateUrl('config'));
+    }
+
+    /**
      * Retrieve config for the current user.
      * If no config were found, create a new one.
      *
--- a/src/Wallabag/CoreBundle/Entity/TaggingRule.php
+++ b/src/Wallabag/CoreBundle/Entity/TaggingRule.php
@@ -7,7 +7,7 @@ use Symfony\Component\Validator\Constraints as Assert;
 use KPhoen\RulerZBundle\Validator\Constraints as RulerZAssert;
 
 /**
- * Config.
+ * Tagging rule.
  *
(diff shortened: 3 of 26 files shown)


tcitworld:
Impressive ! Can you fix the Travis Error (it's just a `use Doctrine\Common\Collections\ArrayCollection;` missing in `Entity/Config.php` ).

K-Phoen:
Should be better now :)

composer.json
```diff
@@ -54,7 +54,8 @@
         "lexik/form-filter-bundle": "~4.0",
         "j0k3r/graby": "~1.0",
         "friendsofsymfony/user-bundle": "dev-master",
-        "friendsofsymfony/oauth-server-bundle": "^1.4@dev"
+        "friendsofsymfony/oauth-server-bundle": "^1.4@dev",
+        "kphoen/rulerz-bundle": "dev-master"
```
j0k3r: Can't we use a tagged version instead?

src/Wallabag/CoreBundle/Form/DataTransformer/StringToListTransformer.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Wallabag\CoreBundle\Form\DataTransformer;
+
+use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Component\Form\DataTransformerInterface;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+class StringToListTransformer implements DataTransformerInterface
```
j0k3r: Could you add a description about what this class is supposed to do and where it is (for now) used ?

src/Wallabag/CoreBundle/Resources/views/themes/material/Config/index.html.twig
```diff
@@ -15,8 +15,9 @@
                         <li class="tab col s3"><a href="#set2">{% trans %}RSS{% endtrans %}</a></li>
                         <li class="tab col s3"><a href="#set3">{% trans %}User information{% endtrans %}</a></li>
                         <li class="tab col s3"><a href="#set4">{% trans %}Password{% endtrans %}</a></li>
+                        <li class="tab col s3"><a href="#set5">{% trans %}Tags{% endtrans %}</a></li>
```
j0k3r: I'll use `Tagging rules` instead

j0k3r:
This a good start! I wasn't aware that this bundle will include all that stuff from Hoa project :flushed:

Anyway, apart my first little comments:
- a better kind of FAQ could be great for novice people, something like "Let assume you want to tag all BLABLA when BLABLA happen, you should put THAT in THIS field", etc ...
- don't forget unit test on `StringToListTransformer` & `RuleBasedTagger`
- did you validate the rule somewhere when adding it? I mean, what if a user set an invalid rule (let's assume it's possible) will it be notified if the rule is wrong?

K-Phoen:
> This a good start! I wasn't aware that this bundle will include all that stuff from Hoa project :flushed:

My bad, I didn't properly introduce RulerZ :)
This library primarily adds a layer on top of Hoa\Ruler allowing us to use rules on objects as well as any other data-source, that's why requiring RulerZ also requires a few libraries from the Hoa Project.
RulerZ also goes a step further than Hoa\Ruler: it really compiles rules instead of having to evaluate on-the-fly an abstract representation of the rule.

> a better kind of FAQ could be great for novice people, something like "Let assume you want to tag all BLABLA when BLABLA happen, you should put THAT in THIS field", etc ...

[Agreed](https://images.duckduckgo.com/iu/?u=http%3A%2F%2Ftse3.mm.bing.net%2Fth%3Fid%3DOIP.M257282ce52643e8f8ac79bce9d9ac84do0%26pid%3D15.1&f=1). I just wrote a list of the usable variables so that you could see the possibilities of the feature. I'll try to write a proper FAQ once the code is done :)

> don't forget unit test on StringToListTransformer & RuleBasedTagger

Done for `StringToListTransformer`, I will write tests for `RuleBasedTagger` later.

> did you validate the rule somewhere when adding it? I mean, what if a user set an invalid rule (let's assume it's possible) will it be notified if the rule is wrong?

It was on my roadmap. I committed a first version of the validator today but it's really a naïve one: it just validates the syntax. For instance, if you use an undefined variable no validation error will be triggered. I'll improve the validator to detect these errors.

nicosomb:
I just gave a try :+1: 

Do you think that it's possible to have a command to assign these tags to existing articles?

K-Phoen:
I also had this idea in mind, that's why I used RulerZ instead of Hoa\Ruler.

While Hoa\Ruler only works on an object (in our case, an `Entry`), RulerZ can fetch all entries matching a given rule directly from the database.

So yeah, writing a command/controller to tag existing articles should be easy :)

nicosomb:
Nice. 

Don't forget to give availability to remove rule. 
I created a new rule and all seems to be broken. My rule `if « content like "%basket%" » then tag as « basket »` (some doc will be necessary ;-) because I'm sure it's not the good syntax ), and I can't save a new entry.

K-Phoen:
Yep, there is still some work to be done on the validation. Also, I'll make sure that even if the automatic tagging fails it does not prevent the user to add new entries.

K-Phoen:
Thanks for your feedback btw :+1:

K-Phoen:
I implemented a better validator for the tagging rules and also provided a way to delete them.
There is also a `wallabag:tag:all` command to tag all entries for a given user according to its tagging rules.

There is still some documentation about the rules to be written. You wanted that on a separate page? Do you already have one?

src/Wallabag/CoreBundle/Entity/TaggingRule.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+namespace Wallabag\CoreBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Symfony\Component\Validator\Constraints as Assert;
+use KPhoen\RulerZBundle\Validator\Constraints as RulerZAssert;
+
+/**
+ * Config.
```
j0k3r: Bad copy/paste?

K-Phoen: Yep.

src/Wallabag/CoreBundle/Tests/Helper/ContentProxyTest.php
```diff
@@ -26,7 +30,7 @@ public function testWithEmptyContent()
                 'language' => '',
             ));
 
-        $proxy = new ContentProxy($graby);
+        $proxy = new ContentProxy($graby, $tagger, $this->getLoggerMock());
```
j0k3r: You can use `Psr\Log\NullLogger` instead of mocking the logger

K-Phoen: Done.

j0k3r:
Looks good :smile: 

About the documentation your are talking about, is that the FAQ displayed below the tagging page or something better which need to be in the documentation of wallabag?
Maybe you can write a small FAQ in the tagging page (like I said in a previous comment: _Let assume you want to tag all BLABLA when BLABLA happen, you should put THAT in THIS field_)

And you can add a better documentation in our doc, which is now part of the repo since #1495. Maybe you'll have to rebase your PR to get that change.

nicosomb:
Conflict with v2 branch. Will try this PR maybe this evening.

K-Phoen:
I rebased against master and tried to write a little FAQ.
Feel free to modify it as I'm not very good at writing documentation…

nicosomb:
* How can we use `isArchived` and `isStared` with tagging rules? All new entries are unread and not unstarred.
* Is that possible to use a `like` operator? For content who contains some words. 
* sometimes, with more than one matching rule, the entry is not saved. No error in logs. Example with this URL `http://www.lequipe.fr/Basket/Actualites/Les-san-antonio-spurs-vainqueurs-des-portland-trailblazers-101-113-s-emparent-seuls-de-la-deuxieme-place-de-la-conference-ouest/607097` and this config : 
  * if « domainName = "www.lequipe.fr" » then tag as « sport » 
  * if « readingTime <= 20 » then tag as « test » 
If I only have the 1st rule, it's OK. I add the 2nd rule, fail. I remove the 2nd rule, fail again.
* do you plan to add the rule edition feature? 

Nice job, really impatient to have it into wallabag!

tcitworld:
> Is that possible to use a like operator? For content who contains some words. 

Maybe we should wait to have a search feature to use it here.

K-Phoen:
> How can we use `isArchived` and `isStared` with tagging rules? All new entries are unread and not unstarred.

As they are booleans, `isArchived` and `isStared` can be used like this: `isArchived == 1`.
I can remove them from the FAQ if you think they aren't relevant. As their value only changes when entries are archived or starred, tagging rules using these variables are useful only when applied with the console command (`wallabag:tag:all`).

> Is that possible to use a `like` operator? For content who contains some words. 

Something like `title matches "basket"` would be enough? The `matches` operator would work like the stripos PHP function, it would return true only if there is an exact match.

> do you plan to add the rule edition feature? 

It could be done but nope, I wasn't planning on writing it.

I also tried to reproduce your bug but I couldn't…

nicosomb:
> Something like title matches "basket" would be enough? The matches operator would work like the stripos PHP function, it would return true only if there is an exact match.

Yes :+1: 

> I also tried to reproduce your bug but I couldn't…

Will retry

> `wallabag:tag:all`

Oh nice, I didn't see this feature! What do you think having a button: "tag all the previous entries"?

j0k3r:
> What do you think having a button: "tag all the previous entries"?
Depending on the number of entries for the current user, it could take long time to accomplish ..

K-Phoen:
I added the `matches` operator :)

nicosomb:
<3 (just need documentation about this new operator maybe)

src/Wallabag/CoreBundle/Operator/Doctrine/Matches.php
```diff
@@ -0,0 +1,15 @@
...
+
+class Matches
+{
+    public function __invoke($subject, $pattern)
+    {
+        if ($pattern[0] === "'") {
+            $pattern = sprintf("'%%%s%%'", substr($pattern, 1, -1));
+        }
+
+        return sprintf('UPPER(%s) LIKE UPPER(%s)', $subject, $pattern);
```
j0k3r: Is it working on MySQL, PostGres & SQLite ? Maybe a working test for the matches keyword could be a good idea.

K-Phoen:
> <3 (just need documentation about this new operator maybe)

Oops, my bad, I forgot :)
I'll add it tonight.

> Is it working on MySQL, PostGres & SQLite ? Maybe a working test for the matches keyword could be a good idea.

It does. I'll add a test.

src/Wallabag/CoreBundle/Operator/PHP/Matches.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Wallabag\CoreBundle\Operator\PHP;
+
+class Matches
```
j0k3r: Could you precise what these classes are used for (both `Operator\Doctrine\Matches` & `Operator\Doctrine\PHP`) ?
In the context of your PR, we can guess but when it will be inside the project, it won't be that easy to guess

K-Phoen:
PHPdoc added, `matches` operator described in the FAQ and a little test was written :)

src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
```diff
@@ -102,6 +102,41 @@ public function testPostNewOk()
...
+        $this->assertEquals(302, $client->getResponse()->getStatusCode());
+
+        $crawler = $client->followRedirect();
+
+        $em = $client->getContainer()
+            ->get('doctrine.orm.entity_manager');
+        $entry = $em
+            ->getRepository('WallabagCoreBundle:Entry')
+            ->findOneByUrl($url);
+        $this->assertCount(1, $entry->getTags());
```
j0k3r: Could you just assert that the only tag for this entry is actually `wallabag`, as per the TaggingRule.

j0k3r:
I think we are really close to get this merged :rocket: 
Thanks for the hard work and for taking care of all the feedback we provided!

src/Wallabag/CoreBundle/Resources/views/themes/material/Config/index.html.twig
```diff
@@ -183,6 +184,153 @@
...
+                                            </tr>
+                                        </tbody>
+                                    </table>
+                                </p>
+                            </div>
+                        </div>
+                    </div>
+
                     {% if is_granted('ROLE_SUPER_ADMIN') %}
                     <div id="set5" class="col s12">
```
j0k3r: It might be a side effect of your change (and some rebase) but could you set `set6` here instead of `set5`? Because we have 2 `set5` in the page and it breaks the layout because, if user is super admin, _Add a user_ will always be displayed.

src/Wallabag/CoreBundle/Resources/views/themes/material/Config/index.html.twig
```diff
@@ -183,6 +184,153 @@
...
+                                        <a href="{{ path('delete_tagging_rule', {id: tagging_rule.id}) }}" title="{% trans %}Delete{% endtrans %}">
+                                            <i class="tool grey-text delete mdi-action-delete"></i>
+                                        </a>
+                                    </li>
+                                    {% endfor %}
+                                </ul>
+                            </div>
+                        </div>
+
+                        <form action="{{ path('config') }}#set5" method="post" {{ form_enctype(form.new_tagging_rule) }}>
```
j0k3r: Could you use `form_start(form.new_tagging_rule)` instead of this line? Mostly because `form_enctype` will be removed in Symfony3.0

src/Wallabag/CoreBundle/Controller/ConfigController.php
```diff
@@ -98,6 +100,24 @@ public function indexAction(Request $request)
             return $this->redirect($this->generateUrl('config'));
         }
 
+        // handle tagging rule
+        $taggingRule = new TaggingRule();
+        $newTaggingRule = $this->createForm(new TaggingRuleType(), $taggingRule);
```
j0k3r: Related to the change to `form_start`, could you add a third parameter to specify the action url of the form?

```php
$newTaggingRule = $this->createForm(new TaggingRuleType(), $taggingRule, array('action' => $this->generateUrl('config').'#set5'));
```

K-Phoen:
I found some time to fix the tests, they're green now :tada: 
Is there anything left to do before this PR is merged?

Anyway, thanks for all the valuable feedbacks you guys provided throughout this PR. Keep up the good work :+1:

tcitworld:
Note for later : Think on how to export rules from one installation to another.

K-Phoen:
Both rules and tags can be represented as plain text so that should not be difficult :)

j0k3r:
@K-Phoen Huge thank to you for that :+1:

Hywan:
:+1: \o/

PR #1483, opened 2015-10-13
fix Content-Disposition header #1481

### LINKED ISSUE #1481: Wrong Content-Disposition header when download epub (rfc1806)
When download an epub file, the content-disposition header value is wrong:

Result:
Content-Disposition: attached; filename="title.epub"

Should be:
Content-Disposition: attachment; filename="title.epub"

see: http://www.ietf.org/rfc/rfc1806.txt

Impact:
The filename is not set properly on some devices like my Sony Reader, and use default filename "filename.bin". The device cannot open the file with that name.

alkahan:
Fix #1481


j0k3r:
Thanks ! :+1:

PR #1484, opened 2015-10-13
2factor authentication via email

nicosomb:
2factor authentication via email.
This setting can be disabled on config screen for each user. 
User can also check "trusted computer" to avoid 2factor authentication for a delay. 

Here is a screenshot after login step. 

It's also possible to enable Google Authenticator, but I don't know if we need this validator. @j0k3r @tcitworld any opinion?


app/config/config.yml
```diff
@@ -171,3 +171,16 @@ fos_oauth_server:
     auth_code_class:     Wallabag\ApiBundle\Entity\AuthCode
     service:
         user_provider: fos_user.user_manager
+
+scheb_two_factor:
+
+    trusted_computer:
+        enabled: true
+        cookie_name: wllbg_trusted_computer
+        cookie_lifetime: 5184000
```
tcitworld: Maybe 30 days without access is enough ? 60 seems big to me.

src/Wallabag/UserBundle/Resources/config/services.yml
```diff
@@ -0,0 +1,19 @@
...
+    wallabag_user.twofactor.email.interactive_login_listener:
+        class: %wallabag_user.twofactor.email.interactive_login_listener.class%
+        tags:
+            - { name: kernel.event_listener, event: security.interactive_login, method: onSecurityInteractiveLogin }
+        arguments: ['@wallabag_user.twofactor.email.provider']
+    wallabag_user.twofactor.email.request_listener:
+        class: %wallabag_user.twofactor.email.request_listener.class%
+        tags:
+            - { name: kernel.event_listener, event: kernel.request, method: onCoreRequest, priority: -1 }
+        arguments: ['@wallabag_user.twofactor.email.provider', '@security.context', '@templating', '@router']
```
j0k3r: Could you make a list of arguments instead of all in a single line?

src/Wallabag/UserBundle/Resources/config/services.yml
```diff
@@ -0,0 +1,19 @@
+parameters:
+    wallabag_user.twofactor.email.provider.class: Wallabag\UserBundle\Security\TwoFactor\Email\Helper
+    wallabag_user.twofactor.email.interactive_login_listener.class: Wallabag\UserBundle\Security\TwoFactor\Email\InteractiveLoginListener
+    wallabag_user.twofactor.email.request_listener.class: Wallabag\UserBundle\Security\TwoFactor\Email\RequestListener
```
j0k3r: Did you forget to add all theses classes in your commit?

app/config/config.yml
```diff
@@ -171,3 +171,16 @@ fos_oauth_server:
     auth_code_class:     Wallabag\ApiBundle\Entity\AuthCode
     service:
         user_provider: fos_user.user_manager
+
+scheb_two_factor:
+
```
j0k3r: This empty line isn't necessary

j0k3r:
Where are tests related to that part?

src/Wallabag/UserBundle/Resources/config/services.yml
nicosomb: No I forgot to remove these lines. Will change it tomorrow. Zzzz time.

nicosomb:
I don't know if tests are needed: the bundle is tested and I don't override the bundle behaviour.

j0k3r:
I don't agree.

For example, you added `form.user.twoFactorAuthentication` in the configuration panel. What if user doesn't enable it, you should test that he won't get the form to put the 2 factor code. On the other side, if user enable it, you should test that the form will be shown when user log in.

Also, you added some logic in the User entity that need to be tested: `isTrustedComputer`.

nicosomb:
OK, I will add tests.

nicosomb:
Tests added.

app/config/config.yml
```diff
@@ -171,3 +171,15 @@ fos_oauth_server:
     auth_code_class:     Wallabag\ApiBundle\Entity\AuthCode
     service:
         user_provider: fos_user.user_manager
+
+scheb_two_factor:
+    trusted_computer:
+        enabled: true
```
j0k3r: I was thinking about these config to enable two factor.
I think they should be in `parameters.yml` so people won't have to update `config.yml` and keep a *modified file* which can be problematic when updating wallabag.

Also, the sender_email should be in parameters.yml

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Config/index.html.twig
```diff
@@ -100,6 +100,14 @@
             </div>
         </fieldset>
 
+        <fieldset class="w500p inline">
+            <div class="row">
```
j0k3r: And maybe we can display or even better disable this widget in case of the two factor isn't enabled

app/config/config.yml
nicosomb: done.

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Config/index.html.twig
nicosomb: done.

Issue #1432, opened 2015-09-23
API V2 Missing endpoints


foxmask:
from http://v2.wallabag.org/api/doc I think an endpoint is missing to get all the items from a given tag eg : 
#### api/tags/Pope.json

Get all entries from a tag. Started here: #1488 

currently (for the tags) we can only do http://v2.wallabag.org/api/tags.json in GET or http://v2.wallabag.org/api/tags/{tag}.json in DELETE 
#### Filter entries by tag

In GET /api/entries.json, we can set a tags array, but WallabagRestController::getEntriesAction() don't manage it. 
#### Filters
- preview picture: In GET /api/entries.json, add a boolean to fetch entries with picture preview only.
- domain name: in GET / api/entries.json, add parameter to fetch entries with specific domain name
- reading time: in GET / api/entries.json, add range to fetch entries with reading time in this range
- language: in GET / api/entries.json, add parameter to fetch entries with specific language
- creation date: in GET / api/entries.json, add range to fetch entries with create date in this range

nicosomb:
I updated the first post to add other missing endpoints. Please tell uf if other endpoints still missing. 
ping @foxmask @jlnostr @freddyb @kevmeyer

jlnostr:
As I told in [#1182](https://github.com/wallabag/wallabag/issues/1182#issuecomment-94302447), user-related information could be there in a `/api/user.format` request, which could contain the following things:

- shaarli/diaspora adresses
- the full name
- the avatar image URL
- oAuth token
- …

This would make the `user` node in each JSON item obsolete, which therefore should be removed.

nicosomb:
> shaarli/diaspora adresses
> the full name
> the avatar image URL

Avatar? We don't have avatar for user. 
Shaarli / diaspora are generic config for wallabag instance, it's not personal information. Do you really need these data?

tcitworld:
Maybe we should.

foxmask:
@nicosomb unfortunatly I dont remember what was the original content on my post. Si I dont know which endpoint(s) has/have been added. So I made a test with http://v2.wallabag.org/api/tags/Pope.json but its failed.
Am I doing something wrong ? Is the doc uptodate ?

foxmask:
@nicosomb unfortunatly I dont remember what was the original content on my post. Si I dont know which endpoint(s) has/have been added. So I made a test with http://v2.wallabag.org/api/tags/Pope.json but its failed.
[...]

PR #1488, opened 2015-10-14
Slug for tags

nicosomb:


j0k3r:
Why do we need this? Are you going to use it somewhere? If yes, why don't you do that in this PR? :)

nicosomb:
I will use it in API (#1432): to fetch entries for a specific tag.

j0k3r:
So maybe you can add the api endpoint right in this PR

j0k3r:
What left to be done here? Do you need help?

j0k3r:
Updated & rebased.

j0k3r:
@nicosomb I think we can merge that and check later for the missing endpoint.

nicosomb:
:+1:

PR #1518, opened 2015-11-26
SMTP from environment variables

mathbruyen:
Configuring SMTP parameters required to be written in files which is not convenient when using docker containers as the credentials need to be in the filesystem. Created environment parameters override for those parameters. A new build script created in order to generate those parameters right before launching the application.

This allows to build containers without [specifying credentials in the container itself](https://github.com/mathbruyen/home/blob/424542ae568b8c23af8d9585d4390db421cf9ca3/wallabag/Dockerfile) and [use them in kubernetes](https://github.com/mathbruyen/home/blob/424542ae568b8c23af8d9585d4390db421cf9ca3/wallabag-controller.yaml) (I'm also looking at [this PR on kubernetes](https://github.com/kubernetes/kubernetes/issues/4710) for storing the password in a secret out of source control rather than in the replication controller configuration but this is independent of wallabag).

All the parameters are prefixed with `WALLABAG_`. I can also add others like all database configuration if you like the way to do it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -85,6 +85,9 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
+        ],
+        "build-parameters": [
+            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters"
         ]
     },
     "extra": {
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -85,6 +85,9 @@
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
+        ],
+        "build-parameters": [
```
j0k3r: Why do you need that? `builParameters` is already called when you perform a `composer install` (or an update)

j0k3r:
I'm not so familiar with `env-map` on `incenteev/composer-parameter-handler` but does it means it overrides the config value _only_ if the global variable exists ?

mathbruyen:
It seems that's what it does. I've never used it before but I did a try without the environment variable set and it either took the default value (in non-interfactive mode) or asks for a value to be given (in interactive mode). The [documentation](https://github.com/Incenteev/ParameterHandler#using-environment-variables-to-set-the-parameters) is not very explicit about that indeed.

composer.json
mathbruyen: I created it for being able to have a hook only generating configuration based on the environment, which I call [before actually starting wallabag](https://github.com/mathbruyen/home/blob/master/wallabag/Dockerfile#L19). I trigger the install fetching dependencies while generating the container but I only _patch_ the configuration from the environment.

That's maybe not the best way to do, I've never used composer before.

j0k3r:
I'll merge your stuff and we'll see in the future if we need to add more configuration.
Thanks for that :rocket:

PR #1529, opened 2015-12-08
changed table name for tagging rule

nicosomb:
ping @K-Phoen


j0k3r:
What was the generated name before that change?

nicosomb:
`wallabag_TaggingRule`

j0k3r:
What is the generated name for the tag user table? Is it `tag_user`? If yes, we should name that table `tagging_rule` instead.

nicosomb:
`wallabag_tag`, `wallabag_entry_tag`, etc.

You're right, I fix it.

PR #1535, opened 2015-12-21
Add popehat.com

juanjobe:


j0k3r:
It seems to work great without that file (at least on wallabag v2). Did you test using wallabag v1 or v2?

juanjobe:
I used wallabag v1

PR #1537, opened 2015-12-23
Updated composer.lock

bdunogier:
> Updates `composer.lock`.

Had to restrict scheb/two-factor-bundle to ~1.4.0 because of a circular service reference.
See https://github.com/scheb/two-factor-bundle/issues/33.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-    - 5.4
     - 5.5
     - 5.6
     - 7.0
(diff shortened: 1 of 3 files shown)


j0k3r:
Oh yes, Rulerz now requires PHP 5.5, could you update Travis & composer.json with that information?

bdunogier:
> Oh yes, Rulerz now requires PHP 5.5, could you update Travis & composer.json with that information?

Sure.

j0k3r:
You forgot to remove PHP 5.4 from .travis.yml file :)

j0k3r:
:+1:

PR #1545, opened 2015-12-29
v2 – Remove user reference in tag

### LINKED ISSUE #1543: v2 – Remove user reference in Tag entry ?
Since we've added a slug in the Tag entry which is unique, we got awful slug when user use the same tag label.

Here is an example with some fixtures:

id | user_id | label | slug
------------ | ------------ | ------------ | ------------
1 | 2 | foo | foo
2 | 2 | bar | bar
3 | 1 | foo | foo-1
4 | 1 | bar | bar-1

As you can see, we got `foo-1` as slug, it can goes pretty far, since we can then have `foo-4893` for a slug. I don't really that.

That's why I'm wondering if we can remove the user assignation in that entry.

1. If we need to retrieve all tags for a user, we can join over Entry & EntryTag to retrieve them
2. We will have unique slug with a cleaner value
3. Removing a tag for user will only have to remove lines in EntryTag

What do you think?

j0k3r:
Fix #1543

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -337,10 +337,6 @@ class WallabagRestController extends FOSRestController
             ->getRepository('WallabagCoreBundle:Entry')
             ->removeTag($this->getUser()->getId(), $tag);
 
-        $em = $this->getDoctrine()->getManager();
-        $em->remove($tag);
-        $em->flush();
-
         $json = $this->get('serializer')->serialize($tag, 'json');
 
         return $this->renderJsonResponse($json);
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -27,7 +27,7 @@ class WallabagRestController extends FOSRestController
                 ->findOneByLabel($label);
 
             if (is_null($tagEntity)) {
-                $tagEntity = new Tag($this->getUser());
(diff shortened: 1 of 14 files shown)


src/Wallabag/ApiBundle/Controller/WallabagRestController.php
```diff
@@ -175,8 +174,8 @@ public function patchEntriesAction(Entry $entry, Request $request)
         $this->validateUserAccess($entry->getUser()->getId());
 
         $title = $request->request->get('title');
-        $isArchived = $request->request->get('is_archived');
-        $isStarred = $request->request->get('is_starred');
+        $isArchived = $request->request->get('archive');
```
nicosomb: oh, well done. this bug is here since a long time...

src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -179,4 +180,53 @@ public function findOneByUsernameAndNotArchived($username)
...
+            ->getQuery()
+            ->getResult();
+
+        foreach ($entries as $entry) {
+            $entry->removeTag($tag);
+        }
+
+        $this->getEntityManager()->flush();
+
+        // An other solution can be to use raw query but I can't find a way to retrieve the `entry_tag` table name since it can be prefixed....
```
nicosomb: Maybe the php doc before the function is necessary and we can remove these lines?

j0k3r: Yeah sure, I'll add the SQL query inside the php doc

PR #1546, opened 2015-12-30
v2 – Ability to reload an entry

### LINKED ISSUE #1503: [v2] reload article content
Button on article page to re-fetch article content.

j0k3r:
Could be useful when we want to update the content or when the content failed to be fetched.

Fix #1503 

Material menu:

image

Baggy menu:

image

I wasn't able to find an icon for baggy, so I put the whole text instead. If anyone can convert it to a nice icon, it could be great !


tcitworld:
On v1 the reload is done with Icomoon CSS classes https://github.com/wallabag/wallabag/blob/dev/themes/baggy/view.twig

j0k3r:
@tcitworld oh great :+1: 
But I needed to back port all icomoon from v1.

PR #1553, opened 2016-01-06
Hide export button if we are not on entries list

nicosomb:


tcitworld:
Isn't something missing here ?

nicosomb:
No, why?

tcitworld:
I didn't have or go or take a deep look, but won't this hide it every time ?

nicosomb:
We have this in init.js (`init_export()`): 

```
    if ($("div").is("#export")) {
        $('#button_export').show();
        $('.button-collapse-right').sideNav({ edge: 'right' });
    }
```

tcitworld:
Right. Your change should have been there from start. :-)

PR #1560, opened 2016-01-09
Fix #1501 Quickstart for beginners

### LINKED ISSUE #1501: [v2] Welcome page when article list is empty
- [ ] Link to documentation
- [ ] Link to importers
- [ ] Link to config

What else?

nicosomb:
Fix #1501 

- [x] create quickstart for baggy theme

- [x] ~~flag for each user to know if he already read the quickstart ?~~

- [x] Test to know if quickstart is displayed when the user has no article (and adapt it with the previous line if necessary)

Some URL are wrong, I know that. We'll change it when documentation will be finished and published.

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -166,6 +166,11 @@ class EntryController extends Controller
      */
     public function showUnreadAction(Request $request, $page)
     {
+        // load the quickstart if no entry in database
+        if ($page == 1 && $this->get('wallabag_core.entry_repository')->countAllEntriesByUsername($this->getUser()->getId()) == 0) {
+            return $this->redirect($this->generateUrl('quickstart'));
+        }
+
         return $this->showEntries('unread', $request, $page);
     }
 
@@ -211,7 +216,7 @@ class EntryController extends Controller
      */
     private function showEntries($type, Request $request, $page)
     {
-        $repository = $this->getDoctrine()->getRepository('WallabagCoreBundle:Entry');
+        $repository = $this->get('wallabag_core.entry_repository');
 
         switch ($type) {
             case 'starred':
--- a/src/Wallabag/CoreBundle/Repository/EntryRepository.php
+++ b/src/Wallabag/CoreBundle/Repository/EntryRepository.php
@@ -122,6 +122,8 @@ class EntryRepository extends EntityRepository
     /**
      * Fetch an entry with a tag. Only used for tests.
      *
+     * @param int $userId
+     *
      * @return Entry
      */
     public function findOneWithTags($userId)
@@ -247,5 +249,22 @@ class EntryRepository extends EntityRepository
         }
 
         return false;
+    }
+
+    /**
+     * Count all entries for a user.
+     *
+     * @param int $userId
+     *
+     * @return integer
+     */
+    public function countAllEntriesByUsername($userId)
+    {
+        $qb = $this->createQueryBuilder('e')
+            ->select('count(e)')
+            ->where('e.user=:userId')->setParameter('userId', $userId)
+        ;
+
+        return $qb->getQuery()->getSingleScalarResult();
     }
 }
--- a/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
+++ b/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
@@ -19,6 +19,36 @@ class EntryControllerTest extends WallabagCoreTestCase
         $this->assertContains('login', $client->getResponse()->headers->get('location'));
     }
 
+    public function testQuickstart()
+    {
+        $this->logInAs('empty');
+        $client = $this->getClient();
+
+        $client->request('GET', '/unread/list');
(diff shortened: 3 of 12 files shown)


src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -250,6 +250,10 @@ private function showEntries($type, Request $request, $page)
         $entries->setMaxPerPage($this->getUser()->getConfig()->getItemsPerPage());
         $entries->setCurrentPage($page);
 
+        if ($type === 'unread' && count($entries) == 0) {
```
nicosomb: @j0k3r not sure if this test is at the good place.

j0k3r: Well, I'll prefer to see that in `showUnreadAction` with a test on the route and an other query to retrieve the count on **all** entries (not only the unread one)

nicosomb:
Any opinion about 

> flag for each user to know if he already read the quickstart ?

j0k3r:
@nicosomb I'm :-1: for the extra flag. A simple query to count all entries is enough for me.

nicosomb:
Oh yes, because we'll count all entries. You're right :+1:

nicosomb:
@j0k3r with data loaded in tests, we don't have an empty database to test if quickstart is loaded when count($entries) = 0. What can I do? Empty database in my test and reload articles in the same test?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -401,4 +414,17 @@ private function checkUserAction(Entry $entry)
...
+
+    /**
+     * @return Pagerfanta
+     */
+    public function _countAllEntriesForUser()
+    {
+        $repository = $this->getDoctrine()->getRepository('WallabagCoreBundle:Entry');
+        $qb = $repository->getBuilderForAllByUser($this->getUser()->getId());
+        $pagerAdapter = new DoctrineORMAdapter($qb->getQuery());
+        $entries = new Pagerfanta($pagerAdapter);
```
j0k3r: No, no no. Use a simple count SQL query with Doctrine (using the repo of course) we don't need to load the full pager etc for a simple count.

j0k3r:
You can add a new user which will have no entry.

src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php
```diff
@@ -19,6 +19,36 @@ public function testLogin()
...
+
+        // Test if quickstart is disabled when user has 1 entry
+        $crawler = $client->request('GET', '/new');
+
+        $this->assertEquals(200, $client->getResponse()->getStatusCode());
+
+        $form = $crawler->filter('button[type=submit]')->form();
+
+        $data = array(
+            'entry[url]' => 'https://www.wallabag.org/blog/2016/01/08/wallabag-alpha1-v2',
```
nicosomb: If I use `$this->url` here, I've got this error further, and I don't know why (`arpenter` is in article content):

```
1) Wallabag\CoreBundle\Tests\Controller\EntryControllerTest::testPostNewOk
Failed asserting that 'arpenter' contains "Google".
/home/nicosomb/dev/wallabag/v2/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php:134
```
Related to #1566?

src/Wallabag/CoreBundle/Controller/EntryController.php
j0k3r: Do we still need this condition ?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -401,4 +410,14 @@ private function checkUserAction(Entry $entry)
...
     }
+
+    /**
+     * @return int
+     */
+    private function _countAllEntriesForUser()
+    {
+        $repository = $this->getDoctrine()->getRepository('WallabagCoreBundle:Entry');
+
+        return $repository->countAllEntriesByUsername($this->getUser()->getId());
```
j0k3r: Since you are using that method only once, you can move it inside `showUnreadAction`

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -166,6 +166,13 @@ public function showAllAction(Request $request, $page)
      */
     public function showUnreadAction(Request $request, $page)
     {
+        // load the quickstart if no entry in database
+        $repository = $this->getDoctrine()->getRepository('WallabagCoreBundle:Entry');
```
j0k3r: Why not using the service?

src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -248,4 +250,21 @@ public function existByUrlAndUserId($url, $userId)
 
         return false;
     }
+
+    /**
+     * Count all entries for a user.
+     *
+     * @param int $userId
+     *
+     * @return Entry
```
j0k3r: Nope, it returns an integer

PR #1564, opened 2016-01-09
Remove Flattr references

nicosomb:
As we don't want to use this service (they changed their website, we lost features, etc.), I think it's better to remove flattr into our application.


tcitworld:
If their API still works, I don't think we should remove it, as a non negligible number of users may still use it. We'll move it at a plugin when it will be possible.

j0k3r:
So, do we remove it?

nicosomb:
> If their API still works

I don't really know what they changed ... I don't know if their API still works.

nicosomb:
I only removed flattr references on about page.

j0k3r:
You can also remove `flattr_url` from `app/config/config.yml`

PR #1574, opened 2016-01-13
Fix typo and set version in the README

jjanvier:
Setting the version will ensure people will have a functionnal app no matter what you do on the master v2 branch :)


j0k3r:
Good idea ! :+1:

PR #1581, opened 2016-01-15
[VERY WIP] V2 asynchronous jobs

nicosomb:
need to be rewriten, but it works.

TODO 

- [x] rewrite some ugly parts ...
- [x] adapt docker-compose.yml 
- [x] add a test to prevent RabbitMQ publication if RabbitMQ it not configured
- [ ] on entries view, we need to hide empty entries (?) and indicates that asynchronous tasks are processing
- [ ] create status table to store information about rabbitmq processing (useful to display a progress bar in wallabag)
- [ ] tests
- [x] documentation

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -213,3 +213,28 @@ lexik_maintenance:
     response:
         code: 503
         status: "wallabag Service Temporarily Unavailable"
+
+old_sound_rabbit_mq:
+    connections:
+        default:
+            host:     %rabbitmq_host%
+            port:     %rabbitmq_port%
+            user:     %rabbitmq_user%
+            password: %rabbitmq_password%
+            vhost:    /
+            lazy:     false
+    producers:
+        wallabag:
+            connection: default
+            exchange_options:
+                name: 'wallabag_exchange'
+                type: topic
+    consumers:
+        entries:
+            connection: default
+            exchange_options:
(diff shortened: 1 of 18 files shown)


app/config/config.yml
```diff
@@ -213,3 +213,28 @@ scheb_two_factor:
...
+            vhost:    /
+            lazy:     false
+    producers:
+        foo:
+            connection: default
+            exchange_options:
+                name: 'foo_exchange'
+                type: topic
+    consumers:
+        bar:
```
j0k3r: :trollface:

Simounet:
Hello,
I tried to import my Pocket's items through RabbitMQ. I had to launch the importer many times. There is maybe a timeout setting from Rabbit that I don't know much about.
Anyway, I get 5138 reads items + 2 unreads. The Pocket API tell me that I have 6967 reads and 26 unreads. If I try to play the importer again, I get « Import summary: 0 imported, 5000 already saved. ».

Hope this will help.

j0k3r:
Yeah, for the moment we only get the first 5k articles, we didn't paginated them to retrieve all articles.
We'll have to implement that.
Thanks for your feedback !

Simounet:
Ok, it would be great to get unreads first.

j0k3r:
Yep, I'l read your tweet about that :+1:

PR #1606, opened 2016-01-20
Remove tag relation when removing an entry

### LINKED ISSUE #1453: [v2] tags are not deleted when you delete an entry
I thought it had been taken of. If not, my bad.

* Get an article
* Set a tag to it
* Delete the article
* Get a new article right after (so that it has the same id)
* The new article has the old tag

j0k3r:
Fix #1453

When removing an entry, it'll also remove all associated tag and avoid crazy stuff like defined in the previous issue.

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/Entry.php
+++ b/src/Wallabag/CoreBundle/Entity/Entry.php
@@ -168,7 +168,7 @@ class Entry
     private $user;
 
     /**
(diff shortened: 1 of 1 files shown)


nicosomb:
Why did Scrutinizer failed with a timeout?

j0k3r:
@nicosomb I relaunched it, it's ok now.

PR #1607, opened 2016-01-20
Ignore composer.lock

j0k3r:
Having a big composer.lock on a final project can have side effect on incoming PR that add a new vendor.
Mostly because conflict are too frequent.

By ignoring composer.lock we ease the PR submission and rebase.

BUT we need to be careful when we release a new version of wallabag. We should manually `git add -f composer.lock` to update it.

Since composer.lock will no longer be commited I switch the `composer install` to a `composer up` in the travis configuration.

/ping @bdunogier what do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ data/db/wallabag*.sqlite
 # Docker container logs and data
 docker/logs/
 docker/data/
+
+# To avoid crazy stuff on some PR, we must manually FORCE ADD IT on each new release
+composer.lock
--- a/build.xml
+++ b/build.xml
@@ -11,7 +11,7 @@
 
     <target name="composer" description="Install deps using Composer">
         <exec executable="composer">
-            <arg value="install"/>
+            <arg value="up"/>
             <arg value="--no-interaction"/>
             <arg value="--no-progress"/>
         </exec>


bdunogier:
Great, it will indeed avoid confusion.

About releases, you could also unignore ´composer.lock´ in the release branch (if you use one).

nicosomb:
About releases, @bdunogier, I was serious on gitter, you can help us on this point (write a releasing documentation,etc.)

bdunogier:
> Since composer.lock will no longer be commited I switch the composer install to a composer up in the travis configuration.

Actually, you don't need that. If there's no lock, `install` == `update`. If you switch to update, it will prevent `composer.lock` from being used when it gets added, on a release branch or in a PR for any reason (if that applies to the build).

> About releases, @bdunogier, I was serious on gitter, you can help us on this point (write a releasing documentation,etc.)

Okay, I'll throw some notes down, we'll see where it goes. Do you have anything about how you currently proceed, what deliverables (tarballs, maybe, etc) you have in mind ?

bdunogier:
Shouldn't you also remove `composer.lock` from v2 in this PR ?

j0k3r:
Yeah can do that. But we need to increase memory for travis 
> [exec] Fatal error: Allowed memory size of 1073741824 bytes exhausted (tried to allocate 134217728 bytes) in phar:///home/travis/.phpenv/versions/5.6.5/bin/composer/src/Composer/DependencyResolver/Solver.php on line 214

bdunogier:
Damnit. Is `xdebug` disabled for composer ?

bdunogier:
About installation, I have drafted that: https://gist.github.com/bdunogier/ecd672322b676fd5790f

j0k3r:
@nicosomb why did your merge that? We still need to figure out how to increase composer memory and remove composer.lock.
Also we need to talk about the first draft of the release process.

@bdunogier About xdebug, since we don't use coverage because it takes 20 minutes to build with it, we might try to remove it.

nicosomb:
@j0k3r grrr, sorry. Revert?

bdunogier:
> we might try to remove it.

Unless I am mistaken, it is already disabled (maybe not on all versions of PHP ?).

I know that with 90 deps, we (eZ) are very close from the 1gb limit. Wallabag has above 110 :)

j0k3r:
I'll create a new PR

PR #1610, opened 2016-01-21
Disable xdebug all the time

j0k3r:
_Discussion started here https://github.com/wallabag/wallabag/pull/1607_

Use travis_wait to avoid travis timeout on `composer up`
Jump to `memory_limit -1` for `composer up`
Also, since composer.lock is ignored, we can remove it.


j0k3r:
@bdunogier memory_limit :white_check_mark:

bdunogier:
Good old recipes do work ! :)

+1

PR #1612, opened 2016-01-21
Settings page

### LINKED ISSUE #1575: [v2] Ease installation at maximum
Hello wallabag team. Thanks for your great app :) The v2 is amazing :)

To ease installation for beginners/newcomers I think it would relevant to delete a lot of parameters that are asked during the process. Personnally I found the install process a little bit too long and I didn't find useful to have all these parameters. 

I think the following parameters should be set directly in the `config.yml`:
```
    twofactor_auth: true
    twofactor_sender: no-reply@wallabag.org
    warning_message: "You're trying wallabag v2, which is in alpha version. If you find a bug, please have a look to <a href=\"https://github.com/wallabag/wallabag/issues\">our issues list<
/a> and <a href=\"https://github.com/wallabag/wallabag/issues/new\">open a new if necessary</a>\n"
    download_pictures: true
    share_twitter: true
    share_mail: true
    share_shaarli: true
    shaarli_url: 'http://myshaarli.com'
    share_diaspora: false
    diaspora_url: 'http://diasporapod.com'
    flattr: true
    carrot: true
    show_printlink: true
    export_epub: true
    export_mobi: true
    export_pdf: true
    wallabag_url: 'http://v2.wallabag.org'
    wallabag_support_url: 'https://www.wallabag.org/pages/support.html'
    items_on_page: 12
    theme: material
    rss_limit: 50
    pocket_consumer_key: xxxxxxxx
```

Also the test database parameters could be dropped from the installation.

And IMHO, all these parameters could be in a wallabag section of the config.yml instead of being root parameters. That would help to clearly identify what can be customized for people that install the app.

I can do a PR if you're interested.

j0k3r:
Fix #1575 

This PR add an _internal config_ page where admin can configure its wallabag instance. Most of the parameters from the `parameters.yml.dist` were moved to this page as they can be switched during run time.

To achieve that I've used the [CraueConfigBundle](https://github.com/craue/CraueConfigBundle). Each parameter has a section which allow to have a less disgusting _internal config_ page.
This page is only accessible by user with role `ROLE_SUPER_ADMIN`.

This table is filled using the InstallCommand (at the same time we build the user config).

Things to check before validate this PR:

- [x] Translation for each parameters
- [x] Section for each parameters

Preview:

image

_Bonus_

I've added a message on the Pocket import page when the consumer key isn't defined:
> Pocket import isn't configured. You need to define pocket_consumer_key.


app/config/security.yml
```diff
@@ -58,3 +58,4 @@ security:
         - { path: ^/resetting, role: IS_AUTHENTICATED_ANONYMOUSLY }
         - { path: /(unread|starred|archive).xml$, roles: IS_AUTHENTICATED_ANONYMOUSLY }
         - { path: ^/, roles: ROLE_USER }
+        - { path: ^/settings, roles: ROLE_SUPER_ADMIN }
```
nicosomb: With my ROLE_USER user, I can access to this URL.

app/config/config.yml
```diff
@@ -41,25 +47,6 @@ wallabag_import:
 twig:
     debug:            "%kernel.debug%"
     strict_variables: "%kernel.debug%"
-    globals:
```
bdunogier: nice removal :-)

src/Wallabag/CoreBundle/Command/InstallCommand.php
```diff
@@ -205,13 +206,102 @@ protected function setupAdmin()
...
+        $config->setItemsPerPage($this->getContainer()->getParameter('wallabag_core.items_on_page'));
+        $config->setRssLimit($this->getContainer()->getParameter('wallabag_core.rss_limit'));
+        $config->setLanguage($this->getContainer()->getParameter('wallabag_core.language'));
 
         $em->persist($config);
 
+        // cleanup before insert new stuff
+        $em->createQuery('DELETE FROM CraueConfigBundle:Setting')->execute();
+
+        $settings = [
```
bdunogier: These could come from an external source, so that the installer doesn't have to modified each time, could it not ?

j0k3r: Yep just need to figure out the _external source_

app/config/security.yml
j0k3r: Oh yeah I haven't tested that case. It's done.

nicosomb:
Still conflicts due to the last merged PR.

src/Wallabag/CoreBundle/Command/InstallCommand.php
nicosomb: About the external source, we'll do that after or in this PR?

j0k3r: We can do that here. Which solution ?

tcitworld:
Don't forget to update parameters according to #1629. :)

j0k3r:
@tcitworld updated.

tcitworld:
You delete the `app.version` parameter but it isn't put elsewhere. Is this intended ?

j0k3r:
@tcitworld It's sent to the template [here](https://github.com/wallabag/wallabag/pull/1612/files#diff-c422f101c02db8dafa81cae2d775997eR28) and I don't have to update the template because the `version` var has the same name (see for [material](https://github.com/wallabag/wallabag/blob/v2/src/Wallabag/CoreBundle/Resources/views/themes/material/Static/about.html.twig#L29) & [baggy](https://github.com/wallabag/wallabag/blob/v2/src/Wallabag/CoreBundle/Resources/views/themes/baggy/Static/about.html.twig#L21)).

app/Resources/CraueConfigBundle/translations/CraueConfigBundle.fr.yml
```diff
@@ -0,0 +1,19 @@
...
+diaspora_url: URL de Diaspora, si le service Diaspora est activé
+export_epub: Activer l'export ePub
+export_mobi: Activer l'export .mobi
+export_pdf: Activer l'export PDF
+export_csv: Activer l'export CSV
+export_json: Activer l'export JSON
+export_txt: Activer l'export TXT
+export_xml: Activer l'export XML
+pocket_consumer_key: Clé d'authentification Pocket pour importer les données (https://getpocket.com/developer/docs/authentication)
+shaarli_url: URL de Shaarli, si le service Diaspora est activé
```
nicosomb: `si le service Diaspora est activé` > `si le service est activé`

PR #1614, opened 2016-01-22
Few fixes

j0k3r:
- Use PHP7 `random_bytes` to generate RSS Token
- Add tests on `TagAllCommand`
- Fix bad redirection when adding a new user
- Fix bad type after using `findByUrlAndUserId`
- Use `FOSUserEvents` instead of c/p a controller
- Add for deleting rule from an other user


nicosomb:
:+1: (for every PR like this, I say "it's very good having a @j0k3r in his team")

j0k3r:
:pray:

PR #1616, opened 2016-01-22
[DON'T MERGE] Release wallabag 2.0.0-alpha.2

nicosomb:


README.md
```diff
@@ -15,7 +15,7 @@ Keep in mind it's an **unstable** branch, everything can be broken :)
 If you don't have it yet, please [install composer](https://getcomposer.org/download/). Then you can install wallabag by executing the following commands:
 
 ```
-composer create-project wallabag/wallabag wallabag 2.0.0-alpha.1
+composer create-project wallabag/wallabag wallabag 2.0.*@dev
```
bdunogier: I'd use `@alpha`. It is more accurate, and should also work for the beta.
But if at some point a `branch-alias` is added to wallabag/wallabag@v2, then this restriction would install the v2 branch directly for everybody instead of the tag.

PR #1647, opened 2016-02-04
Force prod env on project creation

### LINKED ISSUE #1617: Ability to install wallabag without dev deps
When using `composer create-project`, users will have the full deps (including dev one, like PHPUnit). They don't need them.

We should be able to launch `composer create-project [...] --no-dev`. But some bundles in `require-dev` are required in non-dev.

> PHP Fatal error: Class 'Sensio\Bundle\GeneratorBundle\SensioGeneratorBundle' not found in /private/tmp/wallabag/app/AppKernel.php on line 41

j0k3r:
So the only solution I found so far is to force the `prod` env when create the project with composer.
There is other solution.

Since the `cache:clear` task is using `dev` env by default, so it tries to load dev bundle, which aren't there since we specify `--no-dev`.

----

Fix #1617 

/ping @bdunogier 
/see https://github.com/sensiolabs/SensioDistributionBundle/pull/140

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Keep in mind it's an **unstable** branch, everything can be broken :)
 If you don't have it yet, please [install composer](https://getcomposer.org/download/). Then you can install wallabag by executing the following commands:
 
 ```
-composer create-project wallabag/wallabag wallabag 2.0.*@alpha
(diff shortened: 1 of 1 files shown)


bdunogier:
I don't really have a better way for now. +1

PR #1651, opened 2016-02-05
Test if some keys are missing in v2 import

nicosomb:


j0k3r:
Please add a test case that tests the problem this PR is fixing.

nicosomb:
Is that enough if I update fixtures?

j0k3r:
Well, it's not enough. We are losing test when `mimetype` and others fields are present.

PR #1655, opened 2016-02-08
Add 'content' key support in API

### LINKED ISSUE #1634: Pass content through API when creating an entry
It would be useful to be able to send the content when creating an entry using the REST API, as well as being able to set the Starred and Archived states.

tcitworld:
Will fix #1634
Maybe we should check content before saving it, what do you think ?

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -99,6 +99,7 @@ class WallabagRestController extends FOSRestController
      *          {"name"="tags", "dataType"="string", "required"=false, "format"="tag1,tag2,tag3", "description"="a comma-separated list of tags."},
      *          {"name"="starred", "dataType"="boolean", "required"=false, "format"="true or false", "description"="entry already starred"},
      *          {"name"="archive", "dataType"="boolean", "required"=false, "format"="true or false", "description"="entry already archived"},
+     *          {"name"="content", "dataType"="string", "required"=false, "format"="content", "description"="content you want to pass directly"},
      *       }
      * )
      *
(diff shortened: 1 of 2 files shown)


src/Wallabag/ApiBundle/Controller/WallabagRestController.php
```diff
@@ -143,8 +149,21 @@ public function postEntriesAction(Request $request)
             $this->assignTagsToEntry($entry, $tags);
         }
 
+        if (!empty($isStarred)) {
```
j0k3r: Instead of `!empty(` I would rather check the trueness of the value. Like:

```php
if (true === (bool) $isStarred) {
    $entry->setStarred(true);
}
```

Since `isArchived` & `isStarred` are false by default.

j0k3r:
Good PR with tests :ok_hand: 
And yes I think we should check content before saving it. But I'm wondering how can we check without altering the data..

j0k3r:
I prefer to leave it a WIP because we really need to find a solution to sanitize the content to avoid bad things.

nicosomb:
> I prefer to leave it a WIP because we really need to find a solution to sanitize the content to avoid bad things.

What do you mean? Graby enhancement or wallabag one?

j0k3r:
Maybe a wallabag one.
We that endpoint we allow any user to store raw content in the database which is dangerous.

j0k3r:
So I've splitted this PR so only the content part is kept here. See https://github.com/wallabag/wallabag/pull/1754

Until we find a good solution to sanitize the content.

j0k3r:
Woops.

nicosomb:
> Until we find a good solution to sanitize the content.

Need to open an issue?

j0k3r:
In this PR it's ok

PR #1656, opened 2016-02-08
fix labels on new user prompt

tcitworld:
This shows proper labels on the new user creation panel.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/translations/validators.fr.yml
+++ b/src/Wallabag/CoreBundle/Resources/translations/validators.fr.yml
@@ -1,0 +1,3 @@
+# Config screen
+The password fields must match: 'Les deux mots de passe doivent être les mêmes'
+Password should by at least 8 chars long: 'Le mot de passe doit contenir au moins 8 caractères'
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Resources/translations/validators.fr.yml
```diff
@@ -0,0 +1,2 @@
+# Config screen
+The password fields must match: 'Les deux mots de passe doivent être les mêmes'
```
tcitworld: Does french still has those accents ? Just saying.

j0k3r: :trollface: :100:

j0k3r:
What was the problem?

tcitworld:
The problem was that the two passwords label were simply *First* and *Second*. Also, the error message wasn't translated.

I realize the length constraint on the password doesn't seem to work. Looking into it.

src/Wallabag/CoreBundle/Controller/ConfigController.php
```diff
@@ -124,7 +124,6 @@ public function indexAction(Request $request)
         // enable created user by default
         $newUser->setEnabled(true);
         $newUserForm = $this->createForm(NewUserType::class, $newUser, array(
-            'validation_groups' => array('Profile'),
```
j0k3r: I think it's still required for FOSUser

tcitworld: Seems so. But the length assertion still fails.

j0k3r:
Could you squash your commits?

Issue #1657, opened 2016-02-09
[v2] Wallabag v1 import ignores tags


jcharaoui:
Entry tags are completely ignored when importing from Wallabag v1.

j0k3r:
As far as I remember, tags aren't exported in wallabag v1...
Here is [a fixture we use for tests unit in v2](https://github.com/wallabag/wallabag/blob/v2/src/Wallabag/ImportBundle/Tests/fixtures/wallabag-v1.json), this file was exported by wallabag v1.

Maybe @tcitworld good an idea about that?

tcitworld:
Tags are indeed exported from v1 :
```
{
"id":"1","0":"1",
"title":"Framabag, un nouveau service libre et gratuit","1":"Framabag, un nouveau service libre et gratuit",
"url":"http:\/\/www.framablog.org\/index.php\/post\/2014\/02\/05\/Framabag-service-libre-gratuit-interview-developpeur","2":"http:\/\/www.framablog.org\/index.php\/post\/2014\/02\/05\/Framabag-service-libre-gratuit-interview-developpeur",
"is_read":"1","3":"1",
"is_fav":"0","4":"0",
[...]

PR #1658, opened 2016-02-09
import tags from v1 (#1657)

tcitworld:
#1657
Maybe a few more tests ?


src/Wallabag/ImportBundle/Import/WallabagV1Import.php
```diff
@@ -141,6 +142,14 @@ protected function parseEntries($entries)
             $entry->setArchived($importedEntry['is_read']);
             $entry->setStarred($importedEntry['is_fav']);
             $entry->setContent($importedEntry['content']);
+            if ($importedEntry['tags'] != '') {
+                $tags = explode(',', $importedEntry['tags']);
+                foreach ($tags as $tagv1) {
+                    $tag = new Tag();
```
j0k3r: You should check if tag exist first. See https://github.com/wallabag/wallabag/blob/v2/src/Wallabag/ImportBundle/Import/PocketImport.php#L180

We need this method to be called from (almost) everywhere since we already use it in ContentProxy.

tcitworld: > Todo: We need this method to be called from (almost) everywhere since we already use it in ContentProxy.

assignTagsToEntry ?

j0k3r: Yeah. I once set a todo but I guess @nicosomb removed it

j0k3r: As a followup https://github.com/wallabag/wallabag/issues/1667

j0k3r:
Yeah we might need one more test in `WallabagV1ControllerTest`.

You should retrieve with that url `http://www.framablog.org/index.php/post/2014/02/05/Framabag-service-libre-gratuit-interview-developpeur` (ie: the only one with a tag) and check that it has the associated tag.

src/Wallabag/ImportBundle/Tests/Controller/WallabagV1ControllerTest.php
```diff
@@ -39,6 +39,21 @@ public function testImportWallabagWithFile()
...
+            ->getRepository('WallabagCoreBundle:Entry')
+            ->findByUrlAndUserId(
+                'http://www.framablog.org/index.php/post/2014/02/05/Framabag-service-libre-gratuit-interview-developpeur',
+                $this->getLoggedInUserId()
+            );
+
+        $tag = $client->getContainer()
+            ->get('doctrine.orm.entity_manager')
+            ->getRepository('WallabagCoreBundle:Tag')
+            ->findOneByTagLabel('Framabag');
```
j0k3r: Could you try with `findOneByLabel` ? I guess this magic method exist

tcitworld: It does indeed.

j0k3r:
Great ! :+1:

PR #1666, opened 2016-02-11
reimport v1 entries if they were not fetched

tcitworld:
If an article was not properly fetched on v1, this will reimport it.

About tests, does adding a untitled entry in wallabag-v1.json is enough ?


src/Wallabag/ImportBundle/Import/WallabagV1Import.php
```diff
@@ -123,6 +126,10 @@ public function setFilepath($filepath)
     protected function parseEntries($entries)
     {
         $i = 1;
+        /*
+         * Untitled in all languages from v1. This should never have been translated
+         */
```
j0k3r: You can put the comment on one line starting with `//`

src/Wallabag/ImportBundle/Import/WallabagV1Import.php
```diff
@@ -123,6 +126,10 @@ public function setFilepath($filepath)
...
+        $untitled = array('Untitled', 'Sans titre', 'podle nadpisu', 'Sin título', 'با عنوان', 'per titolo', 'Sem título', 'Без названия', 'po naslovu', 'Без назви');
```
j0k3r: Since the v2 import will also get that, can't you also include an empty string ?

j0k3r:
> About tests, does adding a untitled entry in wallabag-v1.json is enough ?

You can add both for the v1 & v2 import file

src/Wallabag/ImportBundle/Import/WallabagV1Import.php
tcitworld: Don't you mean `No title found` ? In regard to https://github.com/j0k3r/graby/pull/42 ?

j0k3r: Maybe both.

j0k3r:
:+1:

PR #1671, opened 2016-02-14
display pictures properly

tcitworld:
So that pictures don't go too far on the right.


j0k3r:
Oh thanks I hated this behavior !
Also why don't the article use the full width like it was on baggy?

tcitworld:
Full width is certainly too much to read in good conditions but we could increase it a bit.

PR #1683, opened 2016-02-17
Demonstration mode to prevent password changes

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -46,5 +46,5 @@ parameters:
 
     from_email: no-reply@wallabag.org
 
-    # set to true to prevent changing password
+    # this parameter is only used for the wallabag public demo, it means user won't be able to change the password of the default account
     demo: false
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -46,5 +46,6 @@ parameters:
 
     from_email: no-reply@wallabag.org
 
-    # this parameter is only used for the wallabag public demo, it means user won't be able to change the password of the default account
+    # this parameter is only used for the wallabag public demo, it means user won't be able to change the password of the demo_user account
     demo: false
+    demo_username: wallabag
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -45,3 +45,6 @@ parameters:
     fosuser_confirmation: true
 
     from_email: no-reply@wallabag.org
+
+    # set to true to prevent changing password
+    demo: false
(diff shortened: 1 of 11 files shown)


src/Wallabag/CoreBundle/Controller/ConfigController.php
```diff
@@ -57,13 +57,20 @@ public function indexAction(Request $request)
...
+                $userManager->updateUser($user, true);
+
+                $this->get('session')->getFlashBag()->add(
+                    'notice',
+                    'Password updated'
+                );
+            } else {
+                $this->get('session')->getFlashBag()->add(
+                    'notice',
+                    'In demonstration mode, you can`t change password.'
```
j0k3r: What that `, can't you use \' ?

nicosomb: my fingers slipped

app/config/parameters.yml.dist
```diff
@@ -45,3 +45,6 @@ parameters:
     fosuser_confirmation: true
 
     from_email: no-reply@wallabag.org
+
+    # set to true to prevent changing password
```
j0k3r: I would rather put :
> this parameter is only used for the wallabag public demo, it means user won't be able to change the password of the default account

nicosomb: :+1: you're right.

nicosomb: or not in fact. User**s** won't be abloe to change password of all accounts.

tcitworld:
Wait, this means we can't modify passwords for user accounts on v2.wallabag.org. It should be only the `wallabag` account that should be blocked.

nicosomb:
v2.wallabag.org is just a demo platform ...

tcitworld:
Yeah but we did kind of push people to subscribe, even if we warned them it was only testing. Furthermore, maybe administrators of wallabag instances may want to propose a demo account to their customers.

j0k3r:
You mean, it could more be a role instead ?

tcitworld:
Or could we have a second parameter : the username of the demo account (I guess one is enough ?)

nicosomb:
> Or could we have a second parameter : the username of the demo account (I guess one is enough ?)

Maybe the easier solution?

j0k3r:
I'm wondering. Should this configuration be enable only using internal config instead? End user won't need to define this information when installing the app.
So moving the config in CraueConfig might be easier to test.

nicosomb:
:+1:

Issue #18, opened 2013-04-05
Search engine


nicosomb:

4oo4:
I see this got removed from the 2.1.0 milestone, any update on where this is at? I would gladly help if I'm able to assist.

Cheers for all the other improvements in v2.1

j0k3r:
Most of the job was done in https://github.com/wallabag/wallabag/pull/1685 but it was closed when we renamed branches.
It can be a good starting point if you want to help us on that.
If we can have it up and running for the 2.2.0 it could be great!

4oo4:
Cool, so should I just pull from the 2.2.0 branch and go from there? Also, is elasticsearch still the backend?

j0k3r:
Yes. And it is still the backend

4oo4:
Sounds good! It will take me a few days to clone my production Wallabag instance and get ES up and running, at that point I'll start digging into that commit. I will probably have some questions since I haven't poked around the Wallabag source much. Is there a code style guide I should go off of?

tcitworld:
For the php side `php-cs-fixer` does much of the work. On the assets side, run `grunt tests`.

tcitworld:
You could start by rebasing the old commits and opening a new branch. :-)

j0k3r:
Yep also don't hesitate to come on Gitter instead :)

nicosomb:
Done in #2543.

PR #1685, opened 2016-02-17
PoC : Search

tcitworld:
How-to :
* Launch elasticsearch (tested with version 1.74)
* `composer update`
* `php bin/console fos:elastica:populate`
* add entries
* search for entries
* profit

TODO :
* [x] a parameter to enable/disable this feature (for those who can't install ES)
* [ ] Make things properly
* [ ] use RabbitMQ to index without loss of perfs
* [ ] tests
* [ ] add ES in docker-compose

Future :
* [ ] find a way to display searched tags
* [ ] search in comments too

Should fix https://github.com/wallabag/wallabag/issues/18

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -36,6 +36,7 @@ class AppKernel extends Kernel
             new Craue\ConfigBundle\CraueConfigBundle(),
             new Lexik\Bundle\MaintenanceBundle\LexikMaintenanceBundle(),
             new Wallabag\AnnotationBundle\WallabagAnnotationBundle(),
+            new \FOS\ElasticaBundle\FOSElasticaBundle(),
         ];
 
         if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -37,7 +37,6 @@ class AppKernel extends Kernel
             new Lexik\Bundle\MaintenanceBundle\LexikMaintenanceBundle(),
             new Wallabag\AnnotationBundle\WallabagAnnotationBundle(),
             new \FOS\ElasticaBundle\FOSElasticaBundle(),
-            new FOS\ElasticaBundle\FOSElasticaBundle(),
         ];
 
         if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -37,6 +37,7 @@ class AppKernel extends Kernel
             new Lexik\Bundle\MaintenanceBundle\LexikMaintenanceBundle(),
             new Wallabag\AnnotationBundle\WallabagAnnotationBundle(),
             new \FOS\ElasticaBundle\FOSElasticaBundle(),
+            new FOS\ElasticaBundle\FOSElasticaBundle(),
         ];
 
         if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -217,7 +217,7 @@ lexik_maintenance:
 
 fos_elastica:
     clients:
-        default: { host: localhost, port: 9200 }
+        default: { host: %elastic_search_host%, port: %elastic_search_port% }
     serializer:
         callback_class: FOS\ElasticaBundle\Serializer\Callback
         serializer: serializer
@@ -228,17 +228,23 @@ fos_elastica:
                 index:
                     analysis:
                         analyzer:
-                            custom_french_analyzer:
-                                type: custom
-                                tokenizer: letter
-                                filter: ["asciifolding", "lowercase", "french_stem", "stop_fr"]
+
                         filter:
-                            stop_fr:
-                                type: "stop"
-                                stopwords: ["l", "m", "t", "qu", "n", "s", "j", "d"]
+
             types:
                 Entry:
-                    mappings: { analyzer: custom_french_analyzer }
+                    mappings:
+                        title: {boost: 5 }
+                        url: {boost: 2 }
+                        content: ~
+                        createdAt: { type: date, format: basic_date_time }
+                        updatedAt: { type: date, format: basic_date_time }
+                        comments: {boost: 1 }
+                        domainName: {boost: 3 }
+                        tag:
+                            type: "object"
+                        user:
+                            type: "object"
                     persistence:
                         driver: orm # orm, mongodb, propel are available
                         model:  Wallabag\CoreBundle\Entity\Entry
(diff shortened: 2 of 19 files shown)


app/AppKernel.php
```diff
@@ -34,6 +34,7 @@ public function registerBundles()
             new Wallabag\ImportBundle\WallabagImportBundle(),
             new Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
             new Craue\ConfigBundle\CraueConfigBundle(),
+            new \FOS\ElasticaBundle\FOSElasticaBundle(),
```
j0k3r: I think you don't need the first `\`

bin/symfony_requirements
```diff
@@ -7,7 +7,7 @@ $lineSize = 70;
 $symfonyRequirements = new SymfonyRequirements();
 $iniPath = $symfonyRequirements->getPhpIniConfigPath();
 
-echo_title('Symfony Requirements Checker');
+echo_title('Symfony2 Requirements Checker');
```
j0k3r: Nope, could you run `composer up` ?

app/config/config.yml
```diff
@@ -199,3 +199,42 @@ scheb_two_factor:
 kphoen_rulerz:
     executors:
         doctrine: true
+
+fos_elastica:
+    clients:
+        default: { host: localhost, port: 9200 }
```
nicosomb: this config must be in parameters.yml, I think (and I forgot to do the same for RabbitMQ PR I think)

nicosomb:
Nice :) 

I added 2 TODO: a parameter to disable this feature and changes in docker-compose.

j0k3r:
And I guess that if we have questions about ES, we can ask for an expert :wave: @damienalexandre

nicosomb:
Just read this blog post http://blog.zenika.com/2016/02/15/consolider-les-logs-docker-dans-un-elk/ you can find configuration for your docker-compose file.

app/config/config.yml
```diff
@@ -199,3 +199,42 @@ scheb_two_factor:
...
+        wallabag:
+            client: default
+            settings:
+                index:
+                    analysis:
+                        analyzer:
+                            custom_french_analyzer:
+                                type: custom
+                                tokenizer: letter
+                                filter: ["asciifolding", "lowercase", "french_stem", "stop_fr"]
```
damienalexandre: The `french_stem` token filter does not exist, you don't have any stemming on this analyzer.

I suggest you use elision too, instead of `stop_fr`. Have a look at [this example](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html#french-analyzer).

Also, why tokenizing for french only? I'm not sure the contents are mono-lang?

src/Wallabag/CoreBundle/Controller/SearchController.php
```diff
@@ -0,0 +1,80 @@
...
+
+        if ($form->isValid()) {
+
+            $repositoryManager = $this->get('fos_elastica.manager.orm');
+            $repository = $repositoryManager->getRepository('WallabagCoreBundle:Entry');
+            $search = $articleSearch->getSearchTerm();
+            $logger = $this->get('logger');
+            $logger->error($search);
+
+            $articles = $repository->find($search);
```
damienalexandre: By doing this, you make use of the QueryString clause (https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Query.php#L84) I think.

This is not a good way to search in Elasticsearch,
as you are using:
- the `_all` field by default, which is a concatenation of everything, analyzed in standard
- you look in ALL the values without type consideration, so dates for example are messed up
- this is very sensible to syntax error and throw exception if the parse fails
- it is possible to forge very costly QueryString search...

Please consider building your own request using the DSL.

app/config/config.yml
tcitworld: > Also, why tokenizing for french only? I'm not sure the contents are mono-lang?

Not at all. It was to test results. Should we put analysers for all languages or is there an other way ?

damienalexandre: There is no way you will be able to get a good analysis if you don't know the language of the content, so my best guess would be to build a triGram analyzer, with html_strip char filter etc... and use multi-fields.

Each "searchable field" could be mapped:

- with a triGram analyzer (to collect a lot of contents)
- with standard analyzer (to improve the pertinence)

It will not be perfect :disappointed: and will need some tuning when it's setup (there are so many ways to do analysis, you have to found one that will fit your contents and how you want to search them).

tcitworld: The library we use can sometime provide us the language content. Can it be helpful too ?

damienalexandre: It can help yes, 
you could chose to support some strong elision on a list of language (the one already in ES core are a [good starting point](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-lang-analyzer.html)), and search on those on a content basis.

Have a quick look at this to learn more about the recommended config: https://www.elastic.co/guide/en/elasticsearch/guide/current/mixed-lang-fields.html#_use_n_grams

tcitworld:
Strange error.

```
Wallabag\CoreBundle\Tests\Controller\EntryControllerTest::testQuickstart
InvalidArgumentException: Unreachable field "entry"
Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php:44
```

app/config/config.yml
```diff
@@ -210,3 +210,64 @@ lexik_maintenance:
...
+                            type: "object"
+                        user:
+                            type: "object"
+                    persistence:
+                        driver: orm
+                        model:  Wallabag\CoreBundle\Entity\Entry
+                        provider: ~
+                        listener: ~
+                        finder: ~
+                    properties:
```
damienalexandre: This should be part of the "mappings"; why is it here?

app/config/parameters.yml.dist
```diff
@@ -45,3 +45,7 @@ parameters:
     fosuser_confirmation: true
 
     from_email: no-reply@wallabag.org
+
+    elastic_search: true
```
damienalexandre: Elasticsearch is one work, no space, no uppercase S. You should write this "elasticsearch".

src/Wallabag/CoreBundle/Controller/SearchController.php
```diff
@@ -0,0 +1,87 @@
...
+
+                $search = $articleSearch->getSearchTerm();
+
+                $query = new Query();
+                $matching = new MultiMatch();
+                $matching->setType(MultiMatch::TYPE_MOST_FIELDS);
+                $matching->setQuery($search);
+                $matching->setFields(array(
+                    'title^3',
+                    'content',
```
damienalexandre: Searching in tag could be nice too :+1:

PR #1690, opened 2016-02-18
translate estimated time minutes

### LINKED ISSUE #1682: Translation string for 'min' in the 'estimated reading time: 10 min'
selection_128
In RTL languages, the estimated time for reading is displayed wrong. The mix between LTR and RTL words has made the order of the words completely messed up.
It will be better if the word 'min' is also translated into the RTL language (I tested that in my browser).
Is the word 'min' in the screenshot anywhere to be translated?

tcitworld:
Fix #1682
@wallabag/translators


nicosomb:
:+1:

Issue #1137, opened 2015-02-25
Implement Piwik


nicosomb:
For Framabag, I think we should implement Piwik to have more data on our users. 

Do you think it can be a good idea or not?

j0k3r:
Even if it's not for Framabag, we can handle few config parameters to be able to activate Piwik.

So people using Wallabag on their own will be able to have stats too (which won't be that interesting if only one guys use it .. ).

tcitworld:
Isn't Piwik kind of big for what we're trying to get ?

Anyway, what kind of data are you interested in ?

nicosomb:
piwik is used by all frama* services.

tcitworld:
Yeah, I was responding to @j0k3r in particular.

j0k3r:
Well, I was just following the idea of @nicosomb :)

For one guy that host its own wallabag instance, I don't see any value to have piwik.

tcitworld:
Okay. :)
Well, some basic stats are still interesting, though, but no need for piwik for that.

willdurand:
Maybe you could just document where to put the `<script>` tag in the `base.html.twig` template (or something similar). Or a config param that, if filled in, would enable the Piwik tag in a template.

nicosomb:
> Or a config param that, if filled in, would enable the Piwik tag in a template.

That was my idea ;-)

nicosomb:
@tcitworld what about adding this feature in next minor of 1.x? Framasoft needs analytics :-)

tcitworld:
Oh, I guess we can do it easily. It's just a JS line to add, right ? Were should it be (login page, homepage) ?
You need also to inform users on that.

nicosomb:
I can send a mail to our newsletter. 
And we can disable tracking here: https://framabag.org/nav/html/legals.html
Today, we are already tracked but only on framabag homepage.

tcitworld:
>  Today, we are already tracked but only on framabag homepage. 

That's why I asked were should it be (login page, homepage) ? Unread articles ? Calls to save item function ?

nicosomb:
> That's why I asked were should it be (login page, homepage) ?
On each instance (unread / read / favs / config) ... all the pages. 

>  Calls to save item function ?
Not necessary.

willdurand:
Is there a banner (cf. http://cookiesdirective.com/) to inform users?

nicosomb:
Oh not yet!

nicosomb:
[...]

PR #1702, opened 2016-02-19
Added Piwik implementation

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | no
| Documentation | no
| Translation   | yes
| Fixed tickets | #1137 
| License       | MIT

I don't know if I have to write test for this, because I use PiwikExtension and CraueConfigBundle, which are already tested.

Tests failed, due to new params I think. I don't know really why.


src/Wallabag/CoreBundle/Twig/WallabagPiwikExtension.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Wallabag\CoreBundle\Twig;
+
+use Craue\ConfigBundle\Util\Config;
+use PiwikTwigExtension\PiwikTwigExtension;
+
+class WallabagPiwikExtension extends PiwikTwigExtension
+{
+    public function __construct(Config $craueConfig)
+    {
+        parent::__construct($craueConfig->get('piwik_host'), $craueConfig->get('piwik_site_id'), $craueConfig->get('piwik_enabled'));
```
j0k3r: Could you split them line by line? I'll more readable.
Also, could you add a comment on the class telling why you are doing that?

src/Wallabag/CoreBundle/Resources/views/base.html.twig
```diff
@@ -71,5 +71,6 @@
         <div id="warning_message">
             You're trying wallabag v2, which is in alpha version. If you find a bug, please have a look to <a href="https://github.com/wallabag/wallabag/issues">our issues list</a> and <a href="https://github.com/wallabag/wallabag/issues/new">open a new if necessary</a>
         </div>
+        {{ piwik() }}
```
j0k3r: What happen is piwik isn't enabled? It'll display nothing?

nicosomb: It'll display nothing.

PR #1710, opened 2016-02-22
Fix #811: Display a human message when graby fails to fetch articles

### LINKED ISSUE #811: unable to retrieve full-text content
when wallabag gives that message

at least a workaround would be that the plain link to the initial article is shown in wallabag in place of the unparsed text

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #811
| License       | MIT

@wallabag/core: is this message clear?


j0k3r:
Maybe we should add something in the doc about that. 

1. Why does the fetch of an article fail?
2. How can I help to fix that
3. How can I try to re-fetch this article

nicosomb:
Just added documentation. 
I don't know if it's enough. Tell me! (I'm not very inspired)

tcitworld:
> Please contact us if problem persists.

Temporary errors are quite not common, I would rather go with something like « please report this issue to us ».

Issue #1718, opened 2016-02-24
Comments created with one theme aren't displayed with another


tcitworld:
### Issue details

Because of the different structure template, the XPath paths are different : for a same paragraph on material : `/div[1]/main[1]/div[1]/div[1]/article[1]/p[2]` and baggy : `/p[5]`.

We should modify the baggy or material themes in order to have the same structure to avoid this.

EDIT : Moreover, the path should consider only the content if we want comments to be retrieved through the API and displayed correctly.

### Steps to reproduce/test case
Use baggy (or material)
Create comments. They are properly displayed.
Change theme to material (or baggy). Access the entry again.
The comments created before aren't shown.

See #1653

PR #1721, opened 2016-02-24
move annotatorjs to the bottom of the body so that the "article" tag is loaded before

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #1718
| License       | MIT


j0k3r:
Does it fix #1718 or just related to it ?

tcitworld:
It fixes it (annotator now applies to the article tag which is the start of our content).

PR #1725, opened 2016-02-25
Remove SMTP configuration environment overrides

mathbruyen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes (previous overrides no longer work)
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| License       | MIT

SMTP configuration has been added in #1518 to use ParameterHandle's `env-map`. But Symfony actually has a native way of allowing parameters to be overriden from environment so rather than having to define a mapping for each possible parameter, users can define any override in `parameters.yml`:

> parameters:
>     database_host: %WALLABAG_DB_HOST%

and define an environment variable `SYMFONY__WALLABAG_DB_HOST`.

Links:
* [env-map](https://github.com/Incenteev/ParameterHandler#using-environment-variables-to-set-the-parameters)
* [Symfony external parameters](http://symfony.com/doc/current/cookbook/configuration/external_parameters.html)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -89,11 +89,8 @@
         "symfony/phpunit-bridge": "^2.7"
     },
     "scripts": {
-        "build-parameters": [
-            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters"
-        ],
         "post-cmd": [
-            "@build-parameters",
+            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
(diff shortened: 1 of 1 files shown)


j0k3r:
Do we still need the `build-parameters` composer script alias?
Since you finally found that we can override information without modifying configuration files, can you update the documentation about that? Sth like: 

> if you have to use ENV variable to define your configuration

I guess `user/installation.rst` is the right place to put that kind of stuff. What do you think @nicosomb ?

nicosomb:
@j0k3r I agree.

mathbruyen:
Indeed I no longer need `build-parameters`: in now install with `--no-scripts` because that happens before my parameters are properly set and then run the whole `post-cmd` right before startup to warmup cache and bootstrap (see the [Dockerfile](https://github.com/mathbruyen/home/blob/c46af5d2b2d0118813ecd716796d30696fb7d2a9/wallabag/Dockerfile)). I modified the commit to remove it as well.

On documentation I don't really know how to specify it because it requires editing `parameters.yml` thus it does not fully fit in the current workflow.

j0k3r:
Could you please rebase your branch against the v2?
We've put a Github token to avoid error like `Could not authenticate against github.com`.

mathbruyen:
Seems all the tests ok postgres failed, it does not seems linked to the current removal. Should I investigate somewhere?

j0k3r:
No it's ok, we fixed it earlier today.

mathbruyen:
The rebase should be ok.

What about documentation? Would an example setting database from
environment at `composer install` time be interesting?
Le 29 févr. 2016 18:54, "Jeremy Benoist" <notifications@github.com> a
écrit :

j0k3r:
Yes I think this can be interesting :+1:

nicosomb:
I open #1824 (with the rebase).

Issue #1726, opened 2016-02-25
timezone error with docker-compose


DjayDev:
Hello,

I've just created the project through the composer command and I had the good surprise to see a docker-compose file.
So with one ```docker-compose up``` I should be able to test this promising application.
But when I start my browser I get the timezone error that most of people using Symfony had one time in his life.
I know that this issue has been already closed (issue #1591 ) when using our own web server configuration.
But here I'm using docker so I would assume to have the timezone fix in the php image.

Personnally I have added directly in the Dockerfile of php the following instruction 
```RUN echo "date.timezone=Europe/Paris" > /usr/local/etc/php/conf.d/date_timezone.ini```
I don't think it's the best but it works...

Thanks for your work.

j0k3r:
Good point and I don't think we can fix it in an other way. Feel free to provide a PR with that change.

Maybe you can use one of these:
- https://github.com/jacksoncage/phppgadmin-docker/blob/master/Dockerfile#L27-L29
- https://github.com/fideloper/docker-nginx-php/blob/master/Dockerfile#L23-L24

PR #1727, opened 2016-02-25
Add the timezone for PHP docker container

DjayDev:
| Q             | A
| ------------- | ---
| Bug fix?      | yes|
| New feature?  |no
| BC breaks?    |no
| Deprecations? | no
| Tests pass?   | yes|no
| Documentation | yes|no
| Translation   | yes|no
| Fixed tickets | #1726 
| License       | MIT

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -5,6 +5,8 @@ RUN apt-get update && apt-get install -y \
     && docker-php-ext-install \
         iconv mcrypt mbstring intl pdo pdo_mysql pdo_pgsql
 
+RUN echo "date.timezone=Europe/Paris" > /usr/local/etc/php/conf.d/date_timezone.ini
+
 RUN usermod -u 1000 www-data
 
 CMD ["php-fpm"]


docker/php/Dockerfile
```diff
@@ -5,6 +5,8 @@ RUN apt-get update && apt-get install -y \
     && docker-php-ext-install \
         iconv mcrypt mbstring intl pdo pdo_mysql pdo_pgsql
 
+RUN echo "date.timezone=Europe/Paris" > /usr/local/etc/php/conf.d/date_timezone.ini
```
0wnrepo: isn't this a hardcoded path ?

nicosomb: And why `Europe/Paris` ?

FabienM: Maybe this could rely on a [build argument](https://docs.docker.com/engine/reference/builder/#arg), so it will be overridable in `docker-compose.yml` without change in `Dockerfile` ? (starting with *docker-compose* 1.6)

Is there some non-"french timezoned" developers on Wallabag ?

j0k3r:
Could you please rebase your branch against the v2?
We've put a Github token to avoid error like `Could not authenticate against github.com`.

docker/php/Dockerfile
DjayDev: In the image phpfpm, it's the path used for ini files.
As it is linked to this image, I don't see any problem for that.

DjayDev: I'll have a look how to pass arguments to the build from the compose to the
dockerfile, but I'm not a big fan either of using a version that is only
available since a month. It forces everyone to upgrade (oups I was still
using 1.4 )


On Fri, Feb 26, 2016 at 10:05 AM, Fabien Meurillon <notifications@github.com

DjayDev:
New pull request #1749

PR #1740, opened 2016-02-29
minor typofix again

goofy-bz:


j0k3r:
Good catch @NumEricR !
Could you update your PR @goofy-bz ? Then we'll merge it. Thanks !

Issue #1697, opened 2016-02-18
Replace sentences by keys in translations file


nicosomb:
See https://github.com/wallabag/wallabag/issues/1692#issuecomment-185712076

Some good starting point:
- http://jolicode.com/blog/translation-workflow-with-symfony2
- http://developer.happyr.com/how-happyr-work-with-symfony-translations

0wnrepo:
Give an example of how to do this, and I can pick it up on Tuesday. Or anybody else if it's faster.

nicosomb:
See **Working with translation placeholder / It’s also a good practice to use abstract keys instead of plain string:** here http://jolicode.com/blog/translation-workflow-with-symfony2

j0k3r:
Yeah we just need to define a convention on how we are going to use keys. Like [BUNDLE].[MODULE].[LABEL] or sth else

0wnrepo:
Just a quick note, not familiar with php, only Objective-C + Android Java + C/C++. Would be nice if you could make an example, and I can pick up from there.

nicosomb:
Good job, @j0k3r!

PR #1774, opened 2016-03-11
Switch to keys in translated files

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | #1697
| License       | MIT

- translation is disable in test so we only check for translate key instead of translated text.
- remove baggy template for login (never used since user isn't logged in and it'll use the default theme: material)
- fix tests about text in response (now checking translation key instead of translated text)
- remove all ugly `<div class="hidden">{{ form_rest(form) }}</div>`

What's left:
- [x] update all translations files with new structure
- [x] update documentation?


j0k3r:
The PR is ready.

I've added 2 extra checks for travis:
- php-cs-fixer is now launch on every PR so that we won't have to tell the PR author about a CS, the build will fail if a typo is found
- since I've disabled translator on test env (because we don't our test to fail is a translation message is updated) I've added a check on each yaml translated file. It'll avoid typo in these files

nicosomb:
:clap: :cocktail: :+1:

PR #1784, opened 2016-03-13
add refresh-token grant-type

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #1783
| License       | MIT

Issue #1780, opened 2016-03-12
Adding entry through the api returns "500 Internal Server Error"


Eisfreak7:
### Issue details

I'm trying to write a simple shell script to add an entry to my wallabag account. I've got the authorization working properly, and getting all entries works without problems.  
Adding an entry however returns an "Internal Server Error".  
If I'm reading the api documentation correctly I provide the server with all the information it needs.  
Here's the script:
```bash
#!/bin/sh

CLIENT_ID='<client_id>'
CLIENT_SECRET='<client_secret>'
USERNAME='<username>'
PASSWORD='<password>'

WALLABAG_URL='http://v2.wallabag.org'

access_token=$(http POST "$WALLABAG_URL/oauth/v2/token" \
    grant_type=password \
    client_id="$CLIENT_ID" \
    client_secret="$CLIENT_SECRET" \
    username="$USERNAME" \
    password="$PASSWORD" | grep 'access_token' | awk -F'"' '{print $4}')

# This works
# http GET "$WALLABAG_URL/api/entries.json" \
#     Authorization:"Bearer $access_token"

# This doesn't
http POST "$WALLABAG_URL/api/entries.json" \
    Authorization:"Bearer $access_token" \
    url:'http://www.text.com/article.html' \
    tags:'' \
    starred:'false' \
    archive:'false'

```

And here's the output:
```json
{"error":{"code":500,"message":"Internal Server Error"}}
```

### Environment

I'm using the hosted beta (beta 2).

### Steps to reproduce/test case

See the script above.

tcitworld:
Can't see why for now, but wallabag doesn't retrieves the URL, which is `NULL`. Working on it.

tcitworld:
OK, so it seems it doesn't work with HTTPie. For some reason the `url` parameter is badly encoded and may be sent as `\"url\"` instead. But it works with curl for instance instead. If any one got an idea...

```
curl -H "Authorization: Bearer $access_token" --data "url=http://www.text.com/article.html&tags=atag" $WALLABAG_URL/api/entries.json
```

Eisfreak7:
I can confirm that it works with `curl`. Weired. Anyway, thanks for taking a look at it.

nicosomb:
Can we close?

Eisfreak7:
For my purposes, the problem is solved. The original issue still exists though. I don't know if the problem lies with wallabag or httpie. If its a problem with wallabag, maybe the issue shouldn't be closed yet.

[...]

PR #1791, opened 2016-03-15
fix api properties and typo

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #1780, #1784
| License       | MIT

Yeah, because a (bool) string is always true... 
Also, cs.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -126,9 +126,13 @@ class WallabagRestController extends FOSRestController
             $this->get('wallabag_core.content_proxy')->assignTagsToEntry($entry, $tags);
         }
 
-        $entry->setStarred((bool) $isStarred);
+        if (!is_null($isStarred)) {
+            $entry->setStarred((bool) $isStarred);
+        }
 
-        $entry->setArchived((bool) $isArchived);
+        if (!is_null($isArchived)) {
+            $entry->setArchived((bool) $isArchived);
+        }
 
         $em = $this->getDoctrine()->getManager();
         $em->persist($entry);
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -126,11 +126,11 @@ class WallabagRestController extends FOSRestController
             $this->get('wallabag_core.content_proxy')->assignTagsToEntry($entry, $tags);
         }
 
-        if (true === (bool) $isStarred) {
+        if ($isStarred === 'true') {
             $entry->setStarred(true);
         }
 
-        if (true === (bool) $isArchived) {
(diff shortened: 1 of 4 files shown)


src/Wallabag/ApiBundle/Controller/WallabagRestController.php
```diff
@@ -122,11 +122,11 @@ public function postEntriesAction(Request $request)
             $this->get('wallabag_core.content_proxy')->assignTagsToEntry($entry, $tags);
         }
 
-        if (true === (bool) $isStarred) {
+        if ($isStarred === 'true') {
```
j0k3r: What if user send `starred=1` or  `starred=0` ?

tcitworld:
OK, then we'll have to change the docs.

src/Wallabag/ApiBundle/Controller/WallabagRestController.php
tcitworld: In that case, we should also change values retrieved to 0/1, what do you think ?

j0k3r: Yep

src/Wallabag/ApiBundle/Tests/Controller/WallabagRestControllerTest.php
```diff
@@ -197,8 +197,8 @@ public function testPatchEntry()
         $this->client->request('PATCH', '/api/entries/'.$entry->getId().'.json', array(
             'title' => 'New awesome title',
             'tags' => 'new tag '.uniqid(),
-            'star' => true,
-            'archive' => false,
+            'starred' => '1',
+            'archive' => '0',
```
j0k3r: Could you add extra tests with value like `'starred' => 1, 'archive' => 0,` without single quotes to check the behavior?

tcitworld: Done.

src/Wallabag/ApiBundle/Controller/WallabagRestController.php
```diff
@@ -126,12 +126,12 @@ public function postEntriesAction(Request $request)
...
-        if (true === (bool) $isStarred) {
-            $entry->setStarred(true);
+        if (!is_null($isStarred)) {
+            $entry->setStarred((bool) $isStarred);
         }
 
-        if (true === (bool) $isArchived) {
-            $entry->setArchived(true);
+        if (!is_null($isArchived)) {
+            $entry->setArchived((bool) $isArchived);
```
tcitworld: I put them this way because you can update an entry with starred/archived at false by adding it again (#1797)

j0k3r: :+1:

Issue #1796, opened 2016-03-16
[Feature request] API endpoint for checking if a URL is already bookmarked


ThibaudDauce:
Hi, 

I'm working on a Firefox extension for Wallabag and I need an endpoint for checking if a URL is or isn't inside my Wallabag yet (I'll put a different icon on the bar depending on the result)

Is there already a method to achieve this goal? And if not, is it possible to implement something? 

Thanks!

tcitworld:
OK, so for starters there was a bug : adding entries through API with the same URL added them a second time. This is fixed in #1797.

Now, how would `/api/exists` with `url` as a parameter would go ?

ThibaudDauce:
That's ok for me! 

If it's possible, it could be useful to be able to pass an array of links and obtain an array of boolean (to check for example all the link in a page with one request)

tcitworld:
Yup, can be done.

PR #1797, opened 2016-03-16
Don't add entry through API if it already exists

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

If an entry was with the same URL was re-sent a second time, a second article was created in DB. This fixes it.

Note that all metadata (archived, starred, tags, ...) associated with the second call to API will be applied to the original entry.

Also, checkout #1796


j0k3r:
Good catch ! :+1:

tcitworld:
Thank @ThibaudDauce :-)

Issue #1800, opened 2016-03-18
Bring annotations to API


jlnostr:
In the current beta there's an inconsistency when it comes to the API. All methods are available under `URL/api/SOMETHING`. However the API for the annotations is available under `URL/annotations` and therefore without the `api` substring. In terms of consistency the annotations API should be available under `URL/api/annotations`, too :)

tcitworld:
That's because for now the annotation API is only for the web interface (ajax calls to [Annotator](http://docs.annotatorjs.org/en/latest/modules/storage.html)). It shouldn't be shown in the documentation for now.

However, when I'll do the work of serving them through our API, the doc shouldn't be much different.

PR #1803, opened 2016-03-18
bring annotations to API

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | WIP
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/1800
| License       | MIT

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -6,12 +6,14 @@ use FOS\RestBundle\Controller\FOSRestController;
 use Hateoas\Configuration\Route;
 use Hateoas\Representation\Factory\PagerfantaFactory;
 use Nelmio\ApiDocBundle\Annotation\ApiDoc;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Wallabag\CoreBundle\Entity\Entry;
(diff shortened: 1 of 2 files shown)


src/Wallabag/ApiBundle/Controller/WallabagRestController.php
```diff
@@ -344,6 +346,138 @@ public function deleteTagAction(Tag $tag)
...
+     *      requirements={
+     *          {"name"="entry", "dataType"="integer", "requirement"="\w+", "description"="The entry ID"}
+     *      }
+     * )
+     *
+     * @return Response
+     */
+    public function getAnnotationsAction(Entry $entry)
+    {
+
```
j0k3r: You have several extra unnecessary lines like that in this file

tcitworld:
Can I call from APIBundle/RestController to AnnotationBundle/Controller instead of having twice the same functions ?

nicosomb:
Which one? 
Maybe you can put this function into a helper class and inject it via a service?

j0k3r:
If we think more globally.
The AnnontationBundle is just an API for annotator.js, right?
Why don't we move that stuff in ApiBundle with Entity & Repository in CoreBundle?

tcitworld:
Well, thing is Annotator.js doesn't go with OAuth, So we need separate routes just for the annotations made/retrieved by the interface.

nicosomb:
@tcitworld need help for tests?

tcitworld:
It's just as @j0k3r said, there must be a more efficient way to do this without having twice the same function. On one side we have an API (without auth atm, but compatible with [this](https://github.com/heckj/authz) it seems) just for annotations in the web interface and on the other one there's our global API that requires O'Auth.

nicosomb:
(PR replaced by https://github.com/wallabag/wallabag/pull/1890)

Issue #1801, opened 2016-03-18
wallabag can't retrieve contents for this article


ThibaudDauce:
### Issue details

Impossible to fetch the content of http://iansommerville.com/systems-software-and-technology/giving-up-on-test-first-development/.

### Environment

* wallabag version (or git revision) that exhibits the issue: beta 2
* How did you install wallabag? git clone and checkout & composer install
* php version: 7.0.4
* OS: Debian 8
* type of hosting (shared or dedicated): VPS
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): SQLite

### Steps to reproduce/test case

Try to add the article

PR #1804, opened 2016-03-18
Force user-agent for iansommerville.com

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #1801
| License       | MIT

When using the default User Agent, the website returns a 403. By forcing the User Agent for this domain, we got the proper behavior: a 200

Since we can't _append_ new user agent to the default graby configuration, we need to rewrite all the default config + the new one.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/config/services.yml
+++ b/src/Wallabag/CoreBundle/Resources/config/services.yml
@@ -39,7 +39,22 @@ services:
     wallabag_core.graby:
         class: Graby\Graby
         arguments:
-            - { error_message: "wallabag can't retrieve contents for this article. Please report this issue to us." }
+            -
+                error_message: "wallabag can't retrieve contents for this article. Please report this issue to us."
+                http_client:
+                    user_agents:
+                        'lifehacker.com': 'PHP/5.2'
+                        'gawker.com': 'PHP/5.2'
+                        'deadspin.com': 'PHP/5.2'
+                        'kotaku.com': 'PHP/5.2'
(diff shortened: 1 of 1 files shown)


src/Wallabag/CoreBundle/Resources/config/services.yml
```diff
@@ -39,7 +39,22 @@ services:
     wallabag_core.graby:
         class: Graby\Graby
         arguments:
-            - { error_message: "wallabag can't retrieve contents for this article. Please report this issue to us." }
+            -
+                error_message: "wallabag can't retrieve contents for this article. Please report this issue to us."
+                http_client:
+                    user_agents:
+                        'lifehacker.com': 'PHP/5.2'
```
nicosomb: We'll have a huge file in next weeks?

j0k3r: Next weeks?
We don't update that list quite often.
But I agree that this isn't greatly designed in graby (inherited from full-text-rss on that part)

nicosomb:
@wallabag/core Do we need to try to move this fix in graby or can I merge?

j0k3r:
This'll require too much work on graby.
This won't be done in the coming weeks.

You can merge that and we'll see later when (if ?) graby evolve, we'll update the code accordingly.

PR #1805, opened 2016-03-21
Fix #1768 Split tags with commas from UI

### LINKED ISSUE #1768: Split tags with commas
### Issue details

When adding tags to an entry, if there's commas, add individual tags.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #1768 
| License       | MIT


tcitworld:
This works ? Damn.

j0k3r:
Could you add some test on the TagControllerTest.php ?
Add a coma separated tag and check that both tag are created / associated.

nicosomb:
In fact, my code is not OK. 

I have to do a `foreach` in the controller before the query.

PR #1812, opened 2016-03-23
Update medium.com.txt

esszett:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
fix wrong x-path selector, avoid pruned content


j0k3r:
Could you target the dev branch instead ?
Also, could you submit that patch here too: https://github.com/fivefilters/ftr-site-config/blob/master/medium.com.txt

And instead of updating the line I think you should better add a new one.

PR #1814, opened 2016-03-24
[Do Not Merge] Wallabag 1.9.2

tcitworld:
Get ready for 1.9.2


themes/default/_messages.twig
```diff
@@ -1 +1,6 @@
-                {{ messages | raw }}
\ No newline at end of file
+                {{ messages | raw }}
+                {% if view == 'home' %}
+                <div class="messages information">
+                    <p><a href="http://doc.wallabag.org">wallabag v2 is out, please read the documentation to upgrade.</a></p>
```
tolkonepiu: but why? why not show a message once? Version 2 runs slowly on my device, so it is not needed.

PR #1824, opened 2016-03-29
Remove SMTP configuration environment overrides

nicosomb:
Previously, here #1725 (by @mathbruyen) ... 

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes (previous overrides no longer work)
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| License       | MIT

SMTP configuration has been added in #1518 to use ParameterHandle's `env-map`. But Symfony actually has a native way of allowing parameters to be overriden from environment so rather than having to define a mapping for each possible parameter, users can define any override in `parameters.yml`:

> parameters:
>     database_host: %WALLABAG_DB_HOST%

and define an environment variable `SYMFONY__WALLABAG_DB_HOST`.

Links:
* [env-map](https://github.com/Incenteev/ParameterHandler#using-environment-variables-to-set-the-parameters)
* [Symfony external parameters](http://symfony.com/doc/current/cookbook/configuration/external_parameters.html)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,11 +90,8 @@
         "fabpot/php-cs-fixer": "~1.9"
     },
     "scripts": {
-        "build-parameters": [
-            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters"
-        ],
         "post-cmd": [
-            "@build-parameters",
+            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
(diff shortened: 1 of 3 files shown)


nicosomb:
Added documentation.

PR #1835, opened 2016-04-03
:rocket: Prepare v2.0.0 :birthday:

nicosomb:
:birthday: :birthday: :birthday:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,8 +8,7 @@ Click, save and read it when you can. It extracts content so that you can read i
 
 More informations on our website: [wallabag.org](https://wallabag.org)
 
-# Want to test the v2?
-Keep in mind it's an **unstable** branch, everything can be broken :)
+# Install wallabag
 
 If you don't have it yet, please [install composer](https://getcomposer.org/download/).
 Then you can install wallabag by executing the following commands:
@@ -17,7 +16,7 @@ Then you can install wallabag by executing the following commands:
 ```
     git clone https://github.com/wallabag/wallabag.git
     cd wallabag
-    git checkout 2.0.0-beta.2
+    git checkout 2.0.0
     SYMFONY_ENV=prod composer install --no-dev -o --prefer-dist
     php bin/console wallabag:install --env=prod
     php bin/console server:run --env=prod
--- a/src/Wallabag/CoreBundle/Resources/views/base.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/base.html.twig
@@ -70,10 +70,6 @@
 
         {% block footer %}{% endblock %}
 
-        <div id="warning_message">
-            You're trying wallabag v2, which is in beta version. If you find a bug, please have a look to <a href="https://github.com/wallabag/wallabag/issues">our issues list</a> and <a href="https://github.com/wallabag/wallabag/issues/new">open a new one if necessary</a>.
-        </div>
-
         {% if craue_setting('piwik_enabled') %}
             {{ piwik(craue_setting('piwik_host'), craue_setting('piwik_site_id')) }}
         {% endif %}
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -8,8 +8,7 @@ Click, save and read it when you can. It extracts content so that you can read i
 
 More informations on our website: [wallabag.org](https://wallabag.org)
 
-# Want to test the v2?
-Keep in mind it's an **unstable** branch, everything can be broken :)
+# Installa wallabag
```
NumEricR: Typo here.
Do you mean "Install wallabag" or "Install a wallabag"?

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Static/howto.html.twig
```diff
@@ -11,7 +11,7 @@
     </ul>
     <h3>{{ 'howto.top_menu.browser_addons'|trans }}</h3>
     <ul>
-        <li><a href="https://addons.mozilla.org/firefox/addon/wallabag/" target="_blank">{{ 'howto.browser_addons.firefox'|trans }}</a></li>
+        <li><a href="https://addons.mozilla.org/fr/firefox/addon/wallabag-v2/" target="_blank">{{ 'howto.browser_addons.firefox'|trans }}</a></li>
```
NumEricR: Why using the french URL?
It should be better to avoid that language string :wink: 

https://addons.mozilla.org/firefox/addon/wallabag-v2/

src/Wallabag/CoreBundle/Resources/views/themes/material/Static/howto.html.twig
```diff
@@ -18,7 +18,7 @@
                     <div class="col s12">
                         <h5>{{ 'howto.top_menu.browser_addons'|trans }}</h5>
                         <ul>
-                            <li><a href="https://addons.mozilla.org/firefox/addon/wallabag/" target="_blank">{{ 'howto.browser_addons.firefox'|trans }}</a></li>
+                            <li><a href="https://addons.mozilla.org/fr/firefox/addon/wallabag-v2/" target="_blank">{{ 'howto.browser_addons.firefox'|trans }}</a></li>
```
NumEricR: Same comment here: why using the french URL?

https://addons.mozilla.org/firefox/addon/wallabag-v2/

README.md
nicosomb: install wallabag ;-) 
Thank you, will fix.

src/Wallabag/CoreBundle/Resources/views/base.html.twig
```diff
@@ -70,10 +70,6 @@
 
         {% block footer %}{% endblock %}
 
-        <div id="warning_message">
```
j0k3r: Can you remove the associated css too ?

tcitworld: Could we instead keep the div and activate the message through parameters ?

tcitworld: (In the future, at least)

nicosomb: It was a parameter before :)

tcitworld: That's what I thought. How come we changed it ?

j0k3r: This can be moved to the config_prod.yml file and should be disabled in the config_dev.yml.
I don’t like this message in dev :)

And yep it was in the parameters.yml before, but I don’t think it should be a param that must be defined by the end user when it install wallabag.


On 3 avril 2016 at 20:31:17, Thomas Citharel (notifications@github.com) wrote:

In src/Wallabag/CoreBundle/Resources/views/base.html.twig:

> @@ -70,10 +70,6 @@
>   
>          {% block footer %}{% endblock %}
>   
> -        <div id="warning_message">
That's what I thought. How come we changed it ?

nicosomb: > (In the future, at least)

We'll add the `<div>`. I'm not fan of the "and if ever...". If we'll need it, we'll add it.

tcitworld: > I don’t like this message in dev :)

Yup, it hides things, especially in mobile view. TODO for next release.

Issue #1747, opened 2016-03-02
v1.9.1: JSON export returns empty file


Rurik19:
### Issue details

Wallabag v1.9.1 ; JSON export returns empty file. I have ~900 entries with pictures stored.
### Environment
- wallabag version Wallabag v1.9.1
- wallabag installed Via `composer create-project`
- Last wallabag version that did not exhibit the issue (if applicable): no
- php version: 5.6.17
- OS: debian 8.0 armhf
- type of hosting: dedicated
- storage system MySQL
### Steps to reproduce/test case

config -> export json -> empty file

Strubbl:
seems to duplicate #1534

ymage:
I took a look yesterday night and it seems related to utf8 encoding issue.

benediktg:
Same here, I have an account with about 1250 articles and the export only delivers a 0-byte-sized file.

Is it possible that the export has problems with a big count of articles?

I wonder if it'd be a feasible workaround to extract the article URLs as well as the read and favorite states of a certain user out of the database table `entries` with

```
SELECT `url`, `is_read`, `is_fav` FROM `entries` WHERE `user_id` = '2';
```

and somehow enter this data automatically into the Wallabag v2 account.

Rurik19:
I already did so. There only three sql to migrate from 1.9.1 to 2.0

ymage:
Hi,

The 0-byte-sized file comes from a JSON encoding throwing a JSON_ERROR_UTF8 error.

Strubbl:
@ymage in which file and line is the errors cause (not in the json file, the src creating the json)?

blackere:
In `inc/poche/Tools.class.php` at function `renderJson` the `json_encode` return the message
`unexpected end of data at line 1 column 1 `

benediktg:
@ymage 
> The 0-byte-sized file comes from a JSON encoding throwing a JSON_ERROR_UTF8 error.

The patch from #1866 was applied on the server but it didn't help with the 0-Byte-File – it just did not add any new problems…

ymage:
You're right.
One reason (the one I encountered) for this 0-byte-file was a JSON_ERROR_UTF8 error.
It may have several other reason I did not experienced.

tcitworld:
Version 1.9.2 is out. Please give it a try.

Rurik19:
Unfortunately with 1.9.2 JSON export also doesn't work. Which log should I inspect for details?

[...]

PR #1866, opened 2016-04-05
Fix UTF-8 encoding before JSON export

ymage:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes|no
| Documentation | no
| Translation   | no
| Fixed tickets | #1747 
| License       | MIT

### DIFF
--- a/inc/poche/Poche.class.php
+++ b/inc/poche/Poche.class.php
@@ -908,7 +908,7 @@ class Poche
         $entries = $this->store->retrieveAllWithTags($this->user->getId());
         if ($entries) {
             echo $this->tpl->render('export.twig', array(
-            'export' => Tools::renderJson($entries),
+            'export' => Tools::renderJson(Tools::utf8ize($entries)),
             ));
             Tools::logm('export view');
         } else {
--- a/inc/poche/Tools.class.php
+++ b/inc/poche/Tools.class.php
@@ -237,18 +237,25 @@ final class Tools
      *
      * @param $data
      */
-    public static function utf8ize($data) 
+    public static function utf8ize($data)
     {
-        if (is_array($data)) 
+       if (is_array($data))
         {
-            foreach ($data as $k => $v) 
+            foreach ($data as $k => $v)
             {
                 $data[$k] = self::utf8ize($v);
             }
         }
-        else if (is_string ($data)) 
+        else if (is_string ($data))
         {
-            return utf8_encode($data);
+            if ('' == mb_detect_encoding($data))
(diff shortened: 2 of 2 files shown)


benediktg:
I have tried this patch (see commands down below) but umlauts got messed up after the articles were imported into Wallabag v2.

```
cd /path/to/wallabag
wget https://github.com/ymage/wallabag/commit/206c2a734e2a0d100af6d772d90b01c5cb8fe870.patch
patch -blp1 < 206c2a734e2a0d100af6d772d90b01c5cb8fe870.patch
```

tcitworld:
> umlauts got messed up after the articles were imported into Wallabag v2.

And the import worked before with these umlauts ?

ymage:
Maybe, I should add a JSON_UNESCAPED_UNICODE during json_encode to avoid this mess

benediktg:
@tcitworld Yes, without or after undoing this patch and trying the export-import-procedure umlauts are not messed up. With only few articles there are no problems at all. I get the empty file because of the hundreds of articles as I have assumed in #1747.

ymage:
@benediktg Can you provide us some test samples in order to improve this patch ?

benediktg:
@ymage The slightly older file is with this patch applied and exported, the other file is after I have undone the patch and exported again.

[wallabag-exports.zip](https://github.com/wallabag/wallabag/files/206927/wallabag-exports.zip)

tcitworld:
Thanks @ymage, this seems to work. I'm just trying to see what happens with huge amount of data.

benediktg:
I tried the patch again with
```
wget https://patch-diff.githubusercontent.com/raw/wallabag/wallabag/pull/1866.patch
patch -bp1 < 1866.patch
```
Here is the resulting exported JSON (I just added .txt to the filename because Github requires it):
[wallabag-export-2-2016-04-10.json.txt](https://github.com/wallabag/wallabag/files/211857/wallabag-export-2-2016-04-10.json.txt)

And when I import that to Wallabag v2 it looks good, too. Now I just have to see how it behaves with big amounts of articles.

nicosomb:
Can we merge?

benediktg:
On the other hand the patch did not create new problems, so there is nothing wrong with merging it in my opinion.

ymage:
Well,

This patch deals only with UTF-8 encoding issue before json_encode. 
It does not cover other export issues.

PR #1887, opened 2016-04-08
Fix duplicate article when added via the bookmarklet

### LINKED ISSUE #1886: Duplicate entry via bookmarklet
### Issue details

When an article already present is added via the "+" button, a message indicate that it was already added and the date.
When adding the same article via the bookmarklet, a new entry is created 

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.0
* Wallabag installation via git clone
* php version: 5.6.19-0+deb8u1
* OS: debian 8.4
* type of hosting (shared or dedicated): dedicated VM
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): PostgreSQL

Open the original article of an entry present in wallabag. Use the bookmarklet to save it. A new entry is present and is a duplicate (the previous entry is still present).

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Tests pass?   | yes
| Fixed tickets | #1886
| License       | MIT

Fix #1886

@j0k3r can you tell me if we need to refactor with `addEntryFormAction`? And do we need tests here?

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -430,7 +430,7 @@ class EntryController extends Controller
      *
      * @return array|bool
      */
-    public function checkIfEntryAlreadyExists($entry)
+    private function checkIfEntryAlreadyExists($entry)
     {
         return $this->get('wallabag_core.entry_repository')->findByUrlAndUserId($entry->getUrl(), $this->getUser()->getId());
     }
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -49,8 +49,7 @@ class EntryController extends Controller
         $form->handleRequest($request);
 
         if ($form->isValid()) {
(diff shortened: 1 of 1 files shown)


j0k3r:
Yeah both :)

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -420,4 +422,16 @@ private function checkUserAction(Entry $entry)
...
     }
+
+    /**
+     * Check for existing entry, if it exists, redirect to it with a message.
+     *
+     * @param $entry
+     *
+     * @return array|bool
+     */
+    public function checkIfEntryAlreadyExists($entry)
```
j0k3r: I'll put that private

j0k3r:
Ok test aren't needed since the new method is shared.

src/Wallabag/CoreBundle/Controller/EntryController.php
nicosomb: Oh yes it was what I wanted to do (but phpstorm refactored with a public method ... and I didn't checked!)

PR #1773, opened 2016-03-11
Fix #1769: Added mysql version in config.yml to fix the unknown database bug

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #1769
| License       | MIT

Issue #1769, opened 2016-03-08
Can't install wallabag via composer install if my mysql database doesn't exist


nicosomb:
### Issue details

When I launch `SYMFONY_ENV=prod composer install --no-dev -o`, if my mysql database doesn't exist before this command, I can't install wallabag: 

```
An exception occured in driver: SQLSTATE[42000] [1049] Unknown database 'wallabag' 
``` 

Found with @chris25640 

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.0-beta.1
* How did you install wallabag? Via `composer install`
* Last wallabag version that did not exhibit the issue (if applicable): N/A
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MySQL

### Steps to reproduce/test case

* git clone https://github.com/wallabag/wallabag.git
* cd wallabag
* git checkout 2.0.0-beta.1
* SYMFONY_ENV=prod composer install --no-dev -o
* Set `pdo_mysql` for `database_driver` parameter
* Set a name of an inexistent database for `database_name`

nicosomb:
See http://stackoverflow.com/questions/27503328/composer-install-fails-when-unable-to-see-mysql-database and https://github.com/doctrine/DoctrineBundle/issues/351#issuecomment-75456547

captainwasabi:
I added the line in apt/config/config.yml and I still get the same error when trying to set up beta 2

nicosomb:
Fixed by #1903.

PR #1903, opened 2016-04-10
Force server version to avoid connexion error

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #1769 
| License       | MIT

When installing a new project with MySQL, if the database doesn't exist before installation will fail because DBAL will try to get the server version by querying it.
And since this query is done using the default database connexion and since the database doesn't exist, DBAL can't connect to the server.

Following #1773
See http://blog.insight.sensiolabs.com/2014/12/22/making-symfony-bootable-with-dbal-2-5.html

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -82,6 +82,7 @@ doctrine:
         password: "%database_password%"
         charset:  UTF8
         path:     "%database_path%"
+        server_version: 5.6
 
     orm:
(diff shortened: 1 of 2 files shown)


nicosomb:
I was trying to fix this issue too ;-)

PR #1904, opened 2016-04-10
Share entry with a public URL

### LINKED ISSUE #95: public mode
I don't know if it's a good idea :  to allow everyone to read the articles, but without being able to archive and delete it, etc.. 

Do you have any idea?

nicosomb:
| Q             | A
| ------------- | ---
| New feature?  | yes
| Tests pass?   | N/A
| Documentation | no
| Translation   | no
| Fixed tickets | #95 
| License       | MIT

Fix #95


src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -284,6 +284,8 @@ public function viewAction(Entry $entry)
     {
         $this->checkUserAction($entry);
 
+        $this->generateEntryUuid($entry);
```
j0k3r: Why do you need it here? To handle previous entries?

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -582,4 +593,31 @@ public function getLanguage()
...
+    {
+        $this->uuid = $uuid;
+
+        return $this;
+    }
+
+    public function generateUuid()
+    {
+        if (empty($this->uuid) || is_null($this->uuid)) {
+            $this->uuid = uniqid();
```
j0k3r: :arrow_right: http://blog.kevingomez.fr/til/2015/07/26/why-is-uniqid-slow/

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
```diff
@@ -101,6 +101,13 @@
                 <ul>
                     {% if craue_setting('share_twitter') %}
```
tcitworld: Not share twitter but share public

src/Wallabag/CoreBundle/Controller/EntryController.php
nicosomb: Right.

nicosomb: Oh we'll write a migration !

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
nicosomb: You're right. It's a draft I pushed to not loose my work ;-)

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
```diff
@@ -17,6 +17,8 @@
             <li><a title="{{ 'entry.view.left_menu.set_as_favorite'|trans }}" class="tool icon icon-star {% if entry.isStarred == 0 %}fav-off{% else %}fav{% endif %}" href="{{ path('star_entry', { 'id': entry.id }) }}"><span>{{ 'entry.view.left_menu.set_as_favorite'|trans }}</span></a></li>
             <li><a id="nav-btn-add-tag" title="{{ 'entry.view.left_menu.add_a_tag'|trans }}"><span>{{ 'entry.view.left_menu.add_a_tag'|trans }}</span></a></li>
             <li><a title="{{ 'entry.view.left_menu.delete'|trans }}" class="tool delete icon icon-trash" href="{{ path('delete_entry', { 'id': entry.id }) }}"><span>{{ 'entry.view.left_menu.delete'|trans }}</span></a></li>
+            {% if craue_setting('share_public') %}<li><a href="{{ path('share', {'uuid': entry.uuid }) }}" target="_blank" class="tool public" title="public"><span>public</span></a></li>
+            {% endif %}
```
j0k3r: Could this be on the same line?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -434,4 +436,32 @@ private function checkIfEntryAlreadyExists($entry)
...
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function shareEntryAction(Entry $entry)
+    {
+        return $this->render(
+            '@WallabagCore/themes/share.html.twig',
+            array('entry' => $entry)
+        );
+    }
```
j0k3r: Since we are sending a real public response, we might want to use a Response with some cache header, etc ...

app/DoctrineMigrations/Version20160410190541.php
```diff
@@ -0,0 +1,53 @@
...
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
```
nicosomb: Migration is not OK yet. I need to work on it again. 
And SQLite doesn't support `DROP COLUMN` ...

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
```diff
@@ -17,6 +17,7 @@
             <li><a title="{{ 'entry.view.left_menu.set_as_favorite'|trans }}" class="tool icon icon-star {% if entry.isStarred == 0 %}fav-off{% else %}fav{% endif %}" href="{{ path('star_entry', { 'id': entry.id }) }}"><span>{{ 'entry.view.left_menu.set_as_favorite'|trans }}</span></a></li>
             <li><a id="nav-btn-add-tag" title="{{ 'entry.view.left_menu.add_a_tag'|trans }}"><span>{{ 'entry.view.left_menu.add_a_tag'|trans }}</span></a></li>
             <li><a title="{{ 'entry.view.left_menu.delete'|trans }}" class="tool delete icon icon-trash" href="{{ path('delete_entry', { 'id': entry.id }) }}"><span>{{ 'entry.view.left_menu.delete'|trans }}</span></a></li>
+            {% if craue_setting('share_public') %}<li><a href="{{ path('share', {'uuid': entry.uuid }) }}" target="_blank" class="tool public" title="public"><span>public</span></a></li>{% endif %}
```
j0k3r: `public` needs to be translated.
Also I think _public link_ should be a better label.

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
```diff
@@ -99,6 +99,13 @@
             </a>
             <div class="collapsible-body">
                 <ul>
+                    {% if craue_setting('share_public') %}
+                        <li>
+                            <a href="{{ path('share', {'uuid': entry.uuid }) }}" target="_blank" class="tool public" title="public">
+                                <span>public</span>
```
j0k3r: Same about translation here.

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -442,4 +445,33 @@ private function checkIfEntryAlreadyExists(Entry $entry)
...
+     * @param Entry $entry
+     *
+     * @Route("/share/{uuid}", requirements={"uuid" = ".+"}, name="share")
+     * @Cache(maxage="25200", public=true)
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function shareEntryAction(Entry $entry)
+    {
+        return $this->render(
```
j0k3r: What if someone doesn't allow public url and a user _magically_ found an uuid entry, there are no check that'll disallow viewing the content.
This'll require a test too.

tcitworld:
I'll try to handle annotations auth so that they would be displayed in read-only mode.

j0k3r:
Rebased ✅

nicosomb:
I changed the behavior. 

For an entry, we now have a private URL, `/share/{id}`. If `uuid` exists, we redirect to the public URL. If it doesn't, we generate it and redirect to the public URL . 
We also have a `/share/delete/{id}` to delete public sharing for a specific entry.

Do you think that something is missing?

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -595,4 +604,37 @@ public function getLanguage()
...
+        $this->uuid = $uuid;
+
+        return $this;
+    }
+
+    public function generateUuid()
+    {
+        if (empty($this->uuid) || is_null($this->uuid)) {
+            // @see http://blog.kevingomez.fr/til/2015/07/26/why-is-uniqid-slow/ for true parameter
+            $this->uuid = uniqid('', true);
```
tcitworld: I don't know about the performances of uniqid, but how about `md5(salt + id)` instead ?

tcitworld:
The `uuid` should be calculated when creating entry, so that the sharing goes faster.

tcitworld:
> For an entry, we now have a private URL, /share/{id}.

You mean public ?

nicosomb:
> You mean public ?

No, private. 

> The uuid should be calculated when creating entry, so that the sharing goes faster.

To avoid "What if someone doesn't allow public url and a user magically found an uuid entry, there are no check that'll disallow viewing the content." (comment from @j0k3r)

tcitworld:
> For an entry, we now have a private URL, /share/{id}.

I don't understand the purpose of this url then.

> To avoid "What if someone doesn't allow public url and a user magically found an uuid entry, there are no check that'll disallow viewing the content." (comment from @j0k3r)

Well, you should just put a test for the config item before displaying content. Also, `uniqid` is...rather unique.

nicosomb:
> I don't understand the purpose of this url then.

To check if the current user can generate uuid.

tcitworld:
>> I don't understand the purpose of this url then.

> To check if the current user can generate uuid.

Why would you need an url for this ? "Just" return a success/error code and the published url through json and handle this with some js. Better start doing things this way from now.

src/Wallabag/CoreBundle/Entity/Entry.php
j0k3r: Actually we won't generate million of uuid at once, so I think uniqid is fine :)

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -595,4 +604,37 @@ public function getLanguage()
...
+    /**
+     * @return int
+     */
+    public function getUuid()
+    {
+        return $this->uuid;
+    }
+
+    /**
+     * @param int $uuid
```
j0k3r: It's a string.

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -595,4 +604,37 @@ public function getLanguage()
     {
         return $this->language;
     }
+
+    /**
+     * @return int
```
j0k3r: Same, it's a string

nicosomb:
@j0k3r what is your opinion about our last messages? (I fixed annotations for uuid field)

j0k3r:
I would like to test it first.
Will give a shot tonight.

app/DoctrineMigrations/Version20160410190541.php
j0k3r: So what are we doing for that? *we don't care*? :)

j0k3r:
I'm ok with the link to check uuid / generate it and then redirect to the public url.
Also I've added some missing migration / translation.

app/DoctrineMigrations/Version20160410190541.php
```diff
@@ -20,12 +19,17 @@ public function setContainer(ContainerInterface $container = null)
         $this->container = $container;
     }
 
+    private function getTable($tableName)
+    {
+        return $this->container->getParameter('database_table_prefix') . $tableName;
```
nicosomb: 👍

nicosomb:
Thank you @j0k3r! 

If you're OK, we can merge this PR.

j0k3r:
:+1:

PR #1923, opened 2016-04-13
Add doctrine clear after flush

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | N/A
| License       | MIT

After each `$em->flush()` while importing, we now ask to Doctrine to do a clear, thanks to @BitOne talk at Symfony Live about php memory leaks :+1: .


j0k3r:
Well the sad part is that these two lines (the flush & clear) aren't tested because there is no test that import more than 20 entries (partially my fault).
Maybe, since we add more stuff in that `if`, it could be a good idea to cover these lines :grimacing:

nicosomb:
Not so complicated to add a huge file for our tests, I can provide my wallabag v1 export.

j0k3r:
We don't need hundred of entries, just at least 21 :)

nicosomb:
I think I can provide 21 links :D

Issue #1611, opened 2016-01-21
Import from wallabag asynchronously 


nicosomb:
As we can see in #1598, import massive files is not possible (we set the limit to 20M on v2.wallabag.org but it's not the good solution. 
We need to implement RabbitMQ as in #1581.
We need to refactor JSON export in wallabag v1 to only download articles URL, not content.

jcharaoui:
I'd like to suggest Wallabag should keep import/export of content for a simple reason. With really large collections, some articles could be years old and the URLs to access them may not work down the road. The website could be completely offline, or the CMS may have changed and the old permalinks not migrated t the new platform.

Also, to solve the problem of loading big export files into memory, another solution would be to use a streaming json parser suck as https://github.com/salsify/jsonstreamingparser. This would allow loading only one entry at a time in memory for processing, making the process much faster and less ressource-intensive.

tcitworld:
> I'd like to suggest Wallabag should keep import/export of content for a simple reason.

We can't retrieve content from Pocket, but however the solution you provide us may help us to keep content from json files.

>  We need to refactor JSON export in wallabag v1 to only download articles URL, not content. 

We should make this a choice, then. If export with content fails, then export just URLs and metadata.

jcharaoui:
> We should make this a choice, then. If export with content fails, then export just URLs and metadata.

That's one option, but it could also be feasible for the export process to keep track of which entries are written out to disk and resume in case of a PHP timeout error. Regardless, unless I'm mistaken I think the export process (database to file) is by nature much faster than import (file to database) because we're having to run multiple database queries per entry. So making the v1 -> v2 JSON import more efficient seems to me like a bigger priority than refactoring the v1 export code.

tcitworld:
> So making the v1 -> v2 JSON import more efficient seems to me like a bigger priority than refactoring the v1 export code.

A number of users also have encountered timeout/memory issues while exporting json from v1, so it stays a concern too.

jcharaoui:
>> So making the v1 -> v2 JSON import more efficient seems to me like a bigger priority than refactoring >> the v1 export code.

> A number of users also have encountered timeout/memory issues while exporting json from v1, so it stays a concern too.

Agreed!

HLFH:
Hi @j0k3r @nicosomb Is [asynchronous export](https://github.com/wallabag/wallabag/issues/1927) easier to implement now you have implemented the asynchronous import feature? But I guess it won't be for the 2.1.0 milestone.

[...]

PR #1941, opened 2016-04-18
Use asynchronous jobs for imports

nicosomb:
Issue: #1611 

See #1581 for discussion.

Will also fix https://github.com/wallabag/wallabag/issues/2257


src/Wallabag/ImportBundle/Import/PocketImport.php
```diff
@@ -252,16 +261,25 @@ private function parseEntries($entries)
...
             }
+
             ++$i;
         }
 
         $this->em->flush();
+
+        if ($this->rabbitMQ) {
+            foreach ($entries as $entry) {
+                $this->producer->publish(serialize($entry));
```
j0k3r: Don't you need to send `$pocketEntry` instead of `$entry`?
Also, why do you need `$pocketEntry` by reference in the foreach? I don't get it...

nicosomb: I forgot what I did on this PR ...

j0k3r:
As described in my last commit ed264c9, instead of queuing real Entry to process, we queue all items to import from Pocket in a raw format.

Then, the worker retrieve that information (the item from Pocket saved in the queue), find / create the entry and save it.

It allows import from the user part to be really fast. In some seconds, 5k messages are queued in Rabbit.

- [ ] loop thru all Pocket items to retrieve ALL items (and not only the first 5k)
- [ ] try to get unread item first from Pocket
- [ ] migrate other import to Rabbit
- [ ] Use [SwarrotBundle](https://github.com/swarrot/SwarrotBundle) to enable more broker? (we'll have to code the one for Redis)
- [ ] Add tests

j0k3r:
So Redis is now available for async import.

- using [javibravo/simpleue](https://github.com/javibravo/simpleue)
- internal config value are now `import_with_redis` & `import_with_rabbit` which are more clear
- if both option are enable rabbit will be choosen
- services imports related to async are now splitted into 2 files: `redis.yml` & `rabbit.yml`

Todo:

- [ ] Test Redis worker command

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: php
 
+services:
+  - rabbitmq
```
nicosomb: Don't forget Redis service.

nicosomb:
Don't forget redis documentation too.

nicosomb:
Just tried redis worker 👍 

@j0k3r do you think it's possible to display a message like that "There are 14 entries (and 10 will be imported)" ?

j0k3r:
Yeah I need to work on displaying a message about what will be imported.
Also, I can try to display number on message still in the queue.

j0k3r:
From the last commits:

- documentation updated
- a message is displayed when async is enabled ([screenshot](https://cl.ly/271o3d051P3s/Image%202016-09-12%20at%207.07.38%20AM.png)) ➡️ **please check that the displayed message looks ok**
- tests added (also on ImportCommand)

Todo:

- [ ] find a way to display how many messages are in queue (Redis is ok ([screenshot](https://cl.ly/3p0s2l2I1A33/Image%202016-09-12%20at%207.10.32%20AM.png)), Rabbit in progress)
- [ ] update message after import telling article are queue and not imported / skipped

tcitworld:
`Import is made asynchronously. Once the import task is started, an external worker will handle jobs one at a time.`

tcitworld:
>  find a way to display how many messages are in queue 

This pop-up bag is shown on every new page the user accesses ?

j0k3r:
@tcitworld for now, yes. Until queue is empty. What do you suggest?

tcitworld:
The perfect thing would have to have a notification system or the pop-up bag to be dismissible. But that's not really a work for right now.

docker-compose.yml
```diff
@@ -46,7 +49,13 @@ services:
     #        - ./docker/data/mariadb:/var/lib/mysql
     #    env_file:
     #        - ./docker/mariadb/env
+
     rabbitmq:
         image: rabbitmq:3-management
         ports:
             - "15672:15672"
+
+    rabbitmq:
```
nicosomb: I think you mean `redis` here.

nicosomb:
I deployed this branch on v2.wallabag.org. 

Todo: 

- [ ] do not display "messages in queue: XXX" because it's not the total of the current user.

j0k3r:
What I've done

- toast with number of messages in queue are now ONLY displayed for admin user (maybe we'll change that in the future)
- moved pocket token to user config
- changed the way to save an article during import / basic save: it now always save the url if fetching content fail

nicosomb:
👍

tcitworld:
:+1:

j0k3r:
🚀

nicosomb:
🎉

Issue #1789, opened 2016-03-13
Idea to simplify UI for results numbers


NumEricR:
Here's an idea to avoid the result sentence (ie. "There are 13 entries.") used on unread, starred and archive categories.

Why not giving the number of articles next to the category title?
We could see something like this:
![capture d ecran 2016-03-13 a 22 27 47](https://cloud.githubusercontent.com/assets/262560/13731515/362f787c-e96b-11e5-9465-d142115a5376.png)
---
... compared to existing UI:
![capture d ecran 2016-03-13 a 22 28 14](https://cloud.githubusercontent.com/assets/262560/13731516/3b904a76-e96b-11e5-8a9b-c15f0cf918dc.png)

tcitworld:
The idea itself is good, but maybe we can apply certain styles to the number so it catches the eye more. Any ideas ?

modos189:
![362f787c-e96b-11e5-9465-d142115a5376](https://cloud.githubusercontent.com/assets/1785196/13820020/9ceb3a04-ebac-11e5-9a2a-f7111fb7c6b7.png)
Could be better so?

tcitworld:
I love it !

modos189:
Now I'm busy, I can do in a weekend

NumEricR:
Much better than my own idea: we can see those informations on any page! :+1:

tcitworld:
> we can see those informations on any page!

Unfortunately, the count for each category isn't transmitted to each page, so there's backend changes to implement first.

tcitworld:
> Now I'm busy, I can do in a weekend

Don't rush, it can't be done with current backend, I've postponed it to v2.1.

NumEricR:
Don't forget to keep sentences "There is no entry" / "There is no tag" when there is no entry in starred and archive sections, search results or no created tag in tags page.

nicosomb:
Done in #2002.

PR #1963, opened 2016-04-21
Display entries number for each category

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | comma-separated list of tickets fixed by the PR, if any
| License       | MIT

Added #1789 

capture d ecran 2016-04-21 a 22 25 37


src/Wallabag/CoreBundle/Twig/WallabagExtension.php
```diff
@@ -16,6 +28,27 @@ public function removeWww($url)
         return preg_replace('/^www\./i', '', $url);
     }
 
+    public function getGlobals()
```
nicosomb: I want to have @j0k3r opinion about this new method. 
Do we need test for it? Because I only use `getBuilderFor*ByUser`, already tested.

nicosomb:
We need to discuss about this feature. @tcitworld told me that we can maybe use cache. Any idea @j0k3r?

j0k3r:
I don't really like to have this kind of magical variables in the Twig template (ie: `unreadEntries`, `allEntries`, etc.) which came from nowhere.
But I don't have better solution atm. Thinking ...

Also yes we definitely need to cache these values because it'll query the database on every page refresh. And since it's a kind of `select *` then `count`, it costs.

nicosomb:
I opened #2002. Wrong destination branch.

PR #1990, opened 2016-04-27
Add CLI import for wallabag v2 files

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| Fixed tickets | N/A
| License       | MIT

* Add CLI import for wallabag v2 with the ``--importer=v2`` option
* Add *mark all as read** support with the ``--markAsRead=yes`` option
* BC breaks with old importer because I change the command name
* No tests on this command because we only use v1/v2 importers which are already tested.

### DIFF
--- a/src/Wallabag/ImportBundle/Command/ImportCommand.php
+++ b/src/Wallabag/ImportBundle/Command/ImportCommand.php
@@ -13,10 +13,12 @@ class ImportCommand extends ContainerAwareCommand
     protected function configure()
     {
         $this
-            ->setName('wallabag:import-v1')
+            ->setName('wallabag:import')
             ->setDescription('Import entries from a JSON export from a wallabag v1 instance')
             ->addArgument('userId', InputArgument::REQUIRED, 'User ID to populate')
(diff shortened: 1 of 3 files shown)


src/Wallabag/ImportBundle/Command/ImportCommand.php
```diff
@@ -35,6 +37,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         $wallabag = $this->getContainer()->get('wallabag_import.wallabag_v1.import');
+
+        if ('v2' === $input->getOption('importer')) {
+            $wallabag = $this->getContainer()->get('wallabag_import.wallabag_v2.import');
+        }
+
+        if ('yes' === $input->getOption('markAsRead')) {
```
j0k3r: Won't this be true / false instead ?

PR #2005, opened 2016-04-30
Fix reading speed not defined when user was created via config page

### LINKED ISSUE #2004: Reading speed not defined when creating user via config page
### Issue details

If I create a user via config page (tag "Add a new user"), this new user don't have reading speed in his configuration. When he'll add new articles, he will have an error when he'll display entries list.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2004 
| License       | MIT

Fix #2004

About tests, I need your opinion.

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/ConfigController.php
+++ b/src/Wallabag/CoreBundle/Controller/ConfigController.php
@@ -140,6 +140,7 @@ class ConfigController extends Controller
             $config->setItemsPerPage($this->getParameter('wallabag_core.items_on_page'));
             $config->setRssLimit($this->getParameter('wallabag_core.rss_limit'));
             $config->setLanguage($this->getParameter('wallabag_core.language'));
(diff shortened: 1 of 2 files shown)


j0k3r:
Is there a test for creating a user from the config page? If yes, we should check that the config is created with all the required fields.

nicosomb:
It's not a field ...

j0k3r:
It is https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Entity/Config.php#L81

nicosomb:
Yes it's a property, you're right. But it's not a field in the config page when we create a new user.
So what can I test?

j0k3r:
- retrieve the config object associated to the created user
- assert all _properties_ are defined and not empty

PR #2020, opened 2016-05-03
Fix number of entries in tag/list

### LINKED ISSUE #2006: Number of entries per tag is not correct in tag/list
### Issue details

On tags list, the number of entries per tag is wrong: it's in fact the total entries per tag, but it doesn't take into account the current user. 

We have to use the good query to fetch tags by label and by user.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2006 
| License       | MIT

Fix #2006

What kind of tests?

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/Tag.php
+++ b/src/Wallabag/CoreBundle/Entity/Tag.php
@@ -117,4 +117,16 @@ class Tag
     {
         return $this->entries;
     }
+
+    public function getEntriesByUser($userId)
+    {
+        $filteredEntries = new ArrayCollection();
+        foreach ($this->entries as $entry) {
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Entity/Tag.php
```diff
@@ -117,4 +117,16 @@ public function getEntries()
     {
         return $this->entries;
     }
+
+    public function getEntriesByUser($userId)
```
j0k3r: Could this be called `getEntriesByUserId` instead?

Issue #2031, opened 2016-05-05
filtering option only for unread articles


Strubbl:
### Issue details

filtering does not work for unread articles. If i click on my unread articles, open the filter side bar and then filter for something, it routes to the "all articles" section and displays archived and starred articles in the result. Altough i did not check those in the filter options. I want to apply my filter only on unread articles. Is that somehow possible?

Maybe we can add the option "unread" next to archived and starred?

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.3
* How did you install wallabag? Via `git clone` or by downloading the package?
* Last wallabag version that did not exhibit the issue (if applicable):
* php version:
* OS:
* type of hosting (shared or dedicated):
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):

### Steps to reproduce/test case

Please provide necessary steps for reproduction of this issue, or better the
reduced test case (without any external dependencies, if possible).

nicosomb:
Fixed in #2052.

PR #2052, opened 2016-05-09
Add unread filter to entries pages

danbartram:
## New feature: Filter Unread
This PR allows users to filter their entries and requiring items to be "unread".

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes*
| Documentation | no
| Translation   | yes
| Fixed tickets | #2031
| License       | MIT

_* I'm getting a couple of PHPUnit test errors/failures, but they also exist on versions without this change._

----

### Possible Issues
There is some coupling caused by the SQL query that might cause future issues if the schema/Doctrine design changes.

Currently in [EntryFilterType.php](https://github.com/wallabag/wallabag/blob/6be103cd39e90849ab614d65bc8ad09f3507a206/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php#L83) the column for finding the read/unread status is also the archived status column. This uses the same logic as the "Unread" page view.

However, in the query you can see that the field `e.isArchived` is hard-coded and won't change if/when the Doctrine models' parameters do.

If there is a better way to design this please let me know, this solution works but it doesn't seem the best design to me.

### DIFF
--- a/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
+++ b/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
@@ -87,6 +87,18 @@ class EntryFilterType extends AbstractType
             ->add('isStarred', CheckboxFilterType::class, [
                 'label' => 'entry.filters.starred_label',
             ])
+            ->add('isUnread', CheckboxFilterType::class, [
+                'label' => 'entry.filters.unread_label',
+                'apply_filter' => function (QueryInterface $filterQuery, $field, $values) {
+                    if (false === $values['value']) {
+                        return;
+                    }
+
+                    $expression = $filterQuery->getExpr()->eq('e.isArchived', 'false');
+
+                    return $filterQuery->createCondition($expression);
+                },
+            ])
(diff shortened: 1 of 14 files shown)


nicosomb:
Hello @danbartram!
Thanks for your PR. 

Can you add a new test like this one please? https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Tests/Controller/EntryControllerTest.php#L455

What happened if you tick the new checkbox and the Archived one?

danbartram:
@nicosomb I sure can, sorry for missing the test out. How would you like it resubmitted? As a new PR or as an extra commit?

nicosomb:
An extra commit ;-)
Le 9 mai 2016 18:00, Dan Bartram <notifications@github.com> a écrit :@nicosomb I sure can, sorry for missing the test out. How would you like it resubmitted? As a new PR or as an extra commit?

j0k3r:
:+1:

PR #2103, opened 2016-05-19
Manage assets through npm

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

**How it works now : 
Run `./install.sh` to run composer, npm, bower and grunt all in a row.**

Still needs to be done :

- [x] Side-panels on material themes


install.sh
```diff
@@ -0,0 +1,14 @@
+#! /usr/bin/env bash
+
+if [[ $ASSETS == 'nobuild' ]]; then
+    composer install --no-interaction --no-progress --prefer-dist -o
+else
+    echo "Installing PHP dependencies through Composer..."
+    composer install
```
j0k3r: Could you still do `--no-interaction --no-progress --prefer-dist -o` ?
Also, is it this command and user will run to install deps ?

tcitworld: I thought the user may prefer having the output of this command.

Yup, the user will just have to run `./install.sh`. I can add the `wallabag:install` command to the script too.

j0k3r: So the composer install command should be `SYMFONY_ENV=prod composer install --no-dev -o --prefer-dist`.
But this shouldn't be that command for test & dev :confused: 

Should we have a install_dev.sh ?

tcitworld: Suits me.

nicosomb:
About MathJax, I think we can add it after this PR. What do you think about that?

tcitworld:
Yup, it's too foggy on how to find a good/modern method on how to handle it anyway.

nicosomb:
OK. About side panels, do you think you can do it soon?

tcitworld:
Will have a look over it over this week-end. Please test everything in meanwhile. ;-)

tcitworld:
ping @modos189 ?

tcitworld:
Side panels are fixed by downgrading the materialize version.

tcitworld:
And it's cleaned. :)

PR #2122, opened 2016-05-28
Fix the deletion of Tags/Entries relation when delete an entry

### LINKED ISSUE #2121: SERIOUS BUG - Tags clears on all articles after deleting article contains tags
### Issue details

When I delete article with tags, this tag clears from all other articles! It deleted also from DB, both from w_tag and from w_entry_tag tables
That behaviour is as in site and also in API.
(I'v just lost tag "programming" :-(  )

### Environment

* wallabag version 2.0.4

Don't check this with your existing tag...... just with new one.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2121 
| License       | MIT

Fix #2121

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -23,10 +23,16 @@ class EntryController extends Controller
     {
         try {
             $entry = $this->get('wallabag_core.content_proxy')->updateEntry($entry, $entry->getUrl());
+
             $em = $this->getDoctrine()->getManager();
             $em->persist($entry);
             $em->flush();
         } catch (\Exception $e) {
+            $this->get('logger')->error('Error while saving an entry', [
+                'exception' => $e,
+                'entry' => $entry,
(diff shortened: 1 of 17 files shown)


src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -23,10 +23,16 @@ private function updateEntry(Entry $entry)
...
+
             $em = $this->getDoctrine()->getManager();
             $em->persist($entry);
             $em->flush();
         } catch (\Exception $e) {
+            $this->get('logger')->error('Error while saving an entry', [
+                'exception' => $e,
+                'entry' => $entry,
+            ]);
+
```
nicosomb: :+1:

nicosomb:
It seems to be OK for me.

Issue #2077, opened 2016-05-13
Filtering by reading time does not work if only one limit is set


Strubbl:
### Issue details

I am filtering articles by reading time. if i only set the upper limit e.g. to 5 minutes and no lower limit, all articles are found. I would have expected that only articles lower equals 5min reading time were displayed. 

Workaround for this issue is to set the lower limit to zero minutes. Then the search works as expected. 

This issue works also the other way round (only set the lower limit and no upper).

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.4
* How did you install wallabag? Via `git clone` or by downloading the package? git clone
* Last wallabag version that did not exhibit the issue (if applicable):
* php version: 5.6.20
* OS: Debian Jessie
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MySQL

### Steps to reproduce/test case

Please provide necessary steps for reproduction of this issue, or better the
reduced test case (without any external dependencies, if possible).

PR #2157, opened 2016-06-17
Handle only upper or only lower reading filter

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | ?
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2077
| License       | MIT

When we select only one value in the reading time filter, we need to perform a query with only one value (greater than OR lower than).

Should fix https://github.com/wallabag/wallabag/issues/2077


nicosomb:
:+1:

PR #2166, opened 2016-06-23
Add tags on entries view

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Documentation |no
| Translation   | yes
| Fixed tickets | #1995
| License       | MIT

tags on entries view

Should be tested on old browsers

Issue #1995, opened 2016-04-29
Display tags on entries view


nicosomb:
When I'm on entries view (unread, fav, archived), I want to see the tags of each entry.

tcitworld:
Done it for material theme.

PR #2170, opened 2016-06-23
Add tags on entries view and minor ui fixes

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Documentation |no
| Translation   | yes
| Fixed tickets | #1995
| License       | MIT

tags on entries view

Should be tested on old browsers.

Originally at #2166

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/translations/messages.en.yml
+++ b/src/Wallabag/CoreBundle/Resources/translations/messages.en.yml
@@ -144,6 +144,7 @@ entry:
         reading_time: 'estimated reading time'
         reading_time_minutes: 'estimated reading time: %readingTime% min'
         reading_time_less_one_minute: 'estimated reading time: <small class="inferieur">&lt;</small> 1 min'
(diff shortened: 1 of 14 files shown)


src/Wallabag/CoreBundle/Resources/translations/messages.en.yml
```diff
@@ -144,6 +144,7 @@ entry:
         reading_time: 'estimated reading time'
         reading_time_minutes: 'estimated reading time: %readingTime% min'
         reading_time_less_one_minute: 'estimated reading time: <small class="inferieur">&lt;</small> 1 min'
+        number_of_tags: '{1}and one other tag|]1,Inf[and %count% other tags'
```
j0k3r: Could you add that line into all others translated file but commented? It'll ease the translation.

nicosomb:
I think we need to change the tag display in an article view, to have the same display than here.

PR #2176, opened 2016-06-25
Add since parameter

### LINKED ISSUE #1982: API improvements
There are some things I'm missing right now at the moment:

- [x] delete tag by slug
- [x] delete multiple tags with one request
- [x] parameter `since` in `/api/entries` to get all items that were created or updated after a specific date/timestamp

Additionally I think the `id` property for each tag is quite senseless. It can be used internal for the database index, but the API doesn't need to have access on it. `slug` and `label` should be enough.

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Documentation |no
| Translation   |no
| Fixed tickets | #1982
| License       | MIT

May need documentation updates.

Fix #1982

### DIFF
--- a/src/Wallabag/CoreBundle/Repository/EntryRepository.php
+++ b/src/Wallabag/CoreBundle/Repository/EntryRepository.php
@@ -95,7 +95,7 @@ class EntryRepository extends EntityRepository
      *
      * @return array
      */
-    public function findEntries($userId, $isArchived = null, $isStarred = null, $sort = 'created', $order = 'ASC')
+    public function findEntries($userId, $isArchived = null, $isStarred = null, $sort = 'created', $order = 'ASC', $since = 0)
     {
         $qb = $this->createQueryBuilder('e')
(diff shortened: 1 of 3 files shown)


nicosomb:
:+1:

src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -95,7 +95,7 @@ public function getBuilderForStarredByUser($userId)
      *
      * @return array
      */
-    public function findEntries($userId, $isArchived = null, $isStarred = null, $sort = 'created', $order = 'ASC')
+    public function findEntries($userId, $isArchived = null, $isStarred = null, $sort = 'created', $order = 'ASC', $since = 0)
```
j0k3r: You should update the phpdoc too

Issue #1640, opened 2016-02-01
Delete account


nicosomb:
In his dashboard (see #1638), a user can delete its account. 

@j0k3r @tcitworld do we need to send a confirmation email here?

j0k3r:
I don't think so.
We can like Github did when you want to delete a repo, just ask at least twice to the user that (s)he is going to remove its account and it can not be undo.

nicosomb:
Good idea :+1:

zertrin:
A typical solution to that is to ask the user to type a string like `yes_delete` or `<username>` as a confirmation. This can't be easily done by mistake and force the user to really read the information given on the warning.

For example GitLab has it when deleting a project and it looks like this:

![fewnuiafhewil](https://cloud.githubusercontent.com/assets/1447163/12721869/30b5ae68-c902-11e5-9897-35b2c8fd8fdf.png)

Something similar but with the username could be done for Wallabag.

j0k3r:
@zertrin yeah it almost the same as Github: 

![](https://cloud.githubusercontent.com/assets/62333/12721958/b3cc0a72-c902-11e5-8301-2ed6a61d0ec5.png)

zertrin:
Oh ok, I didn't try that in GitHub yet, so didn't know they are so close from each other. Sorry for the noise then.

nicosomb:
FYI, @delyriand is working on this feature (it's just to say to other contributors that it's not necessary to work on this feature)

nicosomb:
@delyriand need some help to finish this feature?

tcitworld:
@delyriand @nicosomb Any news on this ?

nicosomb:
FYI, I think this feature is no more WIP.

j0k3r:
One user can't deleted its own account.

nicosomb:
So we can close this issue?

j0k3r:
Nope.
An administrator can delete a user account.
But need to implement the account deletion on the user side.

nicosomb:
I read "one use can deleted ...". 💤

v1kn:
@j0k3r 
I'd like to ask why this issue was closed? I want to be able to have my wallabag account completely deleted, yet I cannot do it, because:

> You are the only user, you cannot delete your own account

Who can delete my account then? I use framabag

j0k3r:
Framabag isn't running wallabag 2, but wallabag 1.
poke @tcitworld

v1kn:
[...]

PR #2177, opened 2016-06-26
Delete user

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Documentation |no
| Translation   | yes
| Fixed tickets | #1640
| License       | MIT

### DIFF
--- a/src/Wallabag/UserBundle/Entity/User.php
+++ b/src/Wallabag/UserBundle/Entity/User.php
@@ -25,7 +25,7 @@ use Wallabag\CoreBundle\Entity\Entry;
  * @UniqueEntity("email")
  * @UniqueEntity("username")
  */
-class User extends BaseUser implements TwoFactorInterface, TrustedComputerInterface
+class User extends BaseUser implements TwoFactorInterface, TrustedComputerInterface, \Serializable
 {
     /**
      * @var int
@@ -240,4 +240,14 @@ class User extends BaseUser implements TwoFactorInterface, TrustedComputerInterf
 
         return false;
     }
+
+    public function serialize()
+    {
+       return serialize($this->id);
+   }
+
+   public function unserialize($serialized)
+   {
+       $this->id = unserialize($serialized);
+   }
 }
--- a/src/Wallabag/UserBundle/Entity/User.php
+++ b/src/Wallabag/UserBundle/Entity/User.php
@@ -243,11 +243,11 @@ class User extends BaseUser implements TwoFactorInterface, TrustedComputerInterf
 
     public function serialize()
     {
-       return serialize($this->id);
-   }
+        return serialize($this->id);
+    }
 
-   public function unserialize($serialized)
-   {
-       $this->id = unserialize($serialized);
-   }
+    public function unserialize($serialized)
+    {
+        $this->id = unserialize($serialized);
+    }
 }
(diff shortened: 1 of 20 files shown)


nicosomb:
Need tests?

tcitworld:
Yup, and not working properly right now.

nicosomb:
We need to add a confirmation step too.

j0k3r:
Rebased ✅

nicosomb:
If I'm an admin, I don't want to see this button in my config page.

tcitworld:
> If I'm an admin, I don't want to see this button in my config page.

Why ? There can be multiple admin accounts and an admin has not necessary shell access to do this as he/she likes.

nicosomb:
Today, we can't specify who is an admin. Only the first user is an admin, so if we delete him, we can create an other one.

tcitworld:
In that case, we should count the admin accounts and hide the button if this admin is the only one.

nicosomb:
why not. The easiest solution is the best for me.

tcitworld:
> why not. The easiest solution is the better one for me.

It's to prepare the future. :-)

tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
```diff
@@ -649,4 +649,89 @@ public function testDemoMode()
...
+    }
+
+    public function testDeleteAccount()
+    {
+        $this->logInAs('wallace');
+        $client = $this->getClient();
+
+        $crawler = $client->request('GET', '/config');
+
+        $deleteLink = $crawler->filter('.red')->last()->link();
```
j0k3r: Can't there be a better selector which are not going to fail when we'll add a new `red` class on an element?

tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
```diff
@@ -649,4 +649,89 @@ public function testDemoMode()
...
+        $em = $client->getContainer()->get('doctrine.orm.entity_manager');
+        $user = $em
+            ->getRepository('WallabagUserBundle:User')
+            ->createQueryBuilder('u')
+            ->where('u.username = :username')->setParameter('username', 'wallace')
+            ->getQuery()
+            ->getOneOrNullResult()
+        ;
+
+        $this->assertTrue(false !== $user);
```
j0k3r: `assertNotNull($user)` ?

nicosomb:
@modos189 We need popin like here https://github.com/wallabag/wallabag/issues/1640 Can you help us please?

src/Wallabag/UserBundle/Entity/User.php
```diff
@@ -240,4 +240,14 @@ public function isTrustedComputer($token)
 
         return false;
     }
+
+    public function serialize()
+    {
+        return serialize($this->id);
```
j0k3r: @tcitworld why do we need to serialize the user?

j0k3r:
Here are some preview:

src/Wallabag/UserBundle/Entity/User.php
tcitworld: I don't remember exactly but I guess it was required for something. Please test without it.

j0k3r: Removed it changed nothing.

tcitworld:
Looks good.

Issue #2033, opened 2016-05-06
Download pictures on my webserver


nicosomb:
When I save a new article, I want that wallabag downloads article's pictures in my `data` folder. 

This feature already exists in wallabag v1.x.

Alixerid:
Related to this feature: I saw that browsers show the "mixed content" warning despite the wallabag installation being secured with TLS (https). Pulling the images from third party servers and not serving them from wallabag instance should be the reason for the warning.
i.e. +1 for this also due to TLS security.

PR #2180, opened 2016-06-28
Download pictures

tcitworld:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | no |
| Documentation | no |
| Translation | no |
| Fixed tickets | #2033 |
| License | MIT |

Needs a setting somewhere. And a correct way to get baseurl from the server. And tests !


composer.json
```diff
@@ -81,14 +81,15 @@
...
+        "mouf/nodejs-installer": "~1.0",
+        "symfony/dom-crawler": "^3.1"
     },
     "require-dev": {
         "doctrine/doctrine-fixtures-bundle": "~2.2",
         "doctrine/data-fixtures": "~1.1.1",
         "sensio/generator-bundle": "^3.0",
         "phpunit/phpunit": "~4.4",
-        "symfony/phpunit-bridge": "^3.0",
+        "symfony/phpunit-bridge": "^2.8.7",
```
j0k3r: ?

tcitworld: Hmm, I don't know why this.

tcitworld:
Since an entry isn't persisted when we're doing this stuff, we can't create subfolders named after the entry's id. Therefore :
* either they're all inside `'assets/images'`, with the risk of collisions if two pictures from different entries have the same name
* either we generate foldernames based on a hash of the url

j0k3r:
Rebased ✅

j0k3r:
So, I've updated the PR:

- downloading images append when an entry in persisted / updated
- using prePersit & preUpdate event from Doctrine
- previewPicture is also updated
- images are now saved to `web/assets/images` so they can be accessible in the UI

Todo:

- [ ] download images using Rabbit or Redis
- [ ] define the host (we can't use `/assets/images` if user defined wallabag in a subfolder, ie: `http://myhost.com/wallabag_instance/`)
- [ ] add more tests

j0k3r:
@wallabag/core should we enable async import for images?
I'm asking because if someone add an entry, it'll take time to download images (but it can acceptable).
If people are already using async to import entries, it won't hurt the global experience if images are downloaded after saving a new entry.

So, is there a good reason to enable async downloading images?
Or should we ship that feature without async and see if people are complaining? :)

tcitworld:
If an user accesses an entry of which it's pictures has not been fetched yet, won't it be ugly to display ?

j0k3r:
Nope because it'll display images of the content, like today.

nicosomb:
👍 
Why do we store the absolute path for images? If we forgot to configure the wallabag URL in internal settings (by using the default configuration), all paths are wrong.

j0k3r:
We store the absolute path only if the `wallabag_url` is defined. Otherwise it'll be stored as `/assets/images/..` a relative link. It should work most of the time. But it's just to handle people who defined their wallabag in a subdomain.

nicosomb:
As discussed on gitter, we need to handle the article's deletion.

j0k3r:
- [ ] article's deletion
- [ ] still needs to handle import

j0k3r:
So enable downloading image on import when you didn't want them to be async is a killer app. If you got one article with many images, it'll take ages to import...
Maybe we should put a warning somewhere for people who don't enable async BUT enable downloading image when they arrive on the import page.

nicosomb:
> Maybe we should put a warning somewhere for people who don't enable async BUT enable downloading image when they arrive on the import page.

New issue?

j0k3r:
It was more a message to got your feedback :)
I can add this message in that PR.

nicosomb:
I think you can add it, you're right.

j0k3r:
Warning message added:

image

Are you ok with translations?

- EN: You enabled downloading images for your articles. Combined with classic import it can take ages to proceed (or maybe failed). We <strong>strongly recommend</strong> to enable asynchronous import to avoid errors.
- FR: Vous avez configuré le téléchagement des images pour vos articles. Combiné à l'import classique, cette opération peut être très très longue (voire échouer). Nous vous conseillons <strong>vivement</strong> d'activer les imports asynchrones.

tcitworld:
Also, it could take a lot of disk space.

j0k3r:
This is a different message we should put.
Maybe in the config description instead?

nicosomb:
👍

Narga:
I've looked for around but still don't find any document to help me enable this feature. How can I active this?

j0k3r:
Could you open an issue so we can add this into the documentation?

andrey-str:
Was an issue opened? If so could you link it tk this ticket?
I also can not find where to enable this option.

j0k3r:
As far as I remember no issue were open

Issue #1913, opened 2016-04-12
FR: Installing as not web root


gerroon:
### Issue details

Hi

Can you please provide a way to install v2 as a non web root app? At the moment the installation setup is done in a way that assumes that one will only have Wallabag as the only app on the server. I was able to run wallabag v1 as a normal server folder without further complications. I feel like v2 has made it complicated for users who want to host their version who also are not very high level web professionals.

Please bear in mind that not everyone is able to understand how Symphony or Ruby etc work. So I am hoping that the devs would consider making self hosting easier for the rest of us.

thanks 
### Environment
- wallabag version (or git revision) that exhibits the issue: 
  v2
- How did you install wallabag? Via `git clone` or by downloading the package?
  git clone 
- Last wallabag version that did not exhibit the issue (if applicable):
  v1
- php version:
- OS:
  linux 
- type of hosting (shared or dedicated):
  self
- which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):
  sqlite
### Steps to reproduce/test case

Please provide necessary steps for reproduction of this issue, or better the
reduced test case (without any external dependencies, if possible).

nicosomb:
Hello @gerroon! 

Can you try the **Installation on a shared hosting** process (here: https://www.wallabag.org/pages/download-wallabag.html) and tell me if it's OK for you please?

gerroon:
@nicosomb 


Thanks for the follow up. I am sorry that I did not mention a critical information. I was trying to use it under apache and the current virtual hosting setup (the one you mentioned) assumes that it should be installed as the web root. I personally do not know how to not to affect my current apache setup with this new virtual host setup that is in the doc. If I go with that I wont be able to access my other web applications on my server.

I hope it makes sense.

PeterTGarner:
> installation setup is done in a way that assumes that one will only have Wallabag as the only app on the server ... not sure if I can have multiple document roots under the same virtualhost (*:443 in my case)

I agree with Gerroon. I wouldn't say I'm a noob at setting up a web server, but I'm not an expert either. My specific issue is that my web server is https-only, so I don't have a wildcard certificate for subdomains. Is there a way to make v2 addressable in the same as v1? 

Wallabag is a great piece of software and I use it constantly - thank you for your work.

**Environment**

* wallabag version (or git revision) that exhibits the issue: 
v2.0.3

* How did you install wallabag? Via git clone or by downloading the package?
git clone

[...]

PR #2181, opened 2016-06-29
Add configurable base_path parameter (url prefix)

alanpearce:
This allows running wallabag from a directory other than the web root. 

For some reason I couldn't get the test suite to work properly even on master, but as this is a set of configuration changes, I'm not sure it's particularly testable.

I would say that it requires some documentation in the installation section under "Virtual Hosts", but right now I'm still working on the nginx configuration

| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | no |
| Documentation | no |
| Translation | no |
| Fixed tickets | #1913 |
| License | MIT |

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -1,31 +1,39 @@
 wallabag_annotation:
   type : rest
   resource: "@WallabagAnnotationBundle/Resources/config/routing_annotations.yml"
+  prefix: "%base_path%/"
 
 wallabag_import:
     resource: "@WallabagImportBundle/Controller/"
     type: annotation
-    prefix: /import
+    prefix: "%base_path%/import"
 
 wallabag_api:
     resource: "@WallabagApiBundle/Resources/config/routing.yml"
-    prefix: /
+    prefix: "%base_path%/"
 
 app:
     resource: "@WallabagCoreBundle/Controller/"
     type: annotation
+    prefix: "%base_path%/"
 
 doc-api:
     resource: "@NelmioApiDocBundle/Resources/config/routing.yml"
-    prefix: /api/doc
+    prefix: "%base_path%/api/doc"
 
 rest :
     type : rest
     resource : "routing_rest.yml"
-    prefix : /api
+    prefix : "%base_path%/api"
+
+homepage_main:
+    path: "%base_path%/"
+    defaults:
+        _controller: WallabagCoreBundle:Entry:showUnread
+        page: 1
 
 homepage:
-    path: "/{page}"
+    path: "%base_path%/{page}"
     defaults:
         _controller: WallabagCoreBundle:Entry:showUnread
         page : 1
@@ -34,11 +42,13 @@ homepage:
 
 fos_user:
     resource: "@FOSUserBundle/Resources/config/routing/all.xml"
+    prefix: "%base_path%/"
 
 fos_oauth_server_token:
     resource: "@FOSOAuthServerBundle/Resources/config/routing/token.xml"
+    prefix: "%base_path%/"
 
 craue_config_settings_modify:
-    path: /settings
+    path: "%base_path%/settings"
     defaults:
         _controller: CraueConfigBundle:Settings:modify
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -19,11 +19,11 @@ security:
     firewalls:
         # disables authentication for assets and the profiler, adapt it according to your needs
         dev:
-            pattern: ^/(_(profiler|wdt)|css|images|js)/
+            pattern: "^%base_path%/(_(profiler|wdt)|css|images|js)/"
             security: false
 
         oauth_token:
-            pattern: ^/oauth/v2/token
+            pattern: "^%base_path%/oauth/v2/token"
             security: false
 
         api:
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -33,7 +33,7 @@ security:
             anonymous: true
 
         login_firewall:
-            pattern: ^/login$
+            pattern: "^%base_path%/login$"
             anonymous:  ~
 
         secured_area:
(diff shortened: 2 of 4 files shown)


tcitworld:
Well tests seems to run fine anyway. Thanks !

app/config/routing.yml
```diff
@@ -1,31 +1,39 @@
...
     type : rest
     resource : "routing_rest.yml"
-    prefix : /api
+    prefix : "%base_path%/api"
+
+homepage_main:
+    path: "%base_path%/"
+    defaults:
+        _controller: WallabagCoreBundle:Entry:showUnread
+        page: 1
```
j0k3r: Why did you create this new route?

alanpearce: Somehow the homepage route isn't matched when the base url is used alone with a trailing slash:

```
$ bin/console route:match /reader/
                                                                                                     
 [ERROR] None of the routes match the path "/reader/"                                                
                                                                                                     
$ bin/console route:match /reader
                                                                                                     
 [OK] Route "homepage" matches                                                                                                                                                                          

```

This seems wrong to me. My solution is a hack, but it's been a year since I worked with Symfony, so I tried the simplest thing that could work. I'd gladly accept input on finding an alternative, though.

app/config/security.yml
```diff
@@ -33,33 +33,36 @@ security:
...
-        - { path: ^/, roles: ROLE_USER }
+        - { path: "^%base_path%/api/doc", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: "^%base_path%/api/version", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: "^%base_path%/login", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: "^%base_path%/register", role: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: "^%base_path%/resetting", role: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: "%base_path%/(unread|starred|archive).xml$", roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: "^%base_path%/settings", roles: ROLE_SUPER_ADMIN }
+        - { path: "^%base_path%/annotations", roles: ROLE_USER }
+        - { path: "^%base_path%/", roles: ROLE_USER }
```
alanpearce: Thanks to the comment above about the extra route, I noticed that the path matched by ^$ isn't secured and causes an exception to be thrown. I will make sure this is fixed once we've decided what to do about the extra route.

app/config/routing.yml
j0k3r: I understand the problem but do we have to handle homepage url with trailing slash?

alanpearce: I think so, yes. If /base_url/x/y matches a route, /base_url/ should match another route.  Personally, I wouldn't consider /base_url to be a valid point of access if it has further routes underneath and would rather not even forward /base_url to Wallabag. FOSUserBundle appears to work this way. However, the rest of Wallabag appears to be inconsistent, with parent routes /developer, /import/ and /import/pocket showing up in the routes list.

Whilst I could technically work around the problem with a URL rewrite in the web server, I would find it surprising that I would have to add such a rule, and it would probably be a hurdle for other users trying to set this up, which may cause more support queries, etc.

app/config/security.yml
j0k3r: Did you got some spare time to fix them?

jpalczewski:
Hi,

I'm checking in with freshly configured and working wallabag behind SSL reverse proxy based on that pull request - I'm a bit curious if one dirty hack that I had done can be solved in another way.
 
Short briefing: the problem which I was facing were redirects from /services/wallabag to /login - at first I tried to deal with it with proxy_redirects, however I found that pull request and ended with another problem that might appear in another PR - app worked, but all graphics and css were missing. 

So, my fix is:

1. base path parameter:
`    base_path: services/wallabag/`
2. omitted virtual host - I'm playing with nginx and virtual host is on some localhost port.
3. reverse proxy:
 ```
location /internal/wallabag/ {

        rewrite ^/services/wallabag(/bundles/.*)$ $1 break;
proxy_intercept_errors on; 
proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto https;
                        proxy_pass http://localhost:59993/services/wallabag/;
   proxy_redirect default;
                }
```

**Comment by alanpearce** [comment]
I had the same problem with assets. I had to add:

```
        location /reader/bundles {  
                alias /opt/wallabag/web/bundles;  
        }  
```

I'm not sure how to improve this situation. It would be nice if this rule were unnecessary. I'll try to look at the asset config soon.

jpalczewski:
Thanks, it's looking a bit better. 

Despite that, I have another question: do you tested access by API?(Or do you have any hints where should I look for adequate logs). Firefox extensions works rather good, however I can't achieve oauth token [in way described in wiki](http://doc.wallabag.org/en/latest/developer/api.html) - I'm getting 302 and redirect to `<path>/login` and don't know what to do. When trying to access `<path>/oauth/v2/token` from browser I'm getting `{"error":"invalid_request","error_description":"Invalid grant_type parameter or parameter missing"}`, so it's working when I'm logged.

j0k3r:
@jpalczewski maybe you can ask this in a separate issue where you'll describe each step you are trying to achieve to get that result? :)

nicosomb:
I really don't know what we can do with this PR. 
As I said in #1913, i tried and it works for me.

alanpearce:
I'm still working on the API access. I pushed a couple of updates, but I can't seem to get it to work with the android app just yet.

callmetango:
@alanpearce I had a "nice" weekend with the attempt to install Wallabag 2.2.1 in a sublocation of my localhost. Wish I had been aware of your patch... After lots of debugging I came to a nearly identical solution which I will attach as a patch in a minute.

I also commented on your pull request regarding the non-working Android client requests. There is simply a regex in the Symfony firewall config that's causing trouble.

callmetango:
Patch to add base path for Wallabag 2.2.1: https://gist.github.com/callmetango/7b143faa5a5b008e6df0d287b7a7ceec

alanpearce:
Ah, thanks for that @callmetango. I wondered why the android app stopped working at some point. I think it might make more sense to create a PR from your patch, since it seems like this branch is now out-of-date and you've already done most of the work 👍

callmetango:
> I think it might make more sense to create a PR from your patch, since it seems like this branch is now out-of-date and you've already done most of the work

@alanpearce Done. Added the updated patch for the configurable base path as PR #2895.

j0k3r:
Duplicate of https://github.com/wallabag/wallabag/pull/2895

Issue #1873, opened 2016-04-06
Disable public user registration


tmelikoff:
It would be really useful to disable public user registration and allowing only the administrator to create new accounts. I believe that some of us use wallabag only for personal use (or family).

Also, if such option (configuration file or switch in control panel) were enabled, registration button should be hidden in the homepage.

Thank you.

NerosTie:
Until there is a solution and if you use Nginx, add this in your configuration:

```
    location ^~ /register {
      deny all;
      return 403;
    }
```

ebal:
This is the work-around for httpd-2.2.15

     <LocationMatch "/register">
         Deny from All
     </LocationMatch>

fabsh:
Uhm. This is a rather important feature, I would think. I'm guessing a lot of people install this app to for their own use only. I would like to be able to prevent other people from using my server resources. Version 2.x suggest production ready to me. But an app with such a glaring oversight can hadly be used in production...

nicosomb:
@j0k3r what is the best way? A parameter (true/false)? If true, we hide the register button and we override the register route like here http://stackoverflow.com/a/32332378?

j0k3r:
@nicosomb Yeah could a good idea. An extra parameter in internal configuration.

nicosomb:
I think it's the easier way. OK, I change the milestone.

nicosomb:
Done by @tcitworld in #2186 👍 .

danicotillas:
@nicosomb I can´t find this option on the Configuration panel, How can i disable it?

tcitworld:
@danicotillas It's in the `app/config/parameters.yml` file, named `fosuser_registration`.

danicotillas:
@tcitworld Thanks, I change it to "false" but it still appears... What are the changes I should make?

tcitworld:
Clear the cache `bin/console cache:clear -e prod`

ecodiv:
I have changed the parameters.yml as suggested. But how do I clear the cache using the above? Do I need command line access I have Wallabag installed on a shared webhosting site, installed via Softaculous script installer)?

ecodiv:
OK, got it, doing:

1) Open the file in the folder `app/config/parameters.yml`
2) Set `fosuser_registration = false`
3) connect to your server using ssh (command line)
4) Go to the root folder of your Wallabag installation
5) Run `php bin/console cache:clear -e prod`

Would still be nice to have this as a click-able option on the internal configuration panel.

[...]

PR #2186, opened 2016-07-02
Add option to disable registration

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   |no
| Documentation |no
| Translation   |no
| Fixed tickets | #1873
| License       | MIT

### DIFF
--- a/src/Wallabag/UserBundle/Controller/RegistrationController.php
+++ b/src/Wallabag/UserBundle/Controller/RegistrationController.php
@@ -13,6 +13,6 @@ class RegistrationController extends FOSRegistrationController
             return parent::registerAction($request);
         }
 
-        return $this->redirectToRoute('fos_user_security_login', array(), 301);
+        return $this->redirectToRoute('fos_user_security_login', [], 301);
     }
 }
--- a/src/Wallabag/UserBundle/Controller/RegistrationController.php
+++ b/src/Wallabag/UserBundle/Controller/RegistrationController.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Wallabag\UserBundle\Controller;
+
+use FOS\UserBundle\Controller\RegistrationController as FOSRegistrationController;
+use Symfony\Component\HttpFoundation\Request;
+
+class RegistrationController extends FOSRegistrationController
+{
+    public function registerAction(Request $request)
+    {
+        if ($this->container->getParameter('wallabag_user.registration_enabled')) {
+            parent::registerAction($request);
+        }
+        else
+        {
+            return $this->redirectToRoute('fos_user_security_login', array(), 301);
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Wallabag/UserBundle/Controller/RegistrationController.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public function registerAction(Request $request)
+    {
+        if ($this->container->getParameter('wallabag_user.registration_enabled')) {
+            parent::registerAction($request);
+        }
+        else
+        {
+            return $this->redirectToRoute('fos_user_security_login', array(), 301);
+        }
```
j0k3r: Could you return the `parent::registerAction($request);` so  you can remove the else?

src/Wallabag/UserBundle/Controller/SecurityController.php
```diff
@@ -0,0 +1,18 @@
+<?php
```
j0k3r: Why do you need this new file?

src/Wallabag/UserBundle/Controller/RegistrationController.php
tcitworld: Well, I want to return http 301, what will happen otherwise ?

src/Wallabag/UserBundle/Controller/SecurityController.php
tcitworld: In order to show - or not - the Register button on the login page.

src/Wallabag/UserBundle/Controller/RegistrationController.php
j0k3r: I mean:

```php
if ($this->container->getParameter('wallabag_user.registration_enabled')) {
    return parent::registerAction($request);
}

return $this->redirectToRoute('fos_user_security_login', array(), 301);
```

src/Wallabag/UserBundle/Controller/SecurityController.php
j0k3r: :+1: 
Maybe add a comment in the class phpdoc about that.

src/Wallabag/UserBundle/Controller/RegistrationController.php
tcitworld: Oh yeah, sure.

j0k3r:
As of now, it's impossible to add test that handle both the registration enabled / disabled because you can't update the container parameter after getting the client. The ParameterBag is frozen ... 😕

j0k3r:
Rebased ✅

nicosomb:
I tried this PR, it works. 

> As of now, it's impossible to add test that handle both the registration enabled / disabled because you can't update the container parameter after getting the client. The ParameterBag is frozen 

Right, it's like the 2auth tests: we skip them if this feature is disabled. 
I added a simple test. When we'll be able to update the container parameter, we'll improve it.

j0k3r:
:+1:

Issue #2223, opened 2016-08-19
Uncaught Exception: Secured pdf file are currently not supported (Smalot/PdfParser)


etoulas:
### Issue details

Hi, I just installed the latest version to migrate from Pocket.
I'm importing thousands of links but very early I get an 500 when trying to download a "secured PDF".

I would appreciate even a temporary workaround to get at least all the remaining articles imported...

From `var/logs/prod.log`
```
[2016-08-19 22:37:10] graby.DEBUG: Graby is ready to fetch [] []
[2016-08-19 22:37:10] graby.DEBUG: Fetching url: {url} {"url":"http://www.iarc.fr/en/media-centre/pr/2015/pdfs/pr231_E.pdf"} []
[2016-08-19 22:37:10] graby.DEBUG: Trying using method "{method}" on url "{url}" {"method":"head","url":"http://www.iarc.fr/en/media-centre/pr/2015/pdfs/pr231_E.pdf"} []
[2016-08-19 22:37:10] graby.DEBUG: Data fetched: {data} {"data":{"effective_url":"http://www.iarc.fr/en/media-centre/pr/2015/pdfs/pr231_E.pdf","body":"(only length for debug): 0","headers":"application/pdf","status":200}} []
[2016-08-19 22:37:11] request.CRITICAL: Uncaught PHP Exception Exception: "Secured pdf file are currently not supported." at /home/et/repositories/wallabag2/vendor/smalot/pdfparser/src/Smalot/PdfParser/Parser.php line 94 {"exception":"[object] (Exception(code: 0): Secured pdf file are currently not supported. at /home/et/repositories/wallabag2/vendor/smalot/pdfparser/src/Smalot/PdfParser/Parser.php:94)"} []
```

### Environment

* wallabag version: git tag 2.0.6 installed via git clone
* php version: PHP 5.5.9-1ubuntu4.19
* OS: Ubuntu 14.04.5 LTS (trusty)
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MySQL

### Steps to reproduce/test case

Add the URL with the PDF that causes the issue: http://www.iarc.fr/en/media-centre/pr/2015/pdfs/pr231_E.pdf

NB: The error in the log file will look slightly different compared to importing from Pocket.

[...]

PR #2224, opened 2016-08-19
Avoid breaking import when fetching fail

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | fix https://github.com/wallabag/wallabag/issues/2223
| License       | MIT

graby will throw an Exception in some case (like a bad url, a restricted url or a secured pdf).

Import doesn't handle that case and break the whole import.
With that commit the import isn't stopped but the entry is just skipped.

Also, as a  bonus, I've added extra test on WallabagImportV2 when the json is empty.

### DIFF
--- a/src/Wallabag/ImportBundle/Import/AbstractImport.php
+++ b/src/Wallabag/ImportBundle/Import/AbstractImport.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Wallabag\ImportBundle\Import;
+
+use Psr\Log\LoggerInterface;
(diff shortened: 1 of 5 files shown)


j0k3r:
Tests are failing for 7.1 only with a segfault looks like a bug on Travis. We can avoid it.

Issue #2249, opened 2016-08-28
Update js & css assets


tcitworld:
There has some changes in css and js files lately. This has to be changed too in 2.1.

j0k3r:
As far as I know, I fixed most of them when fighting against the bad rebase on the 2.1 branch.
But yeah, could be a good idea to check before the release.

tcitworld:
Also, add linters to the build assets travis task (and Grunt watch).

j0k3r:
There is still (from https://github.com/wallabag/wallabag/pull/2314#issuecomment-250390210)

- [ ] Fix for twitter icons becoming facebook one (WTF ?)

nicosomb:
Done 👍 
Good job with the front part, @tcitworld <3

j0k3r:
Fixed here https://github.com/wallabag/wallabag/pull/2353

PR #2252, opened 2016-08-30
Fixed last merge who caused CSS/JS bugs

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

Last merge from master to 2.1 was buggy. 
I just need help, because jquery.tinydot from @modos189 doesn't seem to be loaded. 

Help needed @modos189 @tcitworld @j0k3r :)


j0k3r:
My bad if it was buggy. Mostly because stuff has to be done to include new lib in package.json as soon as they are added in the master branch.
And this PR is related to https://github.com/wallabag/wallabag/issues/2249

nicosomb:
@j0k3r not a problem 😗

j0k3r:
jquery.tinydot is now loaded for material.

tcitworld:
Instead of concating jquery.tinydot, require it through browserify in the main js file.

nicosomb:
👍

nicosomb:
> Instead of concating jquery.tinydot, require it through browserify in the main js file.

@tcitworld why?

tcitworld:
Because I find it better to use js modules instead of just concat all code. Moreover, minification *should* be better.

j0k3r:
@tcitworld tried both `require('tinydot');` 

>> Error: Cannot find module 'jQuery' from '/home/jeremy/github/wallabag/node_modules/jquery.tinydot/src'

`require('jquery.tinydot');`

>> Error: Cannot find module 'tinydot' from '/home/jeremy/github/wallabag/web/bundles/wallabagcore'

Without success.

j0k3r:
It looks too complex to load a jquery plugin with requirejs...
I think we'll keep it like this. I'll switch to the non minified version instead.

tcitworld:
> It looks too complex to load a jquery plugin with requirejs...

I'll have a look.

PR #2261, opened 2016-09-02
Fixes #2259

### LINKED ISSUE #2259: API: Not used tag exists in the db, but is not returned in API  GET /api/tags result.
### Issue details

After DELETE /api/entries/{entry}/tags/{tag}.{_format} tag rests in the db (I think that is OK), but after,
in GET /api/tags.{_format} result it doesn't exists. I think it would be useful to have it there (for autocomplete in my chrome ext :) )

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.6
* Last wallabag version that did not exhibit the issue (if applicable): >2
* which storage system you choose at install  MySQL

Rurik19:
| Q             | A
| ------------- | ---
| Bug fix?      | yes|
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #2259 
| License       | MIT


.gitignore
```diff
@@ -10,6 +10,9 @@
 /var/sessions/*
 !var/sessions/.gitkeep
 !var/SymfonyRequirements.php
+/bin/*
+!/bin/console
+!/bin/symfony_requirements
```
j0k3r: These lines already exists: https://github.com/wallabag/wallabag/blob/master/.gitignore

j0k3r:
Tests are failing because of a CS.
Run `php bin/php-cs-fixer fix src/ --verbose` locally and commit changes

nicosomb:
ping @Rurik19

Rurik19:
Hi! I am back :) 
bin/php-cs-fixer ->./vendor/fabpot/php-cs-fixer/php-cs-fixer
but there is no vendor in project, should I do "compose install" ?

j0k3r:
Yeap you need to `composer up`

j0k3r:
I'm wondering if it's a good idea to retrieve ALL tags using the API.
None of instapaper, readability, pocket or pinboard allow that. It's always in the user scope.

tcitworld:
Yeah, I think REST principles makes requests from the user's POV. An user shouldn't access to tags from other users.

Rurik19:
There is no field user_id in tags table, may be we should add it there? And then we will be able to get all tags with API including that tags-without-articles.

tcitworld:
IMHO, tags are more a characteristic of an entry than something on it's own. Therefore I don't see problems with tags being linked to users through entries only. And a tag without an entry doesn't makes any sense.

Rurik19:
Ok, that's reasonable. I close this PR, and #2259, but creating issue about orphaned tags :)

PR #2262, opened 2016-09-03
Add a check for the database connection

### LINKED ISSUE #2251: Add a requirement in InstallCommand to check database connection (after driver checking)
To decrease error when installing wallabag for _non-techie people_, we should check if the connection to the database can be etablished with information defined in parameters.

So it'll avoid error on the next step of the InstallCommand which is database creation.

image

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | fix #2251
| License       | MIT

Checking for the driver isn't enough.
We are now checking if we can etablish a connection to the database before trying to do anything.
By displaying the error from the Exception (in case of error) we hope to reduce issues overload about people getting error with the database

### DIFF
--- a/src/Wallabag/CoreBundle/Command/InstallCommand.php
+++ b/src/Wallabag/CoreBundle/Command/InstallCommand.php
@@ -72,8 +72,10 @@ class InstallCommand extends ContainerAwareCommand
     {
         $this->defaultOutput->writeln('<info><comment>Step 1 of 5.</comment> Checking system requirements.</info>');
 
-        $fulfilled = true;
+        $rows = [];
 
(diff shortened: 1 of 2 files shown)


nicosomb:
No need to change tests?

j0k3r:
Since some of them are failing I think so 😑

nicosomb:
👍

nicosomb:
:+1:

PR #2263, opened 2016-09-03
Instead of selecting the whole data, just count it

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets |
| License       | MIT

Instead of performing a complex select (to retrieve all data for entry, etc...) just select the counter and retrieve it.

Down from ~50ms to ~30ms on the unread page (with 500 items)

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/DeveloperController.php
+++ b/src/Wallabag/CoreBundle/Controller/DeveloperController.php
@@ -49,7 +49,7 @@ class DeveloperController extends Controller
 
             $this->get('session')->getFlashBag()->add(
                 'notice',
-                $this->get('translator')->trans('flashes.developer.notice.client_created', array('%name%' => $client->getName()))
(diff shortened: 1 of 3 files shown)


j0k3r:
> ERROR: column "e0_.id" must appear in the GROUP BY clause or be used in an aggregate function

F*** Y** PostgreSQL.

nicosomb:
No more cache for these queries?

j0k3r:
@nicosomb the cache is still here.

nicosomb:
OK, but not used, right? (it's just to know :) )

j0k3r:
Since all config about caching (ie: `result_cache_driver` etc.) aren't enable in config_*, nop it's not used.

nicosomb:
👍

Issue #2296, opened 2016-09-18
Invalid EPub export for Kobo devices when title contains '+'


morhelluin:
### Issue details

When exporting an entry with a title containing the `+` character to epub, the exported file does not work in Kobo devices.
Other characters may trigger the problem, but this is untested (titles with `:` have [some problems with another library](https://github.com/Grandt/PHPRelativePath/issues/3)).

### Environment

* wallabag version (or git revision) that exhibits the issue: git rev a707643 
* How did you install wallabag? `git clone`
* Last wallabag version that did not exhibit the issue (if applicable): N/A
* php version: PHP 7.0.10
* OS: Debian testing
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): SQLite

### Steps to reproduce/test case

1. Add an entry with the character `+` in its title, like https://en.wikipedia.org/wiki/C++.
2. Export this entry to the EPub format.
3. Copy the epub file to a Kobo device (tested with Kobo Glo HD).
4. Open the file with the Kobo device : the file opens, displaying only the first two pages : cover and table of content.
5. The content of the entry cannot be read, the device thinks it contains only two pages : the cover and the table of content. The content of the entry is ignored.

(6. Do the same steps, but first edit the entry's title to remove the `+` : the whole content of the file can now be displayed on the Kobo device.)

I'll do a pull request with a possible correction.

morhelluin:
Thanks!

PR #2297, opened 2016-09-18
Fix issue #2296: epub export with special chars in the title.

morhelluin:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2296 
| License       | MIT

I removed all special chars from the filename of an epub chapter, in case other characters break the file (not only `+`).

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/EntriesExport.php
+++ b/src/Wallabag/CoreBundle/Helper/EntriesExport.php
@@ -163,8 +163,12 @@ class EntriesExport
                 $book->setSubject($tag['value']);
             }
 
+            // the reader in Kobo Devices doesn't likes special caracters
+            // in filenames, we limit to A-z/0-9
+            $filename = preg_replace('/[^A-Za-z0-9\-]/', '', $entry->getTitle());
+
             $chapter = $content_start.$entry->getContent().$bookEnd;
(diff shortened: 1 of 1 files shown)


j0k3r:
:+1:

tcitworld:
We could have asked for fix upstream. @Grandt

nicosomb:
@tcitworld Maybe we can open a new issue at https://github.com/Grandt/PHPePub

Grandt:
Please do. 
I'm down with pneumonia at the moment though, so a response might have to wait till the weekend.

morhelluin:
@tcitworld You're right, I initially thought it would be weird for PHPePub to change the value of the $filename parameter (thus changing what the caller asked for), but I guess it's ok if it's for "sanitization" (it already [does that anyway](https://github.com/Grandt/PHPePub/blob/09f343c9e0da804e051953088664c8c56e0e9135/src/PHPePub/Core/EPub.php#L243) to have a clean relative path).
My bad, the change in wallabag was not really necessary...

@Grandt I think this problem is similar to https://github.com/Grandt/PHPePub/issues/24, so maybe we can use this issue instead of opening a new one?

PR #2306, opened 2016-09-24
Some checks about Redis & RabbitMQ

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | 
| License       | MIT

- Disable PHPUnit deprecation that fail tests since Twig 1.25.0 (it didn't seems the error come from something we've done)
- Display a message when async import won’t work
 Preview of the error message:
 
- Avoid failing tests for user who didn’t install Redis (test will be marked as skip)

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -16,6 +16,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_DIR" value="app/" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
     </php>
 
     <filter>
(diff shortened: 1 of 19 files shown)


tcitworld:
:+1:

phpunit.xml.dist
```diff
@@ -16,6 +16,7 @@
     <php>
         <ini name="error_reporting" value="-1" />
         <server name="KERNEL_DIR" value="app/" />
+        <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak" />
```
tcitworld: We'll have to watch updates about this.

PR #2315, opened 2016-09-28
Fixes about upgrade

nicosomb:
Fixes about `2.0.x` to `2.1.0` migration.

@Strubbl @cstuder @jlnostr I need help about german translations for this pull request please 💋


docs/de/index.rst
```diff
@@ -30,7 +30,8 @@ Die Dokumentation ist in anderen Sprachen verfügbar :
 
    user/faq
    user/installation
-   user/upgrade
+   user/upgrade-2.0.x-2.1.0
```
j0k3r: 2.1.x?

docs/en/user/upgrade-2.0.x-2.1.0.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Upgrade on a dedicated web server
+---------------------------------
+
+The last release is published on https://www.wallabag.org/pages/download-wallabag.html. In order to upgrade your wallabag installation and get the last version, run the following commands in you wallabag folder (replace ``2.1.0`` by the last release number):
+
+::
+
+    git fetch origin
+    git fetch --tags
```
j0k3r: `git fetch --all` should be enough, nope? (instead of `git fetch origin` + `git fetch --tags`)

docs/en/user/upgrade-2.0.x-2.1.0.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Download the last release of wallabag:
+
+.. code-block:: bash
+
+    wget http://wllbg.org/latest-v2-package && tar xvf latest-v2-package
+
+(md5 hash of the package: ``4f84c725d1d6e3345eae0a406115e5ff``)
+
+Extract the archive in your wallabag folder and replace ``app/config/parameters.yml`` with yours.
```
j0k3r: Be careful, because we added new stuff in this file: https://github.com/wallabag/wallabag/blob/2.1/app/config/parameters.yml.dist#L44-L52

docs/en/user/upgrade-2.0.x-2.1.0.rst
tcitworld: I remember to have read somewhere that `git fetch --all` doesn't necessary fetch tags. But it should be good. http://stackoverflow.com/a/1208223

docs/de/index.rst
nicosomb: we'll have `upgrade-2.1.0-2.0.x` file after?

docs/en/user/upgrade-2.0.x-2.1.0.rst
nicosomb: Yes that's why we did like that. So?

docs/de/index.rst
j0k3r: Wow no.
I was talking about `user/upgrade-2.0.x-2.1.x` in case some user wants to upgrade from 2.0.6 to 2.1.3, you get it?

docs/en/user/upgrade-2.0.x-2.1.0.rst
j0k3r: Ok let it like that.

docs/en/index.rst
```diff
@@ -56,3 +57,5 @@ The documentation is available in other languages:
    developer/documentation
    developer/translate
    developer/maintenance
+   developer/redis
```
j0k3r: What about `de`?

nicosomb: Because the files don't exist in german ;-)

j0k3r: Ok so we'll let translator to add it?

Strubbl: good hint. didn't see this so far

j0k3r: Thanks @Strubbl

j0k3r:
@nicosomb we should also add Instapaper, Chrome & Firefox in the doc (for async)

Issue #2116, opened 2016-05-26
Feature request: User management


ddeimeke:
I have some users using my wallabag v2 installation.

But after adding a user, I have not possibilities to see who I created.

Please add a user list and a possibility to delete users.

PR #2327, opened 2016-10-01
Add users management UI

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2116
| License       | MIT

Should fix https://github.com/wallabag/wallabag/issues/2116

- remove the “add a user” from the config page
- add a CRUD on user
- fix some missing translations (+ bad indentation)

Planned for 2.2 but if we got time to review for 2.1, it's fine too.

Listing:

Editing:

Create:


nicosomb:
I can remove my own account. Is that a problem? Then, I've have this error

> You cannot refresh a user from the EntityUserProvider that does not contain an identifier. The user object has to be serialized with its own identifier mapped by Doctrine.

j0k3r:
Hum didn't test that but might be related to the fact that you are logged-in with this account. Should be add a test so that the current logged-in user can't delete itself?

nicosomb:
If you can, I think it's a good option.

nicosomb:
👍

Issue #2305, opened 2016-09-24
"next" links in /api/entries responses lack search parameters


deanishe:
Wallabag 2.0.7

Search parameters aren't propagated to the URLs in `_links` in the JSON response. So given the URL
`/api/entries.json?perPage=30&page=1&archive=0`, the `_links.next.href` is `/api/entries.json?perPage=30&page=2` (i.e. the `archive` parameter has disappeared).

As a result, the URL doesn't point to the next page of results, but of my entire collection, and retrieving the `next` URLs will return all entries, not the search results.

j0k3r:
Good catch, thanks 👍

PR #2331, opened 2016-10-02
Fix parameters in API _links

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2305
| License       | MIT

Should fix https://github.com/wallabag/wallabag/issues/2305

We forgot to pass them to the factory.

### DIFF
--- a/src/Wallabag/CoreBundle/Repository/EntryRepository.php
+++ b/src/Wallabag/CoreBundle/Repository/EntryRepository.php
@@ -128,7 +128,7 @@ class EntryRepository extends EntityRepository
             $qb->andWhere('e.isStarred =:isStarred')->setParameter('isStarred', (bool) $isStarred);
         }
 
-        if ($since >= 0) {
+        if ($since > 0) {
             $qb->andWhere('e.updatedAt > :since')->setParameter('since', new \DateTime(date('Y-m-d H:i:s', $since)));
         }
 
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -128,7 +128,7 @@ public function findEntries($userId, $isArchived = null, $isStarred = null, $sor
             $qb->andWhere('e.isStarred =:isStarred')->setParameter('isStarred', (bool) $isStarred);
         }
 
-        if ($since >= 0) {
+        if ($since > 0) {
```
tcitworld: Why not ?

tcitworld:
:+1:

src/Wallabag/CoreBundle/Repository/EntryRepository.php
j0k3r: Because when `$since` is 0, `new \DateTime(date('Y-m-d H:i:s', $since))` with generate a date from January 1970. And on every request to the api, you will always add this condition to the query because 0 is the default value of `$since`.

PR #2349, opened 2016-10-03
Prepare wallabag 2.1.1

nicosomb:
Prepare wallabag 2.1.1 :dash:


nerzhul:
can you provide a github release containing the real application for production. I'm porting this to freebsd and it's just a pain to build app. I looked at the port tree and didn't see any good PHP app using composer for packages, they provide assets and already composed package.
Thanks

Issue #2311, opened 2016-09-27
Export in epub or pdf format in API - improvment


genma:
Actually, in the listed method's API (given by an url like http://localhost:8000/api/doc), there's not Export function. An improvment will be to add this feature in API .

PR #2372, opened 2016-10-05
API: ability to export entry in all available format (epub, pdf, etc...)

pmartin:
Following discussions on Twitter - see https://twitter.com/wallabagapp/status/783608582323462144 - here's kind of a *draft* PR.

There is an endpoint in the API to get an entry as a JSON string: `/api/entries/123.json`
The returned JSON string contains an HTML version of the entry's content, but this is not well suited for offline reading, especially as it doesn't contain the images of the entry.

Instead of this HTML version of the entry's content, I'd rather use an EPUB version: one file to download, embeds the images, and many reading engines exist (like on ereader devices ;-) ), as it's a well-known format.
Great thing is: wallabag already has a "download as EPUB" feature, which is currently not exposed by the API.

So: this PR is a draft (no test included for now -- seems to work on my server, but I haven't used this feature much) of an idea to **add EPUB export to the API**.

I re-used the existing endpoint to get one entry, but used the format information in the URL: instead of always returning a JSON string, this endpoint can now return EPUB content if called with the corresponding extension: `/api/entries/123.epub`

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | N/A
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2311
| License       | MIT

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -127,10 +127,17 @@ class WallabagRestController extends FOSRestController
      *
      * @return JsonResponse
      */
-    public function getEntryAction(Entry $entry)
+    public function getEntryAction(Entry $entry, $_format)
     {
         $this->validateAuthentication();
         $this->validateUserAccess($entry->getUser()->getId());
+
+        if ($_format === 'epub') {
+            return $this->get('wallabag_core.helper.entries_export')
+                ->setEntries($entry)
+                ->updateTitle('entry')
+                ->exportAs($_format);
+        }
 
         $json = $this->get('serializer')->serialize($entry, 'json');
 
--- a/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/WallabagRestController.php
@@ -3,7 +3,7 @@
 namespace Wallabag\ApiBundle\Controller;
 
 use FOS\RestBundle\Controller\FOSRestController;
-use Hateoas\Configuration\Route;
+use Hateoas\Configuration\Route as HateoasRoute;
 use Hateoas\Representation\Factory\PagerfantaFactory;
 use Nelmio\ApiDocBundle\Annotation\ApiDoc;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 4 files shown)


ThibaudDauce:
Thanks for the PR ☺️

As mentioned on Twitter, I would prefer the use the "Accept" HTTP header than a suffix URL that breaks the "one URL = one resource" REST rule. Bit it's not really important 😀

j0k3r:
Thanks for the PR!

As discussed with @nicosomb I would prefer an other entry point instead of re-using `getEntryAction`.

We could introduce a `getEntryExportAction` which will result in `/api/entries/export/123.epub`. And maybe we could easily add an option using HTTP header, as you said @ThibaudDauce 

Also, did you tick _Allow edits from maintainers_ ([see](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)) on that PR? Just to be sure that we can help you with that feature in that same PR :)

pmartin:
Another endpoint would be perfectly fine :-)
(To be honest, I just went the quickest way to set this up in as few lines as possible)

Yes, I checked "allow edit from maintainers" (and I just checked it's checked ^^ )

j0k3r:
I've updated the PR (I love being able to push on a fork).
It should now handle all export.

@pmartin could you give an other try?
If everything is ok for you, this will be merged an available in 2.2.0

src/Wallabag/ApiBundle/Controller/WallabagRestController.php
```diff
@@ -138,6 +139,30 @@ public function getEntryAction(Entry $entry)
...
     /**
+     * Retrieve a single entry as a predefined format.
+     *
+     * @ApiDoc(
+     *      requirements={
+     *          {"name"="entry", "dataType"="integer", "requirement"="\w+", "description"="The entry ID"}
+     *      }
+     * )
+     *
+     * @Route(requirements={"_format"="epub|mobi|pdf|txt|csv"})
```
tcitworld: It lacks xml export, but this should be accessible directly through the regular api calls ?

j0k3r: That's what I said in the commit message 🐸 

> Export isn't available for json & xml because user can use the default
entry endpoint instead.

tcitworld: :roll_eyes:

pmartin:
@j0k3r I've deployed the changes in this PR to my wallabag (still 2.0.x) instance ; and export to epub does still work \o/
I can download EPUB version of entries from my browser and from my ereader app \o/
Thanks a lot :-)

j0k3r:
@pmartin great! So we can merge that, thanks !

PR #2406, opened 2016-10-09
Show number of annotations instead of nbAnnotations placeholder

szafranek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | n/a
| Documentation | no
| Translation   | yes
| Fixed tickets | n/a
| License       | MIT

When there was more than one annotation, the `%nbAnnotations%` placeholder was displayed in both themes, instead of actual number.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
@@ -54,7 +54,6 @@
                 {% endif %}
             </i>
 
-            {% set nbAnnotations = entry.annotations | length %}
             <span class="tool link"><i class="material-icons link">comment</i> {{ 'entry.view.annotations_on_the_entry'|transchoice(entry.annotations | length) }}</span>
             <aside class="tags">
                 <div class="card-entry-tags">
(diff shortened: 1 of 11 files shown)


src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
```diff
@@ -54,7 +54,6 @@
                 {% endif %}
             </i>
 
-            {% set nbAnnotations = entry.annotations | length %}
```
j0k3r: Could you also remove that line in the material theme?

szafranek: That line was never added to the Material theme in a first place.

j0k3r: Oups yeah sorry :)

tcitworld: I guess that's the origin of the issue.

szafranek: I tested both themes and this line was not helping in the baggy theme either – both themes showed a placeholder instead of the number of annotations.

tcitworld: Oh. Well, good catch then.

PR #2411, opened 2016-10-09
Fix entities definition

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

As per Doctrine said in the debug tool bar:

- The field Wallabag\ApiBundle\Entity\Client#refreshTokens is on the inverse side of a bi-directional relationship, but the specified mappedBy association on the target-entity Wallabag\ApiBundle\Entity\RefreshToken#client does not contain the required 'inversedBy="refreshTokens"' attribute.

- The field Wallabag\ApiBundle\Entity\Client#accessTokens is on the inverse side of a bi-directional relationship, but the specified mappedBy association on the target-entity Wallabag\ApiBundle\Entity\AccessToken#client does not contain the required 'inversedBy="accessTokens"' attribute.

### DIFF
--- a/src/Wallabag/ApiBundle/Entity/AccessToken.php
+++ b/src/Wallabag/ApiBundle/Entity/AccessToken.php
@@ -19,7 +19,7 @@ class AccessToken extends BaseAccessToken
     protected $id;
 
     /**
-     * @ORM\ManyToOne(targetEntity="Client")
+     * @ORM\ManyToOne(targetEntity="Client", inversedBy="accessTokens")
      * @ORM\JoinColumn(nullable=false)
      */
     protected $client;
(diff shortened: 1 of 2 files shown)


tcitworld:
Doesn't this implies migrations ?

j0k3r:
I'll check that but since it's just for Doctrine internally I don't think so

nicosomb:
For me, this change doesn't need a migration. It's "only" the name of the PHP variables, right?

j0k3r:
No migration required. Doctrine generated an empty file.

Issue #2210, opened 2016-08-07
Firefox bookmarklet not working 


leveled:
### Issue details

The wallabag addon for Firefox does not seem to be working for me at all, there is no pop-up or anything when I push the hotkey.  The chrome bookmarklet has no problems saving pages. 

### Environment

* wallabag version: latest version 
* How did you install wallabag: this isn't a problem with wallabag, but the firefox extension

### Steps to reproduce/test case

I'm using OS X 10.11.5 and the latest version of Firefox.

j0k3r:
Could you tell us the version of the bookmarklet?

leveled:
It's the v2-0.0.8

Issue #2130, opened 2016-05-31
On Chrome, the bookmarlet doesn't work


NerosTie:
### Issue details

On Chrome, when I click on the bookmark, nothing happen. It was not an issue with the bookmarlet of wallabag 1.x.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.4
* How did you install wallabag? Via `git clone` or by downloading the package?
* Last wallabag version that did not exhibit the issue (if applicable):
* php version:
* OS:
* type of hosting (shared or dedicated):
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):

### Steps to reproduce/test case

1. Use Chrome
2. Add the bookmarklet to your bookmarks 
3. Go to an article to save
4. Click on the bookmark
5. Nothing happen

j0k3r:
Are you using that bookmarklet (https://github.com/Rurik19/wallabagger) from @Rurik19 ?

Rurik19:
I think it about that part of [documentation.](http://doc.wallabag.org/en/master/user/first_article.html)

> By using a bookmarklet
On the Howto page, you have a Bookmarklet tab. Drag and drop the bag it! link to your bookmarks bar of your browser.
Now, each time you're reading an article on the web and you want to save it, click on the bag it! link in your bookmarks bar. The article is saved.

and indeed, in chrome that doesn't work

tcitworld:
May be related to https://github.com/wallabag/wallabag/issues/1960.

nickian:
Here's my solution to this problem. Instead of opening a new window and closing it, just redirect to add article URL:

`javascript:var url=location.href||url;var wllbg=window.location.replace('https://url.com/web/bookmarklet?url=' + encodeURI(url));void(0);`

NerosTie:
@nickian Your solution works well!

softmetz:
I had the same problem in Firefox 47 and @nickian's workaround solved it!

Issue #1960, opened 2016-04-21
Bookmarklet import does not work - page closes too fast


ToeiRei:
### Issue details

Using the bookmarklet 'bag it!' you cannot import articles as it closes too fast

### Environment

* wallabag version: 2.0.2
* How did you install wallabag? Via `git clone`
* Last wallabag version that did not exhibit the issue (if applicable):
* php version: 7.0
* OS: Linux
* type of hosting (shared or dedicated): shared on own server
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MySQL/MariaDB
* Browser: Chrome 49

### Steps to reproduce/test case

1. Use the bookmarklet
2. Wonder why the article isn't in 'unread'
3. Look at the JS code used in the Bookmarklet
4. Remove `wllbg.close();` to keep the window open to maybe get an error message
5. Be astonished that it works and come to the conclusion that `wllbg.close();` kicks in too fast for actually getting the request through the server

ToeiRei:
could be related to #238 in some way.

tcitworld:
I'm not sure what the expected behavior should be : tab staying opened a few seconds ?

Things that we can do are quite limited with the bookmarklet, I'm not even convinced that we should support it in the future.

ToeiRei:
If it stays open, the client got enough time to establish a connection and do the request. I'd suppose that the tab staying open showing the article fetched sounds like a proper solution as you are instantly able to review the spidered page.

tcitworld:
I'll ask for other people's opinion, and make the change if needed.

nicosomb:
Can reproduce on Chrome 50. 
Can't reproduce on Firefox 45.

ddeimeke:
Reproduced on Firefox 45.0.2 (Linux) and Firefox 38.0.x (Windows).

PeterTGarner:
I've been running more tests and it looks like the "wllbg.close;" in the bookmarklet _is_ the problem. If I remove that I have no problem. The web server running Wallabag is quite slow.  This would be a much-appreciated fix, or replace with something else.

hugligan:
PeterTGarner, thank you!! That fixed my issueses as well. Ive been using wallabag for a good year or two and this is the single most important feature if you ask me!

Perksls:
i was also having problems with the bookmarklet 
i've set a timeout on close, and everything seems good now, mine:

[...]

PR #2414, opened 2016-10-09
Remove automatic closing of the window from bookmarklet

szafranek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | n/a
| Documentation | no
| Translation   | no
| Fixed tickets | #2130, #2210, #1960
| License       | MIT

Current bookmarklet is broken, because it opens new Wallabag window, and closes it immediately, before Wallabag has a chance to add the link to the database.

Ideal solution would be to have the bookmarklet act _within_ the article page (like Pocket does), but that's significant amount of work. This PR prevents new window from being automatically closed. I realize it's not ideal, but at least it makes the bookmarklet work.

Additionally, I surrounded the JavaScript code with anonymous function, to avoid polluting page's scope by bookmarklet's variables.


j0k3r:
Thanks ! 👍

PR #2417, opened 2016-10-10
Define a dev version for the master

j0k3r:
This is to be sure that user testing wallabag from the master will have the version defined as dev.

What do you think?


tcitworld:
:+1: 
Also, write this in Release process in order not to forget it.

PR #2432, opened 2016-10-12
Update ISSUE_TEMPLATE.md

j0k3r:
The idea is to guide user to, at least, read the doc about a fetching issue.
Maybe they will try on their side before asking for help ..


nicosomb:
🙏

Issue #2392, opened 2016-10-07
Document parameters.yml and parameters setup during initial install or upgrade


zertrin:
### Issue details

During initial install or upgrade to a new version which introduces new parameters in `parameters.yml`, the step `composer install` will ask the user for parameters. For example:

```
Updating the "app/config/parameters.yml" file
Some parameters are missing. Please provide them.
fosuser_registration (true):
[...]
```

Without any context, and without finding documentation about it ([I searched for `parameters.yml`](http://doc.wallabag.org/en/master/search.html?q=parameters.yml&check_keywords=yes&area=default) and didn't find what I was looking for, [searching for `fosuser_registration`](http://doc.wallabag.org/en/master/search.html?q=fosuser_registration#) gives zero results...), I feel a bit lost and am not sure what to enter.

I propose to add some documentation about `parameters.yml` and make sure this documentation is easily reachable from the upgrade and install documentation pages.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.1.1
* How did you install wallabag? Via `git clone` or by downloading the package? git clone
* OS: Debian Jessie
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): sqlite

j0k3r:
Yep we lack documentation on that part.

When `composer` ask for a new value for a parameter it is usually safe to use the default value (the one in parenthesis).

But I really agree, we need documentation about `parameters.yml`.

j0k3r:
http://doc.wallabag.org/en/master/user/parameters.html

nicosomb:
We need to add this page in the menu.

PR #2434, opened 2016-10-12
Added french documentation for parameters.yml

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  |  no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| Fixed tickets | 
| License       | MIT

In #2392, we forgot french documentation and links for the menu. 

@Strubbl @jlnostr Can you help us for the german translation please? :)

### DIFF
--- a/docs/en/user/parameters.rst
+++ b/docs/en/user/parameters.rst
@@ -9,9 +9,9 @@ What is the meaning of the parameters?
    "database_name", "symfony", "name of your database"
    "database_user", "root", "user that can write to this database"
    "database_password", "~", "password of that user"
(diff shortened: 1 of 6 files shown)


docs/en/user/parameters.rst
```diff
@@ -9,9 +9,9 @@ What is the meaning of the parameters?
    "database_name", "symfony", "name of your database"
```
Strubbl: is the default really symfony and not wallabag?

nicosomb: the default database name is `symfony` (see https://github.com/wallabag/wallabag/blob/master/app/config/parameters.yml.dist). You're right, we can change this default value.

Issue #2129, opened 2016-05-31
UI issue with the panel on the left on small screen (baggy)


NerosTie:
### Issue details

The panel on the left is an issue on small screens with the theme baggy as you can see on the screenshot above:

![screenshot_20160531_053559](https://cloud.githubusercontent.com/assets/4572452/15662518/c2c96c70-26f1-11e6-9440-a761438a1b69.png)

`Déconnexion` enters in conflict with other entries and there is no possibility to access to other entries (like help, developer, import...) because I can't scroll.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.4
* How did you install wallabag? Via `git clone` or by downloading the package?
* Last wallabag version that did not exhibit the issue (if applicable):
* php version:
* OS:
* type of hosting (shared or dedicated):
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):

### Steps to reproduce/test case

1. reduce the windows of the web browser

tcitworld:
Should be better now in next release.

PR #2454, opened 2016-10-15
Fix baggy display on small screens

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2129
| License       | MIT

Also, remove source maps from assets since the sources files are in build/ and not accessible for the server


j0k3r:
Any before/after?

j0k3r:
ok we don't care :)

tcitworld:
The users, internal and developer menu items are removed when the height is too small. Need more ? :p

PR #2456, opened 2016-10-15
use new tcpdf library version

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

Maybe wait a bit for first release

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "jms/serializer-bundle": "~1.1",
         "nelmio/api-doc-bundle": "~2.7",
         "mgargano/simplehtmldom": "~1.5",
-        "tecnickcom/tcpdf": "~6.2",
+        "tecnickcom/tc-lib-pdf": "dev-master",
         "simplepie/simplepie": "~1.3.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -82,7 +81,8 @@
         "white-october/pagerfanta-bundle": "^1.0",
         "php-amqplib/rabbitmq-bundle": "^1.8",
         "predis/predis": "^1.0",
-        "javibravo/simpleue": "^1.0"
+        "javibravo/simpleue": "^1.0",
+        "tecnickcom/tc-lib-pdf": "dev-master"
```
j0k3r: Wow, might be a good idea to keep around before merging it right now. Why not targeting 2.2 branch? So we'll be able to give more tests on it

tcitworld: Hope it won't get dirty when we merge master into 2.2

j0k3r:
Can you rebase it against the 2.2? So we can get it mergerd.

PR #2483, opened 2016-10-21
Mark as mobile app capable

nicofrand:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

This will allow wallabag web-app to be added to the homescreen on smartphones, without the URL bar etc.. so that is acts more like a real application.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/base.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/base.html.twig
@@ -11,6 +11,8 @@
             <!--[if IE]>
             <meta http-equiv="X-UA-Compatible" content="IE=10">
             <![endif]-->
+            
+            <meta name="mobile-web-app-capable" content="yes">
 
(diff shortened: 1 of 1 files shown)


tcitworld:
Better to add a manifest.json file : https://developer.chrome.com/multidevice/android/installtohomescreen#supporting

nicofrand:
I am fine with both even if mobile-app-capable is the shortest solution when not providing different icons etc.

Why would you close the PR though ? Start a review or something so that I can modify it. That really doesn't encourage to contribute anymore.

tcitworld:
Sorry, my comment was a bit harsh, you're welcomed to open another pull request or an issue directly. I just closed because this change is of no use if we have a manifest.json file.

nicofrand:
I don't see any manifest.json anywhere in the repository ?

tcitworld:
I was planning to add it (or to review a PR ;-) )

PR #2499, opened 2016-10-24
Added relation between API Client and User

### LINKED ISSUE #2062: All user can delete clients
### Delete Clients

I've create a client for my Firefox AddOn and now i create a new user and see that this user can see this client and a button to delete the client. 

At the moment i can't delete clients from any account by push the button. I get the following error: 

> Oops! An Error Occurred
The server returned a "500 Internal Server Error".
Something is broken. Please let us know what you were doing when this error occurred. We will fix it as soon as possible. Sorry for any inconvenience caused. 

But i think that only the user who create a client and the admin should be the ability to delete the client. All the other User can only see the client but not the delete button!

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.4
* Last wallabag version that did not exhibit the issue (if applicable): i don't know
* php version: 7
* OS: archlinux

1. Create a client
2. Create a new user
3. Login at this new User and look to the developer section. This new user has the ability to delete the client you have create.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2062
| License       | MIT

Fix #2062 

Todo

- [x] Test (for the delete client action)
- [x] Need to try the database migration

### DIFF
--- a/app/DoctrineMigrations/Version20161024212538.php
+++ b/app/DoctrineMigrations/Version20161024212538.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20161024212538 extends AbstractMigration implements ContainerAwareInterface
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    private function getTable($tableName)
+    {
+        return $this->container->getParameter('database_table_prefix') . $tableName;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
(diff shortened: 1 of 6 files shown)


app/DoctrineMigrations/Version20161024212538.php
```diff
@@ -34,6 +34,7 @@ public function up(Schema $schema)
     {
         $this->skipIf($this->connection->getDatabasePlatform()->getName() == 'sqlite', 'Migration can only be executed safely on \'mysql\' or \'postgresql\'.');
 
+        $this->addSql('ALTER TABLE '.$this->getTable('oauth2_clients').' ADD `user_id` INT(11) DEFAULT NULL');
```
j0k3r: :no_good_man: PostgreSQL doesn't like `

nicosomb: Removed.

app/DoctrineMigrations/Version20161024212538.php
```diff
@@ -0,0 +1,48 @@
...
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
```
j0k3r: You can remove that useless comment.

PR #2506, opened 2016-10-27
Fixed entries export filtered with a tag

### LINKED ISSUE #2505: 500 internal server error when exporting entries filtered by tag
### Issue details

From the list of articles filtered by tag, I get a 500 internal server error when I try to download the filtered articles. 

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.1.2
* How did you install wallabag? Via git clone
* Last wallabag version that did not exhibit the issue (if applicable): unknown
* php version: 7.0.12
* OS: Arch linux
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): SQLite

I just installed wallabag, added few entries and added tags on those entries.
Then I choose a tag and list the articles filtered by this tag and click on "Export" and then on any format.
The exact error that appears is :
Undefined method 'getBuilderForTag_entriesByUser'. The method name must start with either findBy or findOneBy!

### Other indications

From what I could see, no method 'getBuilderForTag_entriesByUser' exists in repository WallabagCoreBundle:Entry but method downloadEntriesAction of CoreBundle/Controller/ExportController.php tries to call it when I try to fetch url /export/tag_entries.pdf.
When the same route is called with the other categories (all, unread, starred, archive, untagged) there is no problem because the corresponding methods exist in the repository.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #2505 
| License       | MIT

Fix #2505

Quick and very dirty. If it's too dirty, please close this PR 😃

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/ExportController.php
+++ b/src/Wallabag/CoreBundle/Controller/ExportController.php
@@ -4,8 +4,10 @@ namespace Wallabag\CoreBundle\Controller;
 
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Wallabag\CoreBundle\Entity\Entry;
+use Wallabag\CoreBundle\Entity\Tag;
 
 /**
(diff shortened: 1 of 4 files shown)


j0k3r:
Yeah it's kind of ugly but I don't see other better solution atm.

PR #2500, opened 2016-10-24
Added a check in Makefile to see if composer is installed

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes, sort of
| Translation   | no
| Fixed tickets | #2498 
| License       | MIT

Fix #2498

PR #2507, opened 2016-10-27
Added require.sh to check if composer is installed

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | 
| License       | MIT

As discussed in #2500, we need to check if composer is installed locally (with a `composer.phar` file). 
I added a `require.sh` file to check it.

### DIFF
--- a/scripts/require.sh
+++ b/scripts/require.sh
@@ -3,6 +3,7 @@
 
 if [ ! -f composer.phar ]; then
     echo "composer.phar not found, we'll see if composer is installed globally."
+    command -v composer >/dev/null 2>&1 || { echo >&2 "wallabag requires composer but it's not installed (see http://doc.wallabag.org/en/master/user/installation.html). Aborting."; exit 1; }
+else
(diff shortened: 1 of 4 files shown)


scripts/require.sh
```diff
@@ -0,0 +1,8 @@
+#! /usr/bin/env bash
+# File used to check dependencies
+
+if [ ! -f composer.phar ]; then
+    echo "composer.phar not found, we'll see if composer is installed globally."
```
j0k3r: if `composer.phar` exist the next command will still fail.

nicosomb: Oops, you're right 😁

nicosomb: and if `composer.phar` is found, we need to use it in the command, instead of `composer`.

nicosomb:
I think it'sOK: variable seems to be global in bash.

PR #2520, opened 2016-10-30
Inject parameter instead of service

j0k3r:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Documentation | no |
| Translation | no |
| Fixed tickets |  |
| License | MIT |

We are injecting CraueConfig service when we only need to retrieve one or two values from it.
Instead I discovered we can directly inject a value from a service in the service definition!


nicosomb:
Oh nice 👍

Issue #2531, opened 2016-11-02
Imported items comes in reverse sorting 


gerroon:
### Issue details
I exported my database as json from wallabag v2 then imported it into the latest version.  

My articles were from latest to the earliest. The articles shows up as first to last in the latest version. So first page shows the first articles so on.

### Environment

* wallabag version (or git revision) that exhibits the issue:
501a4dae5b5a89e76fd4c7b00170a5f42beb1b63

* How did you install wallabag? Via `git clone` or by downloading the package?
git clone, make script (like it was mentioned in the install doc)

* Last wallabag version that did not exhibit the issue (if applicable):
I am not sure. I believe that from v1 to v2 was ok/

* php version:
PHP 5.6.27-1~dotdeb+7.1 (cli) (built: Oct 15 2016 02:45:48) 

* OS:
Debian Wheezy

* type of hosting (shared or dedicated):
Dedicated 

* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):
SQLite

### Steps to reproduce/test case
Export your database from v2.0 as json then import into the latest version

Maybe there is some kind of internal setting for sorting the articles but I was not able to find it. IOt would be nice if someone can also provide a way to resort my articles so they show up properlly.

j0k3r:
This is intriguing. I guess list are sorted by `updated_at` date.
We only override the `created_at` date when importing.

gerroon:
@j0k3r 

Is there a way to force to reverse the list? It is a bit of an issue with the android app because it only downloads the earliest ones naturally. I have hundreds of articles.

thanks

gerroon:
@j0k3r 

thanks

PR #2535, opened 2016-11-04
Use created_at as default sort

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2531 
| License       | MIT

Imported content use the real creation date. If we sort by id, it'll
display contents in reverse order.

Following https://github.com/wallabag/wallabag/pull/2534 but this time I added an index to avoid database to use a `filesort` instead of the index

### DIFF
--- a/app/DoctrineMigrations/Version20161104073720.php
+++ b/app/DoctrineMigrations/Version20161104073720.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Application\Migrations;
+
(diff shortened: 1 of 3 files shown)


nicosomb:
👍

PR #2547, opened 2016-11-06
Added a configuration to define the redirection after archiving an entry

### LINKED ISSUE #496: return to list after marking something as read
i would prefer an automatic return to the item-list/start after marking an arcticle as read instead of staying on the article-page - usually i'm done with this article when i click it as read ...

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   |  no
| Documentation | yes
| Translation   | yes
| Fixed tickets | #496 
| License       | MIT

Added a new option in configuration: 

Fix #496

@j0k3r @tcitworld I need help for the tests. 
@Strubbl @jlnostr I need help to translate this PR.

### DIFF
--- a/app/DoctrineMigrations/Version20161106113822.php
+++ b/app/DoctrineMigrations/Version20161106113822.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class Version20161106113822 extends AbstractMigration
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    private function getTable($tableName)
+    {
+        return $this->container->getParameter('database_table_prefix') . $tableName;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE "'.$this->getTable('config').'" ADD action_mark_as_read INT DEFAULT 0');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'sqlite', 'This down migration can\'t be executed on SQLite databases, because SQLite don\'t support DROP COLUMN.');
+
+        $this->addSql('ALTER TABLE "'.$this->getTable('config').'" DROP action_mark_as_read');
+    }
+}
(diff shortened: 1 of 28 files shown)


nicosomb:
@Strubbl thank you! Can you have a look to messages.de.yml please?

app/DoctrineMigrations/Version20161106113822.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'sqlite', 'This down migration can\'t be executed on SQLite databases, because SQLite don\'t support DROP COLUMN.');
+
+        $this->addSql('ALTER TABLE "'.$this->getTable('config').'" DROP action_mark_as_read');
```
j0k3r: Dropping a column doesn't work on SQLite?

src/Wallabag/CoreBundle/Helper/Redirect.php
```diff
@@ -24,6 +27,10 @@ public function __construct(Router $router)
      */
     public function to($url, $fallback = '')
     {
+        if ($this->actionMarkAsRead == 0) {
```
j0k3r: Maybe we should define some constant in Config class to avoid using this not relevant `0`.
Also, use `===` instead of simple `==` and use yoda condition (put the defined value first).

j0k3r:
For testing, you can shift the config in database before the request, and shift back to the original value after.

app/DoctrineMigrations/Version20161106113822.php
nicosomb: http://stackoverflow.com/questions/8442147/how-to-delete-or-add-column-in-sqlite

tests/Wallabag/CoreBundle/Helper/RedirectTest.php
```diff
@@ -15,21 +15,24 @@ class RedirectTest extends \PHPUnit_Framework_TestCase
...
+            ->getMock();
+        $this->redirect = new Redirect($this->routerMock, $tokenStorage);
     }
 
     public function testRedirectToNullWithFallback()
     {
         $redirectUrl = $this->redirect->to(null, 'fallback');
 
-        $this->assertEquals('fallback', $redirectUrl);
+        $this->assertEquals(null, $redirectUrl);
```
j0k3r: You are wrong when updating these tests. Because here, you are just testing that when a user is null, it returns the given url: `null`.
You are not testing the fallback

nicosomb:
@j0k3r I'm lost with mocks. Could you help me please?

j0k3r:
Rebased and test updated 💚

PR #2601, opened 2016-11-18
Added http_status in Entry entity

### LINKED ISSUE #2589: Store the HTTP status code of the articles
graby already sends the HTTP status code for each article (thank you @j0k3r). 
We need to store this status in the entry table. 
And we need to have a page to list all the 404 articles. 

It will be useful for massive imports (from third services): maybe some articles are not available and it will be more convenient to retrieve these articles.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| Fixed tickets | #2589
| License       | MIT

Fix #2589

We now store the HTTP status code. 
We can filter entries thanks to the new filter, see screenshot (I added the filter in baggy theme too).
In an other PR, it's possible to improve the field (by replacing the text field with a list for example). 

capture d ecran 2016-11-18 a 15 08 13

### DIFF
--- a/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
+++ b/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
@@ -11,6 +11,7 @@ use Lexik\Bundle\FormFilterBundle\Filter\Form\Type\CheckboxFilterType;
 use Lexik\Bundle\FormFilterBundle\Filter\Form\Type\ChoiceFilterType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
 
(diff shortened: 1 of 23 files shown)


tcitworld:
- [ ] Restrict to existing http codes (and add a specific test)

nicosomb:
You mean ... ALL THESE CODES https://en.wikipedia.org/wiki/List_of_HTTP_status_codes ?

tcitworld:
Exactly. :-)

But maybe you could use `\Symfony\Component\HttpFoundation\Response` and check if the code is in the keys of `Response::$statusTexts`.

tcitworld:
Or you can just check if it's RFC-invalid with `$this->statusCode < 100 || $this->statusCode >= 600;`, whatever.

src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
```diff
@@ -91,6 +92,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'label' => 'entry.filters.domain_label',
             ])
             ->add('httpStatus', TextFilterType::class, [
+                'apply_filter' => function (QueryInterface $filterQuery, $field, $values) {
+                    $value = $values['value'];
+                    if (false === array_key_exists($value, Response::$statusTexts)) {
+                        return;
+                    }
+
+                    $paramName = sprintf('%s', str_replace('.', '_', $field));
```
j0k3r: Could you please explain these lines?

nicosomb: As explained in the bundle doc, we need to replace the dot https://github.com/lexik/LexikFormFilterBundle/blob/master/Resources/doc/working-with-the-bundle.md#ii-filter-customization

PR #2605, opened 2016-11-19
Add .travis.yml change to RELEASE_PROCESS

j0k3r:
This change was missing from the release process.


nicosomb:
Oh yes, I use wallabag/releaser and RELEASE_PROCESS.md, so I don't forgot it. 
But thank you 👍

PR #2613, opened 2016-11-20
Fix card action overlap

### LINKED ISSUE #2394: Misalignment of the icons on the bottom right of the entry cards on some resolutions
### Issue details

The icons next to the entry's domain name may be dislocated if the domain name is a bit too long.

See screenshots:

When 3 entries are shown per rows:
nl486uhl27zun17l6

With only 2 entries per row (smaller screen):
1gr6esg46gv6dhz6f
### Environment
- wallabag version (or git revision) that exhibits the issue: **2.1.1**
- How did you install wallabag? Via `git clone` or by downloading the package? **git clone**
- Last wallabag version that did not exhibit the issue (if applicable): **2.0.8**
- php version: **PHP 5.6.24-0+deb8u1**
- OS: **Debian Jessie**
- type of hosting (shared or dedicated): **dedicated**
- which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): **sqlite**
- Browser: **Firefox Developer Edition 51.0a2 (2016-10-03)**

Issue happens for me after bagging for example the following articles (see screenshots above):
- https://www.technologyreview.com/s/602541/physicists-create-worlds-first-time-crystal/
- https://www.quantamagazine.org/20150702-paradoxical-crystal-baffles-physicists/
- http://sebastianraschka.com/blog/2015/why-python.html

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | fix #2394
| License       | MIT

I've reduced the margin on the right for the card-action button, so they might not be go under the card on smaller resolution.

Before:
image

After:
image


app/Resources/static/themes/material/css/main.css
```diff
@@ -544,6 +544,10 @@ a.original {
   line-height: 24px;
 }
 
+.card .card-action ul.tools li a.tool {
+  margin-right: 5px !important;
```
tcitworld: Sure `!important` is needed here ?

tcitworld: Will have a quick look, it shouldn't be like this. :-)

tcitworld:
Does the trick. Good for now.

PR #2614, opened 2016-11-20
Add ability to reload entry from API

### LINKED ISSUE #2489: Reload content via API
When a site config or site contents have changes, it would be handy to be able to programmatically refetch contents.

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes (api doc)
| Translation   | no
| Fixed tickets | fix #2489
| License       | MIT

Entry can be reloaded using the API.
Note that if we can't reload the entry (because we can't retrieve content OR the retrieved content failed), an empty 304 will be returned.
Otherwise, the full new entry will be return.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -287,7 +287,7 @@ class EntryRestController extends WallabagRestController
 
     /**
      * Reload an entry.
-     * An empty response with HTTP Status 304 will be send if we weren't able to update the content (because it hasn't changed or we got an error)
+     * An empty response with HTTP Status 304 will be send if we weren't able to update the content (because it hasn't changed or we got an error).
      *
      * @ApiDoc(
      *      requirements={
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -286,6 +286,54 @@ class EntryRestController extends WallabagRestController
     }
 
     /**
+     * Reload an entry.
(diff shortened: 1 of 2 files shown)


tcitworld:
Maybe we could return an error message with the 304 to know if refreshing failed because of the fetching call or because of wallabag itself ?

j0k3r:
304 or 204 shouldn't return content.
Which status should we use instead if we want to return some error data?
- 412 HTTP_PRECONDITION_FAILED
- 417 HTTP_EXPECTATION_FAILED

tcitworld:
The most meaningful I've found should be 502 Bad Gateway when we've got an fetch issue or 521 Web Server Is Down (cloudflare, not official) and a generic 500 Internal Server Error if wallabag fails.

However, I guess sysadmins wouldn't like this kind of issues in the logs, therefore we can have a generic 400 code and put an extra meaninful code in the content.

nicosomb:
@j0k3r is this PR still RFR?

j0k3r:
Yup.

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -286,6 +286,51 @@ public function patchEntriesAction(Entry $entry, Request $request)
     }
 
     /**
+     * Reload an entry.
+     * A response with HTTP Status 400 will be return if we weren't able to update the content (because it hasn't changed or we got an error).
```
nicosomb: > 400 Bad Request
> The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, too large size, invalid request message framing, or deceptive request routing).

> because it hasn't changed

It's not an error, so I don't know if status 400 is the good one.

PR #2616, opened 2016-11-20
Change documentation links to HTTPS

mathieui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes (install/upgrade doc)
| Translation   | no
| Fixed tickets | none
| License       | MIT

Even though all the websites used in the commands redirect http to
https, a man in the middle can happen with the initial request,
leading to arbitrary code execution.

### DIFF
--- a/docs/de/user/installation.rst
+++ b/docs/de/user/installation.rst
@@ -16,7 +16,7 @@ Composer installieren:
 
 ::
 
-    curl -s http://getcomposer.org/installer | php
+    curl -s https://getcomposer.org/installer | php
 
 Du kannst eine spezifische Anleitung `hier <https://getcomposer.org/doc/00-intro.md>`__ finden.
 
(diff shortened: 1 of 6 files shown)


docs/de/user/installation.rst
```diff
@@ -86,7 +86,7 @@ Führe dieses Kommando aus, um das neueste Paket herunterzuladen und zu entpacke
 
 .. code-block:: bash
 
-    wget http://wllbg.org/latest-v2-package && tar xvf latest-v2-package
+    wget https://wllbg.org/latest-v2-package && tar xvf latest-v2-package
```
j0k3r: This one doesn't work.

mathieui: Thanks, indeed. Could that be fixed by the wllbg.org owner? (the redirection works, but the certificate is invalid for that domain)

I could hardlink the framabag.org destination, but that would defeat the purpose of having the "latest" redirection.

j0k3r: Poke @nicosomb

tcitworld: @mathieui They are now at https://static.wallabag.org/releases/

mathieui: @tcitworld thanks, is there a way to have a static link to the (ever-changing) latest version?

mathieui: @nicosomb the certificate doesn’t match the hostname, which will (hopefully) fail by default in all browsers and tools (this is what I put in the PR already)

nicosomb:
Can you change the destination branch to 2.2 please?

j0k3r:
@nicosomb you can do it too 😉

Issue #2643, opened 2016-11-24
Add reading progress to 


jlnostr:

nicosomb:
Today, we don't store the reading progress for each article. We use HTML5 storage. 

Questions for @wallabag/core: 

How can we implement that?
When can we save this value?

jlnostr:
I've implemented it in my app with this JavaScript snippet:

```javascript

function updateReadingProgress() {
    var root = document.getElementsByTagName("html")[0];
    var top = document.body.scrollTop;
    var scrollPercentage = 100 * top / (root.scrollHeight - root.clientHeight);

    if (root.scrollHeight - top === root.clientHeight || scrollPercentage > 100)
        window.external.notify("finishedReading");
    window.external.notify(scrollPercentage.toString());
}
```

This would return a percentage value of the progress, which then could be saved in the database. In my case it's always saved after the user navigated back.

j0k3r:
We can take inspiration from Instapaper API (`/api/1/bookmarks/update_read_progress`): https://www.instapaper.com/api

tcitworld:
Oh, wait, I've done this already. Will eventually change according to Instapaper API.

nicosomb:
@j0k3r OK for the API. 

My question was for the web application.

tcitworld:
> How can we implement that?
When can we save this value (each N seconds, etc.) ?

Same as for annotations, a dedicated route. We save the value when the scroll changes (like the localstorage thing did I guess)

actuallynotjohndoe:
Do you plan to implement this feature in the near future? The past week I've been using wallabag quite intensive and to be able to sync the reading progress between devices would be a killer feature! I guess I'm not the only one that would really like this feature.

krillin666:
Any updates ?

krillin666:
Almost a year later, is this being worked on ? Thank you !

j0k3r:
No progress on this.

ovizii:
Ah, so there's no progress saving or syncing. 
After reading this on wallabag's main page I had high hopes:

![image](https://github.com/wallabag/wallabag/assets/837902/b538fcb6-4936-4721-ab4a-36d577be527d)

RrOoSsSsOo:
I would like to start by saying that from my point of view this is a fundamental feature to have in Wallabag and should be implemented as soon as possible.
[...]

PR #2644, opened 2016-11-24
Bring article progress sync through API

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #2643
| License       | MIT


src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -680,6 +689,7 @@ public function cleanUuid()
     }
 
     /**
+<<<<<<< Updated upstream
```
tcitworld: Boo

nicosomb:
We also need to add the code in the view of the web application (to send the reading progress, maybe with javascript like @jlnostr wrote)

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -272,6 +274,13 @@ public function patchEntriesAction(Entry $entry, Request $request)
             $entry->setStarred((bool) $isStarred);
         }
 
+        if (!is_null($progress)) {
```
nicosomb: @j0k3r is that better to use `is_null` or `null ===` ?

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -272,6 +274,13 @@ public function patchEntriesAction(Entry $entry, Request $request)
...
+            $progress = (int) $progress;
+            if ($progress >= 0 && $progress <= 100) {
+                $entry->setProgress((int) $progress);
```
nicosomb: $entry->setProgress($progress);

src/Wallabag/ApiBundle/Controller/EntryRestController.php
tcitworld: http://stackoverflow.com/a/8229005

j0k3r: It seems that we don't often use `is_null`, so if we want to keep consistency, we should use `null ===`

tcitworld:
Rebased.

tcitworld:
Rebased.

actuallynotjohndoe:
Do you plan to implement this feature in the near future? The past week I've been using wallabag quite intensive and to be able to sync the reading progress between devices would be a killer feature! I guess I'm not the only one that would really like this feature.

RrOoSsSsOo:
I would like to start by saying that from my point of view this is a fundamental feature to have in Wallabag and should be implemented as soon as possible.
If it is implemented then a new "**Continue reading**" section should also be prepared in true _Netflix-style_ (and perhaps integrated well into a new homepage)

Issue #2641, opened 2016-11-22
Clean up migrations before releasing 2.2


nicosomb:
- [x] I want to add a comment in each migration to explain the aim of the migration
- [x] I want to add a check before each query to avoid some problems (eg: check if a column exists before creating it). I know it can be painful, but without these checks, we'll have many issues after releasing 2.2
- [x] add an explanation in doc.wallabag.org for each migration with its number, and the command to run this command individually.

PR #2646, opened 2016-11-24
Added explanations about migrations

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| Fixed tickets | #2641
| License       | MIT

WIP. I want to know if I'm on the good way. 

- [x] list all SQL code for shared hosting


nicosomb:
:fr: ✅ 
:de: need @jlnostr and @Strubbl help ;-)

nicosomb:
thank you @jlnostr! 

Now I have to write SQL code for shared hosting users.

nicosomb:
@j0k3r with this [migration](https://github.com/wallabag/wallabag/blob/2.2/app/DoctrineMigrations/Version20161001072726.php), I can't write pure SQL for the documentation. @tcitworld suggested me yesterday to write a PHP file for our shared hosted users. What do you think about that?

j0k3r:
Could be an alternative ...

nicosomb:
As I discussed on gitter to @tcitworld, if we create a new PHP file, we need to add in a new bundle (to access to the database parameters). 
**But**, to run a command in the bundle, we need to have a database up-to-date. 

Or a single PHP file which will include parameters.yml. 

Or all the queries in pure SQL.

tcitworld:
Rebase me ❤️

nicosomb:
I rebased you. 

And I added documentation (hardcoded SQL).

scripts/update.sh
```diff
@@ -16,4 +16,5 @@ git fetch --tags
 TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
 git checkout $TAG --force
 SYMFONY_ENV=$ENV $COMPOSER_COMMAND install --no-dev -o --prefer-dist
+php bin/console doctrine:migrations:migrate --env=$ENV
```
j0k3r: You should add `--no-interaction` otherwise it'll ask the user

Issue #2101, opened 2016-05-19
Shouldn't update an article if it cannot retrieve the content anymore


bmillemathias:
### Issue details

As explained in the issue #2100 I had this article fetched from lwn.net when I was using wallabag 1.x that I imported when I migrated to w 2.x with all other articles, which cannot being fetched nowaday.
Should wallabag prevent to update the content of the article if it cannot retrieve the content anymore.
### Environment
- wallabag version (or git revision): 2.0.4
- How did you install wallabag: `git clone`
- Last wallabag version that did not exhibit the issue: N//A
- php version: 7.0
- OS: Ubuntu 16.04
- type of hosting: dedicated
- which storage system you choose at install: MySQL:
### Steps to reproduce/test case
- Fetch a page from lwn.net with wallabag 1.X
- export the data and import it to a wallabag instance 2.0.x
- Fetch the page from the wallabag 2.0.x
- Wallabag informs it cannot retrieve the content of the page.

j0k3r:
Yeah it could be a good alternative to avoid loosing data from previous version.

j0k3r:
It's an old issue but refetching a content with the reload feature doesn't update the html when refetching failed (see https://github.com/wallabag/wallabag/pull/2482).

When using the v1 import, we don't refetch content from the website (for the moment, see https://github.com/wallabag/wallabag/issues/1842).

AFAIK it looks like this issue is fixed, can you confirm @bmillemathias?

nicosomb:
We also have this issue "If parsing fails, just redirects to the article" https://github.com/wallabag/wallabag/issues/1387

PR #2665, opened 2016-11-29
Fix catching update error on import, fixes #2658

### LINKED ISSUE #2658: Import process discards title, content, of entries that are not available anymore (eg. 404)
When importing an entry whose content is unavailable at the original URL, the fetchContent method overwrites the original title to "No title found". It should only update the title if fetching was successful, otherwise, the entry should be imported with the original URL.

jcharaoui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2658 #2101 
| License       | MIT

Even though I'm not very familiar with the framework, I did try updating tests but I couldn't figure out a way to get app parameters in the unit tests. Hopefully this work will still be useful to you somehow.

### DIFF
--- a/src/Wallabag/ImportBundle/Import/AbstractImport.php
+++ b/src/Wallabag/ImportBundle/Import/AbstractImport.php
@@ -94,6 +94,7 @@ abstract class AbstractImport implements ImportInterface
         $updatedEntry = clone $entry;
         $this->contentProxy->updateEntry($updatedEntry, $url, $content);
         if ($updatedEntry->getContent() == $this->fetching_error_message) {
+            $this->logger->warn('Unable to update imported entry ('.$entry->getUrl().')');
             return $entry;
         } else {
             return $updatedEntry;
--- a/src/Wallabag/ImportBundle/Import/AbstractImport.php
+++ b/src/Wallabag/ImportBundle/Import/AbstractImport.php
@@ -93,10 +93,12 @@ abstract class AbstractImport implements ImportInterface
     {
         $updatedEntry = clone $entry;
         $this->contentProxy->updateEntry($updatedEntry, $url, $content);
-        if ($updatedEntry->getContent() == $this->fetchingErrorMessage) {
+        if ($updatedEntry->getContent() === $this->fetchingErrorMessage) {
             $this->logger->warn('Unable to update imported entry ('.$entry->getUrl().')');
+
             return $entry;
         }
+
         return $updatedEntry;
     }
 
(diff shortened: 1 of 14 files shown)


src/Wallabag/ImportBundle/Import/AbstractImport.php
```diff
@@ -16,18 +16,20 @@
     protected $em;
     protected $logger;
     protected $contentProxy;
+    protected $fetching_error_message;
```
j0k3r: Variable name should be lowerCamelCase

src/Wallabag/ImportBundle/Import/AbstractImport.php
```diff
@@ -89,10 +91,13 @@ public function getMarkAsRead()
...
-        try {
-            return $this->contentProxy->updateEntry($entry, $url, $content);
-        } catch (\Exception $e) {
+        $updatedEntry = clone $entry;
+        $this->contentProxy->updateEntry($updatedEntry, $url, $content);
+        if ($updatedEntry->getContent() == $this->fetching_error_message) {
+            $this->logger->warn('Unable to update imported entry ('.$entry->getUrl().')');
             return $entry;
+        } else {
+            return $updatedEntry;
```
j0k3r: You don't need that `else` because the `if` already exit the function

nicosomb:
> I did try updating tests but I couldn't figure out a way to get app parameters in the unit tests.

You can access the container like that: `$client->getContainer()`. You can then get parameters.

jcharaoui:
@nicosomb Thanks for the follow-up. I tried that but unfortunately `$client` is undefined in the context where the `ImportInterface` is defined, (for example: `tests/Wallabag/ImportBundle/Import/WallabagV2ImportTest.php`), because it isn't a controller, if I understand correctly. The parameter needs to be passed to the `AbstractImport` class.

nicosomb:
Tests are OK. 

@j0k3r as you can see, I didn't manage to change base and fix conflicts: I don't know how to do that for a PR of someone else.

src/Wallabag/ImportBundle/Import/AbstractImport.php
```diff
@@ -89,11 +91,15 @@ public function getMarkAsRead()
...
-        } catch (\Exception $e) {
+        $updatedEntry = clone $entry;
+        $this->contentProxy->updateEntry($updatedEntry, $url, $content);
+        if ($updatedEntry->getContent() === $this->fetchingErrorMessage) {
+            $this->logger->warn('Unable to update imported entry ('.$entry->getUrl().')');
+
             return $entry;
         }
+
+        return $updatedEntry;
```
j0k3r: I think this whole stuff shouldn't be in AbstractImport because when it fails, you are re-doing thing that are already perform in updateEntry (like setting the domain name).
Maybe it should be better to move this logic in ContentProxy

jcharaoui: I agree!

jcharaoui:
Closing because I just submitted a better rebased PR, see #2679

PR #2679, opened 2016-12-03
Fix content from imported entried being discarded when URL goes bad

jcharaoui:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2658, #2101 
| License       | MIT

Improved version of PR #2665, and rebased on top of 2.2

If this is accepted I intend to work on an import option to disable update/fetching entries.


tests/Wallabag/CoreBundle/CoreKernelTestCase.php
```diff
@@ -0,0 +1,17 @@
...
+
+abstract class CoreKernelTestCase extends KernelTestCase
+{
+    protected $fetchingErrorMessage;
+
+    public function setUp()
+    {
+        self::bootKernel();
+        $container = self::$kernel->getContainer();
+        $this->fetchingErrorMessage = $container->getParameter('wallabag_core.fetching_error_message');
```
j0k3r: I'm against that. Sorry @nicosomb 

Because you boot the kernel when you don't necessary need it. And most case where you're going to use that message doesn't need the Symfony app to be booted.
You should better use a plain text when you need the fetching error message.

Taking example from `ContentProxyTest.php`:

```php
class ContentProxyTest extends \PHPUnit_Framework_TestCase
{
    private $fetchingErrorMessage = 'We can process it';

    public function testWithBadUrl()
    {
        // ...
        $proxy = new ContentProxy($graby, $tagger, $this->getTagRepositoryMock(), $this->getLogger(), $this->fetchingErrorMessage);
    }
}
```

nicosomb: It was to avoid duplication. But OK, we are in tests, so we can duplicate the string 👍

PR #2680, opened 2016-12-03
Limit rule to 255

### LINKED ISSUE #2678: Tagging Rules Allowed to be created that break Bagging.
### Issue details

Attempted to create a tagging rule: 
Match : `title matches "NSL" or title matches "NSA" or title matches "Snowden" or domainName is "eff.org" or content matches "Hemisphere" or content matches "Black Chamber" or content matches "Project SHAMROCK" or content matches "ECHELON" or content matches "Five Eyes" or content matches "USA Freedom Act" or content matches "Information Awareness Office" or content matches "CALEA"`
Tags: Surveillance, Domestic Spying, 4th

Interface allows this rule to be created, however after adding rule you will get  a 500 error:

> 500: Internal Server Error
> 
> Parse error: syntax error, unexpected 'Five' (T_STRING), expecting ']'

When removing the rule bagging starts to work again.

Based on what happens when I attempt to edit the rule I think the matching rules are limited to 256 characters. The interface and storage should either limit rules to 256 characters or change to a model that doesn't require character limits.

### Environment

* wallabag version (or git revision) that exhibits the issue: `2.1.5`
* How did you install wallabag? Via `git clone` or by downloading the package?  Self Hosted, Git clone followed instructions on site.
* Last wallabag version that did not exhibit the issue (if applicable): N/A (None tested)
* php version: 7.0.13
* OS:  CentOS 7
* type of hosting (shared or dedicated): VPS (Shared)
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MariaDB 10.1

Create the rule noted above (or I think any rule longer than 256 characters) and attempt to bag articles. You should receive 500 errors.

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | fix #2678
| License       | MIT

Limiting tagging rule to 255 to avoid database error when saving

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/TaggingRule.php
+++ b/src/Wallabag/CoreBundle/Entity/TaggingRule.php
@@ -28,6 +28,7 @@ class TaggingRule
      * @var string
      *
      * @Assert\NotBlank()
+     * @Assert\Length(max=255)
      * @RulerZAssert\ValidRule(
      *  allowed_variables={"title", "url", "isArchived", "isStared", "content", "language", "mimetype", "readingTime", "domainName"},
      *  allowed_operators={">", "<", ">=", "<=", "=", "is", "!=", "and", "not", "or", "matches"}
--- a/tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
+++ b/tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
@@ -515,6 +515,29 @@ class ConfigControllerTest extends WallabagCoreTestCase
         }
     }
 
+    public function testTaggingRuleTooLong()
+    {
+        $this->logInAs('admin');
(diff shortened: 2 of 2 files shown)


nicosomb:
Why don't change this field to a `text` one?

j0k3r:
@nicosomb it's already a text one: https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Form/Type/TaggingRuleType.php#L17

nicosomb:
I mean a blob field.

j0k3r:
And why a blob field? What's the different with a text field?

PR #2682, opened 2016-12-04
Ensure craue_config_setting migration are ok

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | could be
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

We now can run the migration every time, new line from `craue_config_setting` are properly checked.

PR #2683, opened 2016-12-05
Store credentials in DB

### LINKED ISSUE #2636: Store credentials for restricted site in database
Following https://github.com/wallabag/wallabag/pull/2317 (ie: ability to retrieve content from a website using an user subscription), we now need to store login / password in database instead of parameters.

Since we are going to use these credentials, we can't encrypt them because we won't be able to re-use them later. So they'll be stored as plain text in the database.
We should warn user about that. Mostly if user use a SAAS service (which might share multiple user).

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | https://github.com/wallabag/doc/pull/17
| Translation   | yes
| Fixed tickets | fix #2636
| License       | MIT

CRUD to add site_credential in database might be ok.
Password is asked every time user'll need to update an item (so password is never displayed).
As of now, the password is stored in plain text in database. I don't find a reliable solution to encode it. Maybe we can generate a base64, but it's simple to decode.

It's hard to embed it in the config panel because there are a lot of actions and different url. That's why it's accessible by this url `/site-credentials`. I was thinking to add it under _configuration_ in the menu only when the option is enabled. But the menu name might be too long. Any suggestions?

Some screenshots:

To be done:

- [x] replace site_credentials in the service `wallabag_core.guzzle_authenticator.config_builder` by the repository `SiteCredentialRepository`
- [x] add tests
- [x] add documentation
- [x] rebase'll be needed when https://github.com/wallabag/wallabag/pull/2682 will be merged
- [x] use [that scenario](https://github.com/defuse/php-encryption/blob/master/docs/Tutorial.md#scenario-1-keep-data-secret-from-the-database-administrator) to keep password secret


src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,138 @@
...
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Method;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route;
+use Wallabag\UserBundle\Entity\User;
+use Wallabag\CoreBundle\Entity\SiteCredential;
+
+/**
+ * SiteCredential controller.
+ */
+class SiteCredentialController extends Controller
```
nicosomb: Maybe we can define a global route?

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,138 @@
...
+{
+    /**
+     * Lists all User entities.
+     *
+     * @Route("/site-credential", name="site_credential_index")
```
nicosomb: sites-credentials?

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Displays a form to edit an existing site credential entity.
+     *
+     * @Route("/site-credential/{id}/edit", name="site_credential_edit")
+     * @Method({"GET", "POST"})
+     */
+    public function editAction(Request $request, SiteCredential $siteCredential)
+    {
+        $deleteForm = $this->createDeleteForm($siteCredential);
```
nicosomb: Need to check if we are the owner of the site credential.

src/Wallabag/CoreBundle/Repository/SiteCredentialRepository.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Wallabag\CoreBundle\Repository;
+
+/**
+ * SiteCredentialRepository.
+ *
+ * This class was generated by the Doctrine ORM. Add your own custom
```
nicosomb: You can remove these lines.

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,138 @@
...
+     * Lists all User entities.
+     *
+     * @Route("/site-credential", name="site_credential_index")
+     * @Method("GET")
+     */
+    public function indexAction()
+    {
+        $em = $this->getDoctrine()->getManager();
+
+        $credentials = $em->getRepository('WallabagCoreBundle:SiteCredential')->findAll();
```
nicosomb: `findAllByUserId` maybe.

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Deletes a site credential entity.
+     *
+     * @Route("/site-credential/{id}", name="site_credential_delete")
+     * @Method("DELETE")
+     */
+    public function deleteAction(Request $request, SiteCredential $siteCredential)
+    {
+        $form = $this->createDeleteForm($siteCredential);
```
nicosomb: Same check as editAction.

src/Wallabag/CoreBundle/Resources/translations/messages.fr.yml
```diff
@@ -511,6 +511,26 @@ user:
         delete_confirm: "Voulez-vous vraiment ?"
         back_to_list: "Revenir à la liste"
 
+site_credential:
+    page_title: Gestion des accès aux sites
+    new_site_credential: Créer un accès à un site
+    edit_site_credential: Éditer l'accès d'un site
+    description: "Ici vous pouvez gérer les accès aux différents sites. Ces accès permettent de récupérer des contenus sur des sites qui requiert une authentification ou un paywall"
```
nicosomb: requiert -> requièrent

src/Wallabag/CoreBundle/Resources/translations/messages.fr.yml
```diff
@@ -511,6 +511,26 @@ user:
...
+        yes: Oui
+        no: Non
+        create_new_one: Créer un nouvel accès à un site
+    form:
+        username_label: 'Identifiant'
+        host_label: 'Domaine'
+        password_label: 'Mot de passe'
+        save: "Sauvegarder"
+        delete: "Supprimer"
+        delete_confirm: "Voulez-vous vraiment ?"
```
nicosomb: "Voulez-vous vraiment supprimer ?" or "Êtes vous sur ?". But "Voulez-vous vraiment ?" seems strange to me.

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
j0k3r: :scream:

nicosomb: Well, it's an opensource solution, right?
It could be a nice feature!

tcitworld:
Rebased on master.

j0k3r:
Hum, I should rebase it against 2.3 instead I think

tcitworld:
Would love to encrypt the user's password here... Maybe with [this](https://github.com/defuse/php-encryption/blob/master/docs/Tutorial.md#scenario-2-encrypting-account-data-with-the-users-login-password) ?

j0k3r:
Might be a good idea.

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /**
+     * Check if the logged user can manage the given site credential.
+     *
+     * @param SiteCredential $siteCredential The site credential entity
+     */
+    private function checkUserAction(SiteCredential $siteCredential)
+    {
+        if (null === $this->getUser() || $this->getUser()->getId() != $siteCredential->getUser()->getId()) {
```
nicosomb: Maybe you can add a check here to see if the paywall feature is enabled or not.

j0k3r: The item isn't visible in the menu is the paywall feature is disabled.
Maybe, instead of adding a hard check, I can just display a banner telling the feature is disabled and performing change will do nothing

nicosomb:
Just tried and looks good for me 👍 It works. 

I agree with @di72nn in #2636.

j0k3r:
Last build should fail because it needs https://github.com/bdunogier/guzzle-site-authenticator/pull/16

src/Wallabag/CoreBundle/Controller/SiteCredentialController.php
```diff
@@ -0,0 +1,154 @@
...
+
+            $this->get('session')->getFlashBag()->add(
+                'notice',
+                $this->get('translator')->trans('flashes.site_credential.notice.added', ['%host%' => $credential->getHost()])
+            );
+
+            return $this->redirectToRoute('site_credentials_edit', array('id' => $credential->getId()));
+        }
+
+        return $this->render('WallabagCoreBundle:SiteCredential:new.html.twig', array(
```
tcitworld: Instead of redirecting to the edit page, let's redirect to the list instead.

j0k3r:
But before validating it, we should choose a way to handle password in database.

fradav:
Tested it, and it works beautifuly, got two paywall on it (mediapart and monde-diplomatique) both working fine with this. Congrats!

nicosomb:
Now, passwords are stored in plain text?

j0k3r:
Yep

fradav:
Even without password encryption, this is already immensely helpful for in-house server: I got a docker image I can deploy anywhere and store paywall credentials, as opposed to hack the docker config to push the credentials in the parameters.yml at build (and to be scared stiff stiff not to spread passwords to the wide internet with the wrong git/docker push).

fradav:
Tested for myself, works as expected, passwords are now big hashed keys in the database. Nice!

j0k3r:
@fradav thanks for being so proactive in testing that feature!

fradav:
@j0k3r you bet I do, I wanted paywalls access in wallabag/pocket since a long time ago, and I tried various hacks (reverse proxy) which didn't turn out very well, and the parameter-stored credential feature was still on the "hackish" side. Now with the help of this and wallabako, I think I got the best of two world: the (almost) newspaper-like readability (old world, provided by e-ink reader), and the always-on feed from the net with wallabag ecosystem. I cannot express how happy I am now with this. Many thanks to you, @bdunogier, who implemented the core of the feature, and of course to the whole team of contributors.

di72nn:
Usernames should be encrypted too.

nicosomb:
Tested too, and 👍

src/Wallabag/CoreBundle/Resources/translations/messages.fr.yml
```diff
@@ -514,11 +515,31 @@ user:
         twofactor_label: "Double authentification"
         save: "Sauvegarder"
         delete: "Supprimer"
-        delete_confirm: "Voulez-vous vraiment ?"
+        delete_confirm: "Êtes-vous sur ?"
```
Progi1984: sur => sûr ?

src/Wallabag/CoreBundle/Helper/CryptoProxy.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Keep first and last character and put some stars in between.
+     *
+     * @param string $value Value to mask
+     *
+     * @return string
+     */
+    private function mask($value)
+    {
+        return $value[0].'*****'.$value[strlen($value) - 1];
```
di72nn: That'll break if the value is empty.

tcitworld: It shouldn't since we have the `@Assert\NotBlank()` line on the field, but let's add a check anyway.

src/Wallabag/CoreBundle/GuzzleSiteAuthenticator/GrabySiteConfigBuilder.php
```diff
@@ -87,7 +87,8 @@ public function buildForHost($host)
 
         $config = new SiteConfig($parameters);
 
-        // do not leak password in log
+        // do not leak usernames and passwords in log
+        $parameters['username'] = '**masked**';
```
j0k3r: So, is it really useful to put them in log if we only display masked values? 🤔

src/Wallabag/CoreBundle/Form/Type/SiteCredentialType.php
```diff
@@ -19,6 +19,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ])
             ->add('username', TextType::class, [
                 'label' => 'site_credential.form.username_label',
+                'data' => '',
```
j0k3r: What it does?

src/Wallabag/CoreBundle/Resources/views/themes/material/SiteCredential/index.html.twig
```diff
@@ -16,14 +16,16 @@
                                 <tr>
                                     <th>{{ 'site_credential.form.host_label'|trans }}</th>
                                     <th>{{ 'site_credential.form.username_label'|trans }}</th>
+                                    <th>{{ 'site_credential.form.password_label'|trans }}</th>
```
j0k3r: I think it'd not useful to display two columns with **** only.

src/Wallabag/CoreBundle/Form/Type/SiteCredentialType.php
tcitworld: Makes the username field blank, otherwise it's filled with the encrypted value.

src/Wallabag/CoreBundle/GuzzleSiteAuthenticator/GrabySiteConfigBuilder.php
tcitworld: The domain name is logged, right ?

src/Wallabag/CoreBundle/Resources/views/themes/material/SiteCredential/index.html.twig
tcitworld: Yeah. Let's get rid of it.

src/Wallabag/CoreBundle/GuzzleSiteAuthenticator/GrabySiteConfigBuilder.php
j0k3r: Yeah I forgot that we log all other values

j0k3r:
Are we good to merge that one?

nicosomb:
:+1:

nicosomb:
🎉

PR #2686, opened 2016-12-06
Casted maxIterations for Redis import

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | N/A
| License       | MIT

Because Simpleue wants an integer (see https://github.com/javibravo/simpleue/blob/master/src/Simpleue/Worker/QueueWorker.php#L88), we need to cast the parameter. 

The problem is in the library I think, but it's faster to fix on our side.


j0k3r:
You can still submit a PR on simpleue

nicosomb:
Yes ,it's planned. Will do that after my lunch

PR #2696, opened 2016-12-12
Run migration on each test

### LINKED ISSUE #2666: Run each migration on tests
Actually migrations aren't really tested.
When we want to add a new field or update sth in the database, we just add a new migration and we (usually) only test it on your dev platform. That's why we sometimes got errors when releaseing minor version. Because migration weren't fully tested.

What we can do:
- return to the v2.0 and retrieve the generated SQL schema for each platform
- create the database with that SQL during tests
- run each migration to have the latest schema

With that running, each migration will be launched everytime. And it'll ensure they are always OK on each platform.

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | kind of
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | :trollface: 
| Documentation | no
| Translation   | no
| Fixed tickets | might fix #2666
| License       | MIT

The idea is to run migration on each test to be sure they are ok.

- I've extracted the v2.0.0 sql schema for each SGBD
- the file is inserted
- each migration are launched
- test suite run on the modified database

Changes

- **MySQL**: every thing is ok
- **PostgreSQL**: `doctrine:database:import` doesn't work on a file with multiple queries. I use `psql` to import the file.
- **SQLite**: the migration doesn't run with a strange error `There is no table with name '/home/jeremy/github/wallabag/app/..wallabag_entry' in the schema.`


j0k3r:
Postgres fail on delete cascade :thinking: 
> Foreign key violation: 7 ERROR:  update or delete on table \\"wallabag_entry\\" violates foreign key constraint \\"fk_c9f0dd7cba364942\\" on table \\"wallabag_entry_tag\\"

nicosomb:
👏 👏 Thank you @j0k3r!

PR #2699, opened 2016-12-12
Removed user join in query builder

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | N/A
| License       | MIT

I removed join on user table. We only need user id.
I think we can have performance gain. 
I saw very looong queries on app.wallabag.it with this QueryBuilder.


nicosomb:
Yes, that's why I wanted yesterday, and I saw `Only used for tests.`. But I can if you want.

PR #2702, opened 2016-12-14
Added index on entry.uuid

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   |no
| Fixed tickets | N/A
| License       | MIT

Public sharing was quite long on app.wallabag.it 🐌

### DIFF
--- a/app/DoctrineMigrations/Version20160410190541.php
+++ b/app/DoctrineMigrations/Version20160410190541.php
@@ -33,8 +33,9 @@ class Version20160410190541 extends AbstractMigration implements ContainerAwareI
 
         $this->skipIf($entryTable->hasColumn('uuid'), 'It seems that you already played this migration.');
 
-        $entryTable->addColumn('uuid', 'guid', [
+        $entryTable->addColumn('uuid', 'string', [
             'notnull' => false,
+            'length' => 23,
         ]);
         $this->addSql('INSERT INTO '.$this->getTable('craue_config_setting')." (name, value, section) VALUES ('share_public', '1', 'entry')");
     }
--- a/app/DoctrineMigrations/Version20160410190541.php
+++ b/app/DoctrineMigrations/Version20160410190541.php
@@ -31,9 +31,9 @@ class Version20160410190541 extends AbstractMigration implements ContainerAwareI
     {
         $entryTable = $schema->getTable($this->getTable('entry'));
 
-        $this->skipIf($entryTable->hasColumn('uuid'), 'It seems that you already played this migration.');
+        $this->skipIf($entryTable->hasColumn('uid'), 'It seems that you already played this migration.');
 
-        $entryTable->addColumn('uuid', 'string', [
+        $entryTable->addColumn('uid', 'string', [
             'notnull' => false,
             'length' => 23,
         ]);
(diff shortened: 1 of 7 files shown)


nicosomb:
And the build is green? 😱

Kdecherf:
From MySQL 5.7 documentation:

> For string columns, indexes can be created that use only the leading part of column values, using col_name(length) syntax to specify an index prefix length:
> 
> - Prefixes can be specified for CHAR, VARCHAR, BINARY, and VARBINARY column indexes.
> 
> - **Prefixes must be specified for BLOB and TEXT column indexes.**

On a side-note, IMO I don't see the benefit of having TEXT for the uuid field. It could be replaced by CHAR(36) (_and so the prefix arg for index could become optional_).

nicosomb:
I fixed the problem with creation table. 
@Kdecherf I changed the field type to guid. MySQL now creates a char(36) field. We don't need a migration, because our migration was correct (field type to guid, see https://github.com/wallabag/wallabag/blob/2.2/app/DoctrineMigrations/Version20160410190541.php#L36)

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -44,7 +47,7 @@ class Entry
     /**
      * @var string
      *
-     * @ORM\Column(name="uuid", type="text", nullable=true)
+     * @ORM\Column(name="uuid", type="guid", nullable=true)
```
j0k3r: :+1:

nicosomb:
The error in the test is: 

```
An exception occurred while executing &#039;UPDATE &quot;wallabag_entry&quot; SET uuid = ?, updated_at = ? WHERE id = ?&#039; with params [&quot;585c4abe3c5116.40862576&quot;, &quot;2016-12-22 21:50:54&quot;, 6]:

SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for uuid: &quot;585c4abe3c5116.40862576&quot; (500 Internal Server Error)
```

I tried with a more recent postgres version (we require postgres > 9.2 but postgres on travis is 9.1.x), I tried by creating new extension (uuid-ossp), but I don't see what I can do to fix the problem.

app/DoctrineMigrations/Version20160410190541.php
```diff
@@ -33,8 +33,9 @@ public function up(Schema $schema)
 
         $this->skipIf($entryTable->hasColumn('uuid'), 'It seems that you already played this migration.');
 
-        $entryTable->addColumn('uuid', 'guid', [
+        $entryTable->addColumn('uuid', 'string', [
             'notnull' => false,
+            'length' => 23,
```
Kdecherf: UUID is 36 chars long

Kdecherf: UUID is 36-chars long

j0k3r: Yep but we don't generate a UUID, because we are using `uniqid()` which doesn't generate an UUID.

Kdecherf: I see, but in this case naming the field 'uuid' is quite misleading

nicosomb: what do you propose?

nicosomb:
Squashed and already rebased.

app/DoctrineMigrations/Version20160410190541.php
Kdecherf: You may rename the field "uid" (_or "uniqid"_)

tcitworld: ping @nicosomb

nicosomb:
Field renamed.

nicosomb:
Thank you @j0k3r, @tcitworld and @Kdecherf for your comments.

PR #2708, opened 2016-12-16
Import disableContentUpdate

jcharaoui:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | https://github.com/wallabag/doc/issues/4
| Translation   | no
| Fixed tickets | 
| License       | MIT

Adding a disableContentUpdate option, which defaults to false, to allow users to control whether Wallabag will fetch content from the Internet during an import operation. Also fixes a few minor things.

Still WIP, we might want to add the option to the UI and API. Also the recent changes to ContentProxy require a good revamp of tests.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -51,9 +51,9 @@ class ContentProxy
     /**
      * Import entry using either fetched or provided content.
      *
-     * @param Entry  $entry                Entry to update
-     * @param array  $content              Array with content provided for import with AT LEAST keys title, html, url to skip the fetchContent from the url
-     * @param bool   $disableContentUpdate Whether to skip trying to fetch content using Graby
+     * @param Entry $entry                Entry to update
+     * @param array $content              Array with content provided for import with AT LEAST keys title, html, url to skip the fetchContent from the url
+     * @param bool  $disableContentUpdate Whether to skip trying to fetch content using Graby
      */
     public function importEntry(Entry $entry, array $content, $disableContentUpdate = false)
     {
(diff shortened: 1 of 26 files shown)


jcharaoui:
A quick assessement would be appreciated. I'd like to continue working on tests, but I'd like to know if you're interested in merging this feature.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -158,15 +183,17 @@ public function assignTagsToEntry(Entry $entry, $tags, array $entitiesReady = []
...
+     * Validate that the given content has at least a title and a url.
      *
      * @param array $content
-     *
-     * @return bool true if valid otherwise false
      */
     private function validateContent(array $content)
     {
-        return isset($content['title']) && isset($content['html']) && isset($content['url']) && isset($content['language']) && isset($content['content_type']);
+        if (!(isset($content['title']) && !empty($content['title']))) {
```
j0k3r: You can just use `!empty($content['title'])` (which handle isset too)

src/Wallabag/ImportBundle/Import/AbstractImport.php
```diff
@@ -82,21 +83,43 @@ public function getMarkAsRead()
     }
 
     /**
+     * Set whether articles should be fetched for updated content.
+     *
+     * @param bool $markAsRead
```
j0k3r: Wrong copy/paste

j0k3r:
I'll wait for https://github.com/wallabag/wallabag/pull/3106 to be merged because it'll generate some conflict and some refacto I guess.
But for the moment, tests should be fine.

@jcharaoui will work on that PR again? :slightly_smiling_face:

jcharaoui:
I hope to pick it up again, but not in the near future. If anyone wants to pick up where I left off, go right ahead!

j0k3r:
Ok thanks for the quick feedback, I'll try to finish it then.

j0k3r:
The only chance here is adding a new option to the import command: `disableContentUpdate` which disable fetching content from an url when enabled.
This should speed up any imports (because it'll save the entry without fechting any urls) but entry without content will still have no content.

Should we document that somewhere? (we don't have any documentation about commends except one which only list all commands)

nicosomb:
We can start a more detailed documentation with this command.

j0k3r:
Thanks @jcharaoui for starting the job!
It'll be available in 2.3

Issue #2475, opened 2016-10-20
Text truncated in menu


nicofrand:
### Issue details

"Marquer comme non lu" is truncated in the menu. I think the text should be broken into two lines.
Looks to me like there is enough space on the right to display "lu". Maybe the margin could be reduced ?

![wallabag-text-truncated](https://cloud.githubusercontent.com/assets/3419050/19569082/9c6c5a42-96f4-11e6-967e-e61fa1d7526a.png)
### Environment

Wallabag 2.1.2
### Steps to reproduce/test case

Mark an article as read and then open the menu to look at the "Marquer comme non lu" item.

PR #2713, opened 2016-12-19
label truncated 'Marquer comme non lu'

foxmask:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | dunno
| Documentation | no
| Translation   | yes
| Fixed tickets | #...
| License       | MIT

### Issue details

when we are on an article that has already been read, the label action to set it as unread is truncated in french "Marquer comme non" may be "Marq. comme non lu" do the trick :)

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.1.5
* php version: 7 
* OS: debian 5.8
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): postgresql 9.6

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/translations/messages.fr.yml
+++ b/src/Wallabag/CoreBundle/Resources/translations/messages.fr.yml
@@ -185,7 +185,7 @@ entry:
             back_to_top: "Revenir en haut"
             back_to_homepage: "Retour"
             set_as_read: "Marquer comme lu"
-            set_as_unread: "Marquer comme non lu"
+            set_as_unread: "Marq. comme non lu"
             set_as_starred: "Mettre en favori"
             view_original_article: "Article original"
             re_fetch_content: "Recharger le contenu"


nicosomb:
I'm not fan of the truncated text. OK for the mobile view, but on desktop view, I want to read the whole sentence...

foxmask:
The text is truncated with the desktop version
So you will need to expand the size of the menu.
You can rejzct the PR 🤗

nicosomb:
Idea: you can use CSS class to show/hide div on mobile view. Maybe we can have 2 divs, one for desktop view, and an other for mobile view.

foxmask:
You will need @media query for that.

nicosomb:
We already have class for that: `hide-on-med-and-down` and `hide-on-large-only`

j0k3r:
Instead of truncating text (not a big fan too) we might try to find a shorter version :thinking:

nicosomb:
`Archiver` ?

foxmask:
Marquer comme non lu , if its unread how could it be Archiver ?

nicosomb:
Oops! 
https://fr.wiktionary.org/wiki/d%C3%A9sarchiver ?

foxmask:
Perso, sémantiquement parlant, "marquer comme lu" ça ne veut pas dire "archiver" alors desarchiver ça ne veut pas pas non plus dire "non lu" . Du coup si on mélange "lu/desarchiver" ça manque de cohérence, non ?  😮

nicofrand:
You should put "Fix #2475" in the description or commit message :).

j0k3r:
Bon, on fait quoi là?
On mets "lu" ?

Pocket utilise "archive" par tout. En disant par exemple "Élément archivé".

foxmask:
Vas pour faire comme pocket :)

tcitworld:
My webpack PR increases the width of the sidebar, so this is fixed anyway.

j0k3r:
Okay, so we don't care anymore :+1:

PR #2714, opened 2016-12-20
Fix the box shadow on the card entry

### LINKED ISSUE #2597: Help about shortcuts is missing
In 2.2, the documentation inside wallabag is missing about shortcuts (there is nothing in /howto)
So it's really unclear on how we shoud use them and also how can we disabled them.

Also, the display of the _selected_ article is really disturbing me when we arrive on a the default page. It is not clear that the first item is selected and then we can navigate. At first, I though of a bug :confused: 

image

Following https://github.com/wallabag/wallabag/issues/512

Simounet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

Fixing the box shadow on the card entry. It was set on the parent DOM node and the effect wasn't so graceful. I also did a little facto on some duplicated node fetches.

Have a nice day!

Should fix #2597


nicosomb:
Screenshot please?

PR #2724, opened 2016-12-27
Fix entry reading time

Simounet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no

Before:
Reading time of entries on the main view was broken because of a wrong rounded value.

After:
The reading time code is now shared between both views.


tcitworld:
Shouldn't this be fixed on the baggy theme too ?

Simounet:
I verified and it seems to be fine on the baggy theme. The code is not shared between the views. I can do that if you want.

PR #2729, opened 2016-12-29
Fix double overlay on expanded right sidenav

Simounet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Tests pass?   | yes

Before:
- Click on the filters button
- check the overlay color
- exit from this menu
- click again
- sometimes the overlay is darker

On mobile, it was even weirder because the overlay wasn't always hidden.

Why:
- JS was calling `sideNav()` for both `.button-collapse-right` buttons everytime

After:
- Only one overlay displayed


nicosomb:
I already had this bug. Thank you for this fix 👍

PR #2751, opened 2017-01-14
Improved Guzzle subscribers extensibility

bdunogier:
Allows 3rd parties to register new guzzle subscribers by adding extra calls to the http_client_factory service. Used in https://github.com/bdunogier/guzzle-site-authenticator/pull/14 to fix a wrong URI during a login process.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| Fixed tickets | n/a
| License       | MIT

Allows 3rd parties to register new guzzle subscribers by adding extra calls to `addSubscribers` on the `http_client_factory` service.

Needed to fix a bug with a particular site's response. I think it improves wallabag's extensibility, and better decouples features such as paywalls.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/HttpClientFactory.php
+++ b/src/Wallabag/CoreBundle/Helper/HttpClientFactory.php
@@ -27,7 +27,7 @@ class HttpClientFactory
      *
      * @param \GuzzleHttp\Cookie\CookieJar $cookieJar
      * @param string                       $restrictedAccess This param is a kind of boolean. Values: 0 or 1
-     * @param LoggerInterface                       $logger
+     * @param LoggerInterface              $logger
      */
     public function __construct(CookieJar $cookieJar, $restrictedAccess, LoggerInterface $logger)
     {
--- a/src/Wallabag/CoreBundle/Helper/HttpClientFactory.php
+++ b/src/Wallabag/CoreBundle/Helper/HttpClientFactory.php
@@ -13,8 +13,8 @@ use Psr\Log\LoggerInterface;
  */
 class HttpClientFactory
 {
-    /** @var \GuzzleHttp\Event\SubscriberInterface */
-    private $authenticatorSubscriber;
+    /** @var [\GuzzleHttp\Event\SubscriberInterface] */
+    private $subscribers = [];
 
     /** @var \GuzzleHttp\Cookie\CookieJar */
     private $cookieJar;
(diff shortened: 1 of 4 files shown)


src/Wallabag/CoreBundle/Helper/HttpClientFactory.php
```diff
@@ -23,13 +23,15 @@ class HttpClientFactory
     /**
      * HttpClientFactory constructor.
      *
-     * @param \GuzzleHttp\Event\SubscriberInterface $authenticatorSubscriber
+     * @param \GuzzleHttp\Event\SubscriberInterface $authenticatorSubscriber Deprecated since 2.2. Use addSubscriber instead
```
j0k3r: 2.2 wasn't yet released so it's safe to break thing and avoid deprecated stuff

bdunogier:
Removed the deprecated argument altogether.

j0k3r:
Looks good, you can squash your last commit

bdunogier:
Squashed.

But I've just made another change. It can be moved to another PR, but it should be fairly safe, and it is kind of related: it splits, in the graby config -> guzzle auth config mapping, `login_extra_fields` to a hash by splitting the string on `=`.

It works for me, but I'm open for better places or ways to do this.

j0k3r:
We is that other change? Elsewhere? I can't see it in this PR

bdunogier:
Forgot the --force after squashing, sorry @j0k3r :)

j0k3r:
Looks like you need to update your tests too.

bdunogier:
Travis should be green with the last one.

bdunogier:
The extensibility part is final, as far as I am concerned.

About the change to `loginExtraFields`, I'm not sure: I'm moving to a different approach on paywall that may make it not as useful as it was in the previous iteration. On the other hand, it makes no sense that you can set fields, but can't distinguish key/value, so overall, it's a good change to have.

I'd say "your call" on the 2nd one :)

bdunogier:
And rebased on top of the latest 2.3.

j0k3r:
@bdunogier I've fixed CS.
Can we merge that with / without the PR on guzzle-site-authenticator or do we need them to be both merged?

bdunogier:
I think you can. The related guzzle plugin PR takes advantage of it, but isn't required for it to work.

PR #2763, opened 2017-01-17
Sort list of available tags

janLo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

This adds an orderBy clause to the listing of tags for a user. This makes the list at `/tag/list` more useable as the order of the tags seems not to be that random.


j0k3r:
Thanks @janLo !

PR #2774, opened 2017-01-22
add cli export

tcitworld:
- [x] One more test for export file path
- [ ] Little documentation

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | WIP
| Translation   | no
| Fixed tickets | 
| License       | MIT

### DIFF
--- a/src/Wallabag/CoreBundle/Command/ExportCommand.php
+++ b/src/Wallabag/CoreBundle/Command/ExportCommand.php
@@ -23,7 +23,7 @@ class ExportCommand extends ContainerAwareCommand
                 'User from which to export entries'
             )
             ->addArgument(
-                'filename',
+                'filepath',
                 InputArgument::OPTIONAL,
                 'Path of the exported file'
             )
@@ -46,7 +46,7 @@ class ExportCommand extends ContainerAwareCommand
 
         $output->write(sprintf('Exporting %d entrie(s) for user « <comment>%s</comment> »... ', count($entries), $user->getUserName()));
 
-        $filePath = $input->getArgument('filename');
+        $filePath = $input->getArgument('filepath');
         if (!$filePath) {
             $filePath = $this->getContainer()->getParameter('kernel.root_dir') . '/../' . sprintf('%s-export', $user->getUsername());
         }
--- a/src/Wallabag/CoreBundle/Command/ExportCommand.php
+++ b/src/Wallabag/CoreBundle/Command/ExportCommand.php
@@ -7,7 +7,6 @@ use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Output\StreamOutput;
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Command/ExportCommand.php
```diff
@@ -0,0 +1,82 @@
...
+            ->getBuilderForAllByUser($user->getId())
+            ->getQuery()
+            ->getResult();
+
+        $output->write(sprintf('Exporting %d entrie(s) for user « <comment>%s</comment> »... ', count($entries), $user->getUserName()));
+
+        $filePath = $input->getArgument('filepath');
+        if (!$filePath) {
+            $filePath = $this->getContainer()->getParameter('kernel.root_dir') . '/../' . sprintf('%s-export', $user->getUsername());
+        }
```
nicosomb: maybe you can add lines to separate blocks here.

nicosomb: I don't if it's a good thing to have export filenames based on username (even if I know that these files are not available from the web server).

tcitworld: Well it's better in order to find them.

tcitworld:
No idea what the hell `php-cs-fixer` does here.

src/Wallabag/CoreBundle/Command/ExportCommand.php
j0k3r: I'm ok with that.

src/Wallabag/CoreBundle/Command/ExportCommand.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Fetches a user from its username.
+     *
+     * @param string $username
+     *
+     * @return \Wallabag\UserBundle\Entity\User
+     */
+    private function getUser($username)
```
j0k3r: You don't really need a dedicated function when you only use that function once.

tests/Wallabag/CoreBundle/Command/ExportCommandTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+        $command = $application->find('wallabag:export');
+
+        $tester = new CommandTester($command);
+        $tester->execute([
+            'command' => $command->getName(),
+            'username' => 'admin',
+        ]);
+
+        $this->assertContains('Exporting 6 entrie(s) for user « admin »... Done', $tester->getDisplay());
```
j0k3r: Maybe you can add an extra test on the file (like you did for the test below)

src/Wallabag/CoreBundle/Command/ExportCommand.php
nicosomb: OK with the username in the filename?

j0k3r: Yep. It is sluggified. So it should be safe.

tcitworld:
Will update doc when it will be migrated (need a proper admin documentation, separated from user docs).

Issue #2765, opened 2017-01-18
iOS App can't connect to wallabag.it


Zettt:
I think this might still be in the works but am wondering if the iOS app (Wallabag 2 Official), or any other clients for that matter, can be configured with the non-self-hosted version of Wallabag (the paid version). I couldn't find useful information in the documentation, other than I'm probably supposed to create a new "developer app".

Thanks.

nicosomb:
Hello @Zettt! 

Yes, you need to create a developer application. Then, you fill the form with the given informations. 

We need to write more documentation about these steps;

Zettt:
I'd appreciate it.

Zettt:
Hello @nicosomb. Thanks for helping me so far. I figured I can contribute the documentation for this. See: https://github.com/wallabag/wallabag/pull/2788

nicosomb:
Thank you @Zettt. 

I will translate your documentation in french. That's why this issue will stay open.

Zettt:
@nicosomb Thanks. Good to know my work is helpful. I'd translate to German but my time doesn't allow it. Keep up the good work!

nicosomb:
Done here https://github.com/wallabag/wallabag/pull/2882

Whitou:
Hi, sorry to re-open this dusty post, but I can't get my phone to connect on wallabag.it.
I created the api client and I get an "Unknown error" each time. Deleted the client, re-created it... Nothing...
Also created another account, did not help.
I'm wondering if I'm using the right URL (tried https://wallabag.it, app.wallabag.it, v2.wallabag.org, wallabag.it/MyUserAccount,...)

PR #2788, opened 2017-01-27
Documentation on how to configure mobile apps to work with wallabag.it

Zettt:
See issue #2765.

### DIFF
--- a/docs/en/user/configuring_mobile.rst
+++ b/docs/en/user/configuring_mobile.rst
@@ -1,0 +1,13 @@
+Configuring Mobile Apps to Work With wallabag.it
+================================================
+
+Steps to configure the app
+---------------------------
+
+- First *create a new client* in the ``Developer`` section. The name of your app doesn't matter as much as the ``Client ID`` and the ``Client secret``. Write these two values down.
+- When you download a mobile app, it's going to ask for the server address of the hosted wallabag instance. With wallabag.it that is: ``https://app.wallabag.it``.
+- It's also going to ask for the aforementioned ``Client ID`` and ``secret``. Please insert those in the text fields, when asked.
+- Lastly you need to provide your ``username`` and ``password``. These are the same credentials you use to login in to wallabag.it.
+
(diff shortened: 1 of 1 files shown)


j0k3r:
From where did you get that iOS screenshot? It doesn't look like the official iOS app.

Zettt:
It isn't, but the official iOS app, apparently, doesn't want to have anything to do with wallabag itself. Therefore I figured it's irrelevant where the screenshot comes from. The screenshot is from Wallabag Pro.

j0k3r:
> It isn't, but the official iOS app, apparently, doesn't want to have anything to do with wallabag itself. 

Following your reflexion, Wallabag Pro should be handled in the same way. So we won't have to put Wallabag Pro screenshot in the doc :thinking: 

Well I don't agree about _the "official iOS app" to have anything to do with wallabag itself_.

The problem you are encountered is related to Wallabag Pro, not wallabag itself, nor wallabag.it, nor the official iOS app. So it have nothing to do in the official documentation.

If we have to define something in the doc, I think it's preferable to be oriented with official app. We won't put a screenshot for all app which use wallabag. It'll be a mess. Just think if Twitter put screenshot of each app on the way to connect to Twitter using an app? :flushed: No way.

So I think the best solution is to put a screenshot of the official app **as an example**.

Also, there shouldn't be a mention to wallabag.it in the official wallabag doc. It should just be called "wallabag" instead.

Zettt:
So you are saying it's better to make instructions with three screenshots, instead of one, although the information those three screenshots display are the same as this one screenshot?

nicosomb:
Hello @Zettt ! 

@j0k3r said that we only put documentation about our iOS official application. If you update your screenshots with our iOS application, we can merge your PR.

Zettt:
@nicosomb I've deleted the screenshot now. I don't know how this now works. Do I need to send another PR?

nicosomb:
@bourvill Is this documentation correct?

bourvill:
Sound like good.

Maybe squash this pr ?

Issue #2795, opened 2017-01-28
Database migrations issue during upgrade from 2.1.5 to 2.2.0


Aldarone:
### Issue details

Please provide issue details here.
Remember, this is _not_ a place to ask questions. For that, go to http://gitter.im/wallabag/wallabag.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.2.0
* How did you install wallabag? Using git clone
* Last wallabag version that did not exhibit the issue (if applicable): 2.1.5
* php version: PHP 5.6.26-1+deb.sury.org~trusty+1
* OS: Ubuntu 14.04.5 LTS
* type of hosting (shared or dedicated): Dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): SQLite

### Steps to reproduce/test case

After upgrading from 2.1.5 to 2.2.0 using `make update` I can't log in anymore.

At first I was just sent back to the login page and I tried to reset my password, but there was this error : 

```
An exception occurred while executing 'SELECT t0.username AS username_1, t0.username_canonical AS username_canonical_2, t0.email AS email_3, t0.email_canonical AS email_canonical_4, t0.enabled AS enabled_5, t0.salt AS salt_6, t0.password AS password_7, t0.last_login AS last_login_8, t0.confirmation_token AS confirmation_token_9, t0.password_requested_at AS password_requested_at_10, t0.roles AS roles_11, t0.id AS id_12, t0.name AS name_13, t0.created_at AS created_at_14, t0.updated_at AS updated_at_15, t0.authCode AS authCode_16, t0.twoFactorAuthentication AS twoFactorAuthentication_17, t0.trusted AS trusted_18, t19.id AS id_20, t19.theme AS theme_21, t19.items_per_page AS items_per_page_22, t19.language AS language_23, t19.rss_token AS rss_token_24, t19.rss_limit AS rss_limit_25, t19.reading_speed AS reading_speed_26, t19.pocket_consumer_key AS pocket_consumer_key_27, t19.action_mark_as_read AS action_mark_as_read_28, t19.list_mode AS list_mode_29, t19.user_id AS user_id_30 FROM "wallabag_user" t0 LEFT JOIN "wallabag_config" t19 ON t19.user_id = t0.id WHERE t0.email_canonical = ? LIMIT 1' with params ["redacted@fake.com"]: SQLSTATE[HY000]: General error: 1 no such column: t19.action_mark_as_read
```

[...]

PR #2809, opened 2017-01-30
Fixed duplicate entry for share_public in craue_setting_table

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| Fixed tickets | #2795
| License       | MIT

- [x] Duplicate entry for `share_public` internal setting (see https://github.com/wallabag/wallabag/issues/2795#issuecomment-275868820)
- [x] Missing documentation about renaming `uuid` to `uid` into entry table. (see https://github.com/wallabag/wallabag/issues/2795#issuecomment-275934106)
- [x] some migrations aren't well played

### DIFF
--- a/app/DoctrineMigrations/Version20160410190541.php
+++ b/app/DoctrineMigrations/Version20160410190541.php
@@ -34,7 +34,7 @@ class Version20160410190541 extends AbstractMigration implements ContainerAwareI
     {
         $entryTable = $schema->getTable($this->getTable('entry'));
 
-        $this->skipIf($entryTable->hasColumn('uid'), 'It seems that you already played this migration.');
+        $this->skipIf($entryTable->hasColumn('uid') || $entryTable->hasColumn('uuid'), 'It seems that you already played this migration.');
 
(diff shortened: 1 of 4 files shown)


nicosomb:
About migrations which are not played (SQLite only), I wrote this documentation: https://gist.github.com/nicosomb/7c537995f2b845a30b4d6cdf23c1e22c

It will be published on the blog post for 2.2.1.

nicosomb:
We also forgot that when users lauched `make update`, the commands which are played are those present in `update.sh` from 2.1.6, where migration command is missing. 
That's why they need to launch `make update` twice ([as explained here](https://www.reddit.com/r/selfhosted/comments/5qyid6/has_anyone_had_trouble_upgrading_to_the_latest/dd3m9ta/)): 
* first `make update` checkouts the 2.2 files
* second `make update` launched the migrations

PR #2832, opened 2017-02-02
Search by term: extend to entries url

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | idk
| Documentation | no
| Translation   | no
| Fixed tickets | -
| License       | MIT

This PR extends search to entries url to provide an alternative way to domain filtering.

### DIFF
--- a/tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
+++ b/tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
@@ -1093,5 +1093,29 @@ class EntryControllerTest extends WallabagCoreTestCase
         $crawler = $client->submit($form, $data);
 
         $this->assertCount(0, $crawler->filter('div[class=entry]'));
+
+        // test url search on list of all articles
+        $crawler = $client->request('GET', '/all/list');
+
+        $form = $crawler->filter('form[name=search]')->form();
+        $data = [
+            'search_entry[term]' => 'domain', // the search will match an entry with 'domain' in its url
+        ];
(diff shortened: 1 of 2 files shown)


tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
```diff
@@ -1093,5 +1093,17 @@ public function testSearch()
         $crawler = $client->submit($form, $data);
 
         $this->assertCount(0, $crawler->filter('div[class=entry]'));
+
+        // test domain on list of all articles
+        $crawler = $client->request('GET', '/all/list');
+
+        $form = $crawler->filter('form[name=search]')->form();
+        $data = [
+            'search_entry[term]' => 'domain', // a string not available in the database
```
j0k3r: You might change the comment and maybe use a real domain name instead?

Kdecherf: Yes, I will update the comment to say that the filtering is not limited to domains but is applied to the whole url.

For the value used for the test, I took the value from `testFilterOnDomainName`.

Kdecherf:
The current change is expected to fail with PostgreSQL. It will help confirm if there is case sensitivity issue on the search part

Kdecherf:
Ready for review

j0k3r:
Great!

PR #2841, opened 2017-02-04
Show active list in the left menu during search

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | -
| Deprecations? | no
| Tests pass?   | -
| Documentation |  no
| Translation   | no
| Fixed tickets | -
| License       | MIT

This PR adds `active` class to the list used for the current search. The active list is retrieved from the `currentRoute` query parameter.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
@@ -33,17 +33,29 @@
             {% endblock %}
 
             {% set currentRoute = app.request.attributes.get('_route') %}
+            {% set currentRouteFromQueryParams = app.request.query.get('currentRoute') %}
 
-            <li class="bold {% if currentRoute == 'unread' or currentRoute == 'homepage' %}active{% endif %}">
+            {% set activeRoute = null %}
+            {% if currentRoute == 'all' or currentRouteFromQueryParams == 'all' %}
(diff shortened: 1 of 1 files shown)


src/Wallabag/CoreBundle/Resources/views/themes/material/layout.html.twig
```diff
@@ -33,17 +33,18 @@
             {% endblock %}
 
             {% set currentRoute = app.request.attributes.get('_route') %}
+            {% set currentRouteFromQueryParams = app.request.query.get('currentRoute') %}
```
j0k3r: Can we a centralized place where we defined which route is active? It might be less messy in the html.
Like:

```twig
{% if currentRoute == 'unread' or currentRoute == 'homepage' or currentRouteFromQueryParams == 'unread' %}
    {% set activeRoute = 'unread' %}
{% elseif currentRoute == 'starred' or currentRouteFromQueryParams == 'starred' %}
    {% set activeRoute = 'starred' %}
{% endif %}
```

Kdecherf: ack, will do

Kdecherf: done

j0k3r:
Thanks !

PR #2874, opened 2017-02-14
Add informations about Apache 2.4

kgaut:
The documentation is missing information about the correct directives to use with apache 2.4. Without the "Require All granted" apache throw a 403 error.


j0k3r:
Thanks for that!

Issue #2864, opened 2017-02-12
Performance issue with assignTagsToEntry method


nicosomb:
When we use ContentProxy::assignTagsToEntry() method, we have performance issue. 

We need to investigate on this issue, but maybe it's the usage of `findOneByLabel()` when we have many tags in database (17k on wallabag.it). 

The `label` field is a longtext one, without length. I think that it's not the better way to add index on this field. 

We already have an index on `slug` field, but we don't have the slug when we call `assignTagsToEntry` (particularly when we called it during a Pocket/Pinboard import). 

Is that a good way to "slugify" the label and find the tag by slug?

j0k3r:
Interesting!
Since the Tag slug is limited to 128 chars, maybe we can convert it to a simple `varchar(125)` and add an index on it.

nicosomb:
That's fine for me. 
FYI, the longest tag on wallabag.it contains 61 characters.

nicosomb:
Adding an index doesn't fix this bug. 
Need to investigate on it.

nicosomb:
The performance issue is in `addTag` in `Entry.php`:

```
    /**
     * @param Tag $tag
     */
    public function addTag(Tag $tag)
    {
        if ($this->tags->contains($tag)) {
            return;
        }

        // check if tag already exist but has not yet be persisted
        // it seems that the previous condition with `contains()` doesn't check that case
        foreach ($this->tags as $existingTag) {
            if ($existingTag->getLabel() === $tag->getLabel()) {
                return;
            }
        }

        $this->tags->add($tag);
        $tag->addEntry($this);
    }
```

I will investigate on this method later.

nicosomb:
FYI, I re-enabled this part of Import on wallabag.it few days ago ... and I didn't have performance issue.

j0k3r:
All right, then we can clear the milestone and see if you got the issue again later.

nicosomb:
The issue is still here on wallabag.it. 

@tcitworld do you have performance issues on Framabag.org during Pocket imports via RabbitMQ (or maybe imports are disabled on Framabag, I can't remember).

tcitworld:
They are indeed disabled (that's how you fix things over here).

nicosomb:
:trollface: Thank you for your reply.

Kdecherf:
It could be interesting to see slow queries log

PR #2897, opened 2017-02-16
Adding tags in instapaper import

itayfo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

### DIFF
--- a/src/Wallabag/ImportBundle/Import/InstapaperImport.php
+++ b/src/Wallabag/ImportBundle/Import/InstapaperImport.php
@@ -5,6 +5,8 @@ namespace Wallabag\ImportBundle\Import;
 use Wallabag\CoreBundle\Entity\Entry;
 use Wallabag\CoreBundle\Entity\Tag;
 
+ini_set("memory_limit", "-1");//Let's home god will forgive
+
 class InstapaperImport extends AbstractImport
 {
     private $filepath;


src/Wallabag/ImportBundle/Import/InstapaperImport.php
```diff
@@ -115,6 +117,22 @@ public function parseEntry(array $importedEntry)
         $entry->setUrl($importedEntry['url']);
         $entry->setTitle($importedEntry['title']);
 
+        if($importedEntry['tags']) {
+            foreach(explode(',', $importedEntry['tags']) as $tag) {
+                if(strlen($tag) === 0)continue;
+                $existingTag = $this->em
+                    ->getRepository('WallabagCoreBundle:Tag')
+                    ->findOneByLabel($tag);
```
nicosomb: There is a performance issue with this method, see https://github.com/wallabag/wallabag/issues/2864. 
We need to work on it.

nicosomb:
@itayfo: how did you manage to add multiple tags (folders) on a URL? With the Instapaper web application, it's not possible. 

Moreover, you defined `$data[4]` for tags, but here is a sample line from my own export: 
`http://www.lemonde.fr/les-decodeurs/article/2017/01/17/comment-benoit-hamon-a-rabote-son-revenu-universel-en-pleine-campagne_5064235_4355770.html,"En pleine campagne, Benoît Hamon rabote (puis remet) son « revenu universel »",,plop` where `plop` is a folder I created. So, it's the 3rd element, not the 4th. 

Finally, I have my other lines with `unread` in this column. So with your pull request, we'll have `Unread` as a tag.

src/Wallabag/ImportBundle/Import/InstapaperImport.php
```diff
@@ -115,6 +117,24 @@ public function parseEntry(array $importedEntry)
         $entry->setUrl($importedEntry['url']);
         $entry->setTitle($importedEntry['title']);
 
+        if ($importedEntry['tags']) {
```
j0k3r: Might be better to re-use `assignTagsToEntry`

```php
        if (!empty($data['tags'])) {
            $this->contentProxy->assignTagsToEntry(
                $entry,
                $data['tags'],
                $this->em->getUnitOfWork()->getScheduledEntityInsertions()
            );
        }
```

itayfo:
@nicosomb 

Apologizes. It is not a valid pull request. 
I wanted to quickly support mass import of URLS into Wallabag.
I believe The proper way to do is create some 'custom' importer, which I did not. 

We used this code as a part of internal importer.
Info was taken from RSS feeds.

nicosomb:
Can I close this PR?

j0k3r:
But adding tag from instapaper import isn't a bad idea, nope? :thinking:

nicosomb:
Yeah, sorry, I replied too quickly.

itayfo:
Yes, you can. 

But I do think we need to support some custom import type, that could importing tags as well.

j0k3r:
Since you removed your repository I can no longer work on what you already did.
So I'll close that PR in favor of a new one I just created https://github.com/wallabag/wallabag/pull/3168

Thanks for the idea!

Issue #2906, opened 2017-02-17
Api : separate function for get list of ID of deleted articles


Rurik19:
Something like that : /api/deleted{format}?since=date

tcitworld:
The table in which the deleted articles id are stored should be cleaned from time to time. Maybe with a cron ?

di72nn:
It is important to make a clear distinction between "nothing is deleted" and "data can't be provided" (e.g. `since` is too far).

tcitworld:
Also, instead of getting only the deleted articles, why not all articles that were changed (title edited, fav, archive, tag, ...) ?

Rurik19:
/api/history?since=date

di72nn:
So basically `api/entries` but with deleted articles? That would work too.

nicosomb:
As I said here https://github.com/wallabag/wallabag/pull/2918#issuecomment-289868189, I don't know if it's a good thing to store the deletion history:  
I added a relation between Change and Entry, so each time an entry is deleted, the changes will be deleted too. And if we store only the entry ID, we have a problem with SQLite :trollface: because SQLite reuses old ID 🎉 

So I want your opinion: **are your really sure that we need to store this information?**

What is the main purpose of this feature? 
How will you implement that in third apps? ping @Rurik19 @jlnostr @di72nn @bourvill

di72nn:
We need this to detect that some of the locally stored articles were removed on the server. Otherwise we have to **iterate over all articles** and check existence of each one (limited batching is possible).

**It is a huge resource drain:**
Clients have to make a lot of requests (especially critical for mobile clients).
Server has to check existence of a huge number of articles. **The number of requests grows with the number of clients.**
One of the worst parts is that this has to be done every time, even if nothing was removed.

Having an API like that would allow to do this in a single and simple request.

Anyway, weren't you going to drop SQLite support?

bourvill:
retrieve only changed  (with new?) article seem be good.

We can use for retrieve only missing data. On mobile app it's cool to retrieve only small part of data and store change locally for offline mode for exemple (currently not available in iOs app) 

But we need a strong idenfier for an article.

Generate an UUID to avoid problem with sqlite ? But with this we need to generate for all existing article

nicosomb:
> Anyway, weren't you going to drop SQLite support?

It's planned, I know, but it's complicated. We can't drop SQLite easily: I think we can only depreciate it for 2.3 release. 

[...]

PR #2918, opened 2017-02-20
Provide changes API endpoint

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2906
| License       | MIT

Start work for the changes API

### DIFF
--- a/app/DoctrineMigrations/Version20170328185535.php
+++ b/app/DoctrineMigrations/Version20170328185535.php
@@ -1,0 +1,99 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\SchemaException;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Doctrine\DBAL\Migrations\SkipMigrationException;
+
+/**
+ * Creates the Change table.
+ */
+class Version20170328185535 extends AbstractMigration implements ContainerAwareInterface
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    private function getTable($tableName)
+    {
+        return $this->container->getParameter('database_table_prefix').$tableName;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        try {
+            $schema->getTable($this->getTable('change'));
+        } catch (SchemaException $e) {
+            // The Change table doesn't exist, we need to create it
+            if (10 == $e->getCode()) {
+                if ($this->connection->getDatabasePlatform()->getName() == 'sqlite') {
+                    $this->addSql('CREATE TABLE '.$this->getTable('change').' (id INTEGER NOT NULL, entry_id INTEGER DEFAULT NULL, type INTEGER NOT NULL, created_at DATETIME NOT NULL, PRIMARY KEY(id), CONSTRAINT FK_133B9D0FBA364942 FOREIGN KEY (entry_id) REFERENCES '.$this->getTable('entry').' (id) NOT DEFERRABLE INITIALLY IMMEDIATE)');
+
+                    return true;
+                }
+
+                $changeTable = $schema->createTable($this->getTable('change'));
+                $changeTable->addColumn(
+                    'id',
+                    'integer',
+                    ['autoincrement' => true]
+                );
+                $changeTable->addColumn(
+                    'type',
+                    'integer',
+                    ['notnull' => false]
+                );
+                $changeTable->addColumn(
+                    'created_at',
+                    'datetime',
+                    ['notnull' => false]
+                );
+                $changeTable->addColumn(
+                    'entry_id',
+                    'integer',
+                    ['notnull' => false]
+                );
+
+                $changeTable->setPrimaryKey(['id']);
+
+                $changeTable->addForeignKeyConstraint(
+                    $this->getTable('entry'),
+                    ['entry_id'],
+                    ['id'],
+                    ['onDelete' => 'CASCADE'],
+                    'IDX_change_entry'
+                );
+
+                return true;
+            }
+        }
+
+        throw new SkipMigrationException('It seems that you already played this migration.');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        try {
+            $changeTable = $schema->getTable($this->getTable('change'));
+            $schema->dropTable($changeTable->getName());
+        } catch (SchemaException $e) {
+            throw new SkipMigrationException('It seems that you already played this migration.');
+        }
+    }
+}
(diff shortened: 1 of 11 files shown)


j0k3r:
I targetted the 2.3 endpoint (since you set the milestone to 2.3.0).
Could you then rebase your banche?

src/Wallabag/CoreBundle/Event/Subscriber/ChangesSubscriber.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    /** @var LoggerInterface $logger */
+    private $logger;
+
+    /** @var EntityManager $em */
+    private $em;
+
+    public function __construct(EntityManager $em, LoggerInterface $logger)
+    {
+        die;
```
j0k3r: :trollface:

src/Wallabag/CoreBundle/Event/EntryUpdatedEvent.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Wallabag\CoreBundle\Event;
+
+use Symfony\Component\EventDispatcher\Event;
+use Wallabag\CoreBundle\Entity\Entry;
+
+/**
+ * This event is fired as soon as an entry was saved.
```
j0k3r: Comment needs to be updated

src/Wallabag/CoreBundle/Event/EntryTaggedEvent.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Wallabag\CoreBundle\Event;
+
+use Symfony\Component\EventDispatcher\Event;
+use Wallabag\CoreBundle\Entity\Entry;
+use Wallabag\CoreBundle\Entity\Tag;
+
+/**
+ * This event is fired as soon as an entry was saved.
```
j0k3r: Comment needs to be updated

src/Wallabag/CoreBundle/Event/Subscriber/ChangesSubscriber.php
```diff
@@ -0,0 +1,72 @@
...
+        $this->em->persist($change);
+        $this->em->flush();
+
+        $this->logger->debug('saved updated entry '.$event->getEntry()->getId().' event ');
+    }
+
+    /**
+     * @param EntryDeletedEvent $event
+     */
+    public function onEntryRemoved(EntryDeletedEvent $event)
```
j0k3r: Should be `onEntryDeleted`

src/Wallabag/CoreBundle/Entity/Change.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace Wallabag\CoreBundle\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * Change.
```
j0k3r: You should add a little description on what it does

src/Wallabag/CoreBundle/Event/Subscriber/ChangesSubscriber.php
```diff
@@ -0,0 +1,72 @@
...
+
+        $this->logger->debug('saved removed entry '.$event->getEntry()->getId().' event ');
+    }
+
+    /**
+     * @param EntryTaggedEvent $event
+     */
+    public function onEntryTagged(EntryTaggedEvent $event)
+    {
+        $change = new Change(Change::CHANGED_TAG_TYPE, $event->getEntry()->getId());
```
j0k3r: In all that class, `Change` got the entry id when it should take the entry object instead.

src/Wallabag/CoreBundle/Entity/Change.php
```diff
@@ -0,0 +1,109 @@
...
+        return $this->timestamp;
+    }
+
+    /**
+     * @param DateTime $timestamp
+     */
+    public function setTimestamp(DateTime $timestamp)
+    {
+        $this->timestamp = $timestamp;
+    }
```
j0k3r: I think you can remove all the setters.
All values are defined when you create the object and shouldn't be modified after.

src/Wallabag/CoreBundle/Entity/Change.php
```diff
@@ -0,0 +1,109 @@
...
+     * @ORM\ManyToOne(targetEntity="Wallabag\CoreBundle\Entity\Entry", inversedBy="changes")
+     */
+    private $entry;
+
+    /**
+     * @var \DateTime
+     *
+     * @ORM\Column(name="timestamp", type="datetime")
+     */
+    private $timestamp;
```
j0k3r: Maybe better to name it `created_at`

nicosomb:
@tcitworld why did you create an event for article deletion?

tcitworld:
To listen to it.

nicosomb:
I added a relation between Change and Entry, so each time an entry is deleted, the changes are deleted too. 
And if we store only the entry ID (as you did), we have a problem with SQLite :trollface: because SQLite reuses old ID 🎉

app/DoctrineMigrations/Version20170328185535.php
```diff
@@ -0,0 +1,88 @@
...
+                );
+                $changeTable->addColumn(
+                    'entry_id',
+                    'integer',
+                    ['notnull' => false]
+                );
+
+                $changeTable->setPrimaryKey(['id']);
+
+                $changeTable->addForeignKeyConstraint(
```
nicosomb: I don't understand why the foreign key is not created here.

src/Wallabag/CoreBundle/Event/EntryTaggedEvent.php
```diff
@@ -0,0 +1,36 @@
...
+    /** @var Entry */
+    protected $entry;
+
+    /** @var Tag[] */
+    protected $tags;
+
+    public function __construct(Entry $entry, $tags)
+    {
+        $this->entry = $entry;
+    }
```
j0k3r: You forgot to set `$this->tags`

app/DoctrineMigrations/Version20170328185535.php
j0k3r: Any error? What SQL queries did you see in your log?

nicosomb: OK, f*ck SQLite.

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -350,9 +354,6 @@ public function deleteEntriesAction(Entry $entry)
         $em->remove($entry);
         $em->flush();
 
-        // entry deleted, dispatch event about it!
-        $this->get('event_dispatcher')->dispatch(EntryDeletedEvent::NAME, new EntryDeletedEvent($entry));
```
tcitworld: Why is this removed ?

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -481,9 +486,6 @@ public function deleteEntryAction(Request $request, Entry $entry)
             UrlGeneratorInterface::ABSOLUTE_PATH
         );
 
-        // entry deleted, dispatch event about it!
-        $this->get('event_dispatcher')->dispatch(EntryDeletedEvent::NAME, new EntryDeletedEvent($entry));
```
tcitworld: Same here.

tcitworld:
We also need events for Tags and Annotations changes.

tcitworld:
If you agree with the statement above, let's remove the `entry` field from the Change entity and use `entity` instead. @wallabag/core

nicosomb:
@tcitworld if we rename `entry` into `entity`, we need to remove the relation, right? And only store the ID, as you did? 
@tcitworld for the deletion event, see https://github.com/wallabag/wallabag/issues/2906#issuecomment-290071286

tcitworld:
I was thinking of using https://www.w3.org/TR/activitypub (and provide a similar api than the Nextcloud Activity app [provides](https://github.com/nextcloud/activity/blob/master/docs/endpoint-v2.md)).

tcitworld:
Sadly, I don't think this will be in 2.3.

Issue #2896, opened 2017-02-16
Api exists should return id if article exists


Rurik19:
Api exists should return id if article exists. It would help me to work with cached articles. For now I use tab.url as cache key.

tcitworld:
Was merged in 2.3.

PR #2919, opened 2017-02-20
Api exists returns

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2896
| License       | MIT

If the entry exists, `/api/exists` will now return it's id, otherwise still `false`.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -41,7 +41,7 @@ class EntryRestController extends WallabagRestController
                     ->getRepository('WallabagCoreBundle:Entry')
                     ->findByUrlAndUserId($url, $this->getUser()->getId());
 
-                $results[$url] = false === $res ? false : true;
+                $results[$url] = $res instanceof Entry ? $res->getId() : false;
             }
 
             $json = $this->get('serializer')->serialize($results, 'json');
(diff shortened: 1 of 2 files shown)


tcitworld:
Maybe rebase from 2.3, not master.

nicosomb:
You're right, I think we need to tag this PR for 2.3 because we change the response format.

nicosomb:
About the documentation, it's written: `* @return JsonResponse`. Is there a way to give more details about what the method returns?

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -41,7 +41,7 @@ public function getEntriesExistsAction(Request $request)
                     ->getRepository('WallabagCoreBundle:Entry')
                     ->findByUrlAndUserId($url, $this->getUser()->getId());
 
-                $results[$url] = false === $res ? false : true;
+                $results[$url] = ($res instanceof Entry) ? $res->getId() : false;
```
j0k3r: You don't need parenthesis here (same at line 63)

tcitworld:
Done.

nicosomb:
@j0k3r do you have information about "About the documentation, it's written: * @return JsonResponse. Is there a way to give more details about what the method returns?" ?

j0k3r:
Well, we can add information about the return JSON (like keys inside it) but I don't know what can we do more..

di72nn:
Any chance you can keep backwards compatibility? That can become a PITA when clients start to crash (I'm pretty sure not everyone update clients on time).

There is more to it: the return type is now mixed. So it is also a PITA for implementing decent binding.

Here are two different suggestions:
 * Create a different method. This one is called **`exists`**, you can add something like `id`, `getid` or `id-by-url`, you get the idea.
 * Add a parameter like `return-id` or simply `id`. By default the method would return `boolean`, but newer clients can pass `return-id=true` and get they want. In this case the return type should depend on the parameter and be `int`/`null` (not `int`/`false`) if the parameter is present.

PR #2960, opened 2017-03-03
Added migration to create index on (user_id, url) on Entry table

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | N/A
| License       | MIT

Added index on [user_id, url] because check on api/exists took ages with wallabagger (ping @Simounet & @Rurik19).
With this issue, I also saw this one: https://github.com/wallabag/wallabagger/issues/63

### DIFF
--- a/app/DoctrineMigrations/Version20170303183149.php
+++ b/app/DoctrineMigrations/Version20170303183149.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Added indexes on wallabag_entry.url and wallabag_entry.user_id
+ */
+class Version20170303183149 extends AbstractMigration implements ContainerAwareInterface
+{
+    /**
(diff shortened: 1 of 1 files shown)


nicosomb:
FYI, here is the wallabag.it charge. 

I added the index at 6pm. 

capture d ecran 2017-03-03 a 20 12 30

nicosomb:
Oh! Some indexes are missing in entities :/

nicosomb:
We can't define a length for an index with Doctrine (see https://github.com/doctrine/dbal/pull/419). 

Because URL field has no length (it's a longtext field), we need to change it and define a length for this field. 

How many characters? 255?

Other point: do you think that we need to add a new field to store a hash of the URL (it will be easier to look for a hash instead of the plain URL)?

j0k3r:
Be careful with field length & utf8mb4.
About the hash for the url, maybe we can skip for the moment and think about it when we'll implement the url checking.

nicosomb:
> Be careful with field length & utf8mb4.

could you give me more info please?

j0k3r:
Don't you remember that problem with field length and utf8mb4 the 197 length?

j0k3r:
No more needed as we are now using hash to search for url: https://github.com/wallabag/wallabag/pull/3158

PR #2968, opened 2017-03-06
remove language on html tag

chrido:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Documentation | -
| Translation   | -
| Fixed tickets | -
| License       | MIT

I use Chrome on an English/German Ubuntu. The language of my user is also configured to be English. Annoyingly the translation integrated in Chrome detects that Wallabag is French (which is not the case, it's English) and translates it then from English to German. To fix it, i removed the language from the html tag. 

I'm not aware of a potential side effect. I think another way to fix it would be to take the language of the current user and set it in the tag.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/base.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/base.html.twig
@@ -3,7 +3,7 @@
 <!--[if lte IE 7]><html class="no-js ie7 ie67 ie678" lang="en"><![endif]-->
 <!--[if IE 8]><html class="no-js ie8 ie678" lang="en"><![endif]-->
 <!--[if gt IE 8]><html class="no-js" lang="en"><![endif]-->
-<html lang="fr">
+<html>
     <head>
         {% block head %}
             <meta name="viewport" content="initial-scale=1.0">


tcitworld:
The only interesting thing to do here would be to set it to the entries language if we have on, but it can be done in another PR.

tcitworld:
> As you said, we can detect the selected language of the logged in user

If we want Chrome to suggest to translate content, the `lang` attribute should correspond to the language of the article. Anyway, I don't think it can translate non-public pages, but for published articles it may be handy.

chrido:
I think it does not make sense to translate wallabag itself since it supports already many languages.
As you said it could be useful to translate articles. The translation integrated in chrome also works for non-public articles, otherwise it would not have worked for my self hosted instance. 

Although there are ways to extract the language of an article, e.g. tokenizing into N-Grams and classifying using a pre-built model, I think it's probably enough to just let the user select the language of the article. (This could then be used in future to train a model). 

On the other side, I normally don't select articles I cannot understand for reading later, hence I don't know if it is useful?

nicosomb:
Thank you @chrido

Issue #500, opened 2014-02-25
Possibility to run script everytime an article is added


aschilling:
Hi everybody,

I really like wallabag however only in combination with my kindle. In order to implement a solution where new articles are pushed to my kindle I am trying to call a calibre recipe everytime a new article is added. 

Is there any possibility to hook a script which is called everytime a new article is added to wallabag ?

Thanks in advance

Andy

tcitworld:
@j0k3r Will our custom events help such a thing ?

j0k3r:
Yes and no.
We should more things globally.
Check that conference for example: https://www.youtube.com/watch?v=U8vBoBmN68E

j0k3r:
Sumup (from https://github.com/wallabag/wallabag/pull/2982):
- add a table called `webhook` with uuid (as PK), name, url, boolean to enable/disable and user_id
- fire each url on saving an article with the article as json in payload (should better be a POST request)
- _optional_: log each request with http response and sent payload to be able to re-play them (like Github webhook)
- _optional_: add webhooks uuid in header request
- _optional_: signature header based on a secret?

https://developer.github.com/webhooks/ is a good idea.

j0k3r:
Interesting read https://brandur.org/webhooks

nicosomb:
I started to work on this feature.

nicosomb:
I played with IFTTT webhooks. 
It works well. 

But, if there is an error on IFTTT side, I can't have the information and the IFTTT response is "ok, all was fine, the applet was triggered successfully". 😓

tctlrd:
Thanks for those who put work into this already!

+1 for a webhook feature in wallabag.
My use is to have wallabag links automatically sent to [linkding](/sissbruecker/linkding) and possibly to [paperless-ngx](/paperless-ngx/paperless-ngx) via webhook. I may have a self-hosted [n8n](/n8n-io/n8n) instance receive the webhook to customize the integration.

Interval checking the wallabag RSS feed with n8n is a way to have some of this functionality in the meantime.
See: https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.rssfeedreadtrigger

## webhooks.fyi
[webhooks.fyi](/ngrok/webhooks.fyi) is a great resource for all things webhook.

[...]

PR #2982, opened 2017-03-14
Add support for calling hook when new article is added

dschoepe:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #500 
| License       | MIT

This pull request adds preliminary support for running a hook whenever a new article is added. However, more testing and documentation is required, but this hopefully provides a reasonable starting point.


src/Wallabag/CoreBundle/Event/Subscriber/NewArticleHookSubscriber.php
```diff
@@ -13,10 +13,12 @@
 
 class NewArticleHookSubscriber implements EventSubscriberInterface
 {
+    private $client;
     private $logger;
 
     public function __construct(LoggerInterface $logger)
     {
+        $this->client = new \GuzzleHttp\Client;
```
j0k3r: Please no, inject it in your services.yml definition. So we can easily inject a mock in test.

app/DoctrineMigrations/Version20170312203635.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return $this->container->getParameter('database_table_prefix') . $tableName;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE '.$this->getTable('config').' ADD new_article_hook VARCHAR(512) DEFAULT NULL');
```
j0k3r: Could you please add a check in case the field already exist (ie: the migration was already runned)? (check other migration to see how you can do)

dschoepe:
I'm running the version from [http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar](http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar) via `php-cs-fixer fix src` and `php-cs-fixer fix app`.  For some reason, it doesn't result in any further changes. Is travis/this project perhaps using a different configuration than the standard one?

I'll also fix the other points you mentioned.

tcitworld:
@dschoepe Use `bin/php-cs-fixer`.

j0k3r:
Hello @dschoepe, are you still working on that?

Also, I don't really know if this is the right way to handle that but what if user wants multiple outgoing hooks? (in case user wants to handle multiple zap on zapier for example)

I think we should better add a new table which will gather all the outgoing hooks for each user.
What do you think @wallabag/core?

nicosomb:
I agree with @j0k3r. New table with name, value, boolean to enable/disable this hook, etc.

dschoepe:
Unfortunately I haven't really had time to keep working on this, and probably won't be able to in the near future either. I agree that supporting multiple hooks would be better though.

j0k3r:
Thanks for letting us know.
I've sumup the work to be done in the associated issue: https://github.com/wallabag/wallabag/issues/500#issuecomment-300136215

I'm going to close that PR then.
For the work done :+1:

PR #3053, opened 2017-04-24
Added API endpoint to handle a list of URL

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | N/A
| License       | MIT

You can pass an array like this one: 
```
["http://...","http://..."]"
``` 

and add / delete URL in mass.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -55,6 +55,7 @@ wallabag_core:
     list_mode: 0
     fetching_error_message: |
         wallabag can't retrieve contents for this article. Please <a href="http://doc.wallabag.org/en/master/user/errors_during_fetching.html#how-can-i-help-to-fix-that">troubleshoot this issue</a>.
+    api_limit_mass_actions: 10
 
 wallabag_user:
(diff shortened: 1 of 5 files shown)


j0k3r:
I really think we need to restrict the maximum of urls we can handle.
Otherwise the server might take took long to answer and might even time out.

tcitworld:
I'm not really for this. Deleting an URL should be made through a `DELETE` method, not a `POST` one. You should have a different method to delete a list of urls, and ditch the `action` parameter.

nicosomb:
@j0k3r is that good for you?

j0k3r:
Still needs to restrict limit on number of urls you sent

nicosomb:
@j0k3r I need help for `testLimitBulkAction` please. The Exception seems to be thrown but my test fails.

nicosomb:
Thank you @j0k3r !

Issue #3050, opened 2017-04-21
Paginate user management page


tcitworld:
If you have 13k users like on Framabag I let you imagine the rest. Also, have a search function.

nicosomb:
Hum... see #2695 😁

tcitworld:
:)

PR #3060, opened 2017-04-27
Search & paginate users

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | #3050
| License       | MIT


src/Wallabag/UserBundle/Controller/ManageController.php
```diff
@@ -19,17 +22,32 @@ class ManageController extends Controller
     /**
      * Lists all User entities.
      *
-     * @Route("/", name="user_index")
+     * @Route("/index/{page}", name="user_index")
```
j0k3r: Why not just `"/{page}"`?

tcitworld: Because we have the "/{userid}/edit" route and I thought it would be confusing

j0k3r: Hum I see. I really don't like that `/index/` in url.
Can't you use `/?page=`?

tcitworld: If you want. I don't really like these kind of urls ;)

j0k3r: :bulb: Otherwise, use `/list/` instead of `/index/`

j0k3r:
We might merge both search & list page so the url doesn't change and we only have one method.
What do you think?

Also, I don't think the label is necessary here, since the full description of the field is the placeholder

image

tcitworld:
Suits me.

nicosomb:
* I merged search and list
* I removed `label` and icon for the search form.

src/Wallabag/UserBundle/Controller/ManageController.php
```diff
@@ -169,52 +137,44 @@ private function createDeleteForm(User $user)
...
-
         $form->handleRequest($request);
 
         if ($form->isSubmitted() && $form->isValid()) {
             $this->get('logger')->info('searching users');
-            $em = $this->getDoctrine()->getManager();
 
             $searchTerm = (isset($request->get('search_user')['term']) ? $request->get('search_user')['term'] : '');
 
             $qb = $em->getRepository('WallabagUserBundle:User')->getQueryBuilderForSearch($searchTerm);
```
tcitworld: Where is `getQueryBuilderForSearch` ?

nicosomb: In `src/Wallabag/UserBundle/Repository/UserRepository.php`.

tcitworld: I know, I added it myself. But github didn't show it in the diff. o_0

Issue #3005, opened 2017-03-21
Add support for oauth scopes


Eisfreak7:
### Issue details

If #2800 requires a significant revamp of the authentication handling it might be a good time to look into oauth scopes. It would enable users to take advantage of external services which they don't fully trust.

An example use case would be a service like [ifttt](https://ifttt.com/). Such a service would need the permission to add new articles, but not permissions to read or modify my list of existing entries.
I could imagine scopes like `add`, `read`, `modify` or maybe more fine grained ones (`read:<tag>`, `add-annotations`).
For reference: https://www.brandur.org/oauth-scope

I don't actually need this feature in the moment. I just thought it would be worth bringing up, since it would add considerable value and might be easier to add now (together with #2800) rather than later.

Issue #2800, opened 2017-01-29
API shouldn't require a username and password


anarcat:
### Issue details

The API currently requires new "API client" to be created to operate, which includes a `client_id` and a `client_secret`.

But to *actually* work, the API *also* needs an `access_token` and `refresh_token`. And *those* are very short lived - one hour maximum.

This means to implement a client, you need to not only ask the user for an incomprehensible `client_id` and `client_secret`, but *also* for their username and password. Every client then needs to *store* those username and passwords indefinitely to regenerates the access tokens.

### Environment

* wallabag version: 2.1.0 API documentation
* How did you install wallabag? irrelevant.
* Last wallabag version that did not exhibit the issue (if applicable): N/A
* php version: N/A
* OS: N/A
* type of hosting (shared or dedicated): N/A
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): N/A

### Steps to reproduce/test case

N/A.

### Suggested solution

The username and password should be used only to create access tokens, which should be permanent. This would make the whole API way more efficient because then a single HTTP request would be require for API calls instead of two (assuming the worst case of less than one request per hour per client).

Alternatively, client tokens should  be user-specific. I believe there may already be an issue about this, but right now, anyone can see (*and* delete?) anyone else's tokens on the server. That seems like a Real Bad Idea.

Clients should only ask the user's username and password as a transient secret, that can be used to create client tokens on the fly, which then can be used to create access tokens.

j0k3r:
> Alternatively, client tokens should be user-specific

It's already in place in the 2.2.0 (see https://github.com/wallabag/wallabag/pull/2499)

And we need to discuss about your interesting point.

anarcat:
from what i see in #2806 - token ownership is not actually checked?

so there would still be work needed there.

jlnostr:
I agree. Of course, you can refresh the tokens after they expired but in the case something fails after submitting the request, the new tokens are lost forever and the user needs to be asked for their username and password again. The suggested solution by @anarcat seems good to me.

Otherwise issues like https://github.com/wallabag/windows-app/issues/32 can happen at any time.

jlnostr:
[...]

PR #3065, opened 2017-04-29
Register through API

tcitworld:
Signed-off-by: Thomas Citharel <tcit@tcit.fr>

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #3056 
| License       | MIT

PR #3068, opened 2017-04-30
Changes to API Auth

tcitworld:
Therefore, username and password are no longer needed

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| Fixed tickets | #2800 #3005
| License       | MIT

I'll like to put in more things in this PR, but feel free to review and test.

- [x] Allow admin to create global clients (not user related) for automatic login for wallabag.it / framabag users in apps. This will need the `/oauth/v2/auth` route and the `authorization_code` `grant_type`.
- [x] Make `redirect_uri` not required on simple client creation.
- [x] Add back `password` `grant_type` but depreciate it.
- [ ] Tests
- [x] Translations
- [x] Better authorize page
- [ ] Database migration

Ref. https://github.com/wallabag/wallabag/issues/2800#issuecomment-286679410


tcitworld:
### User specific clients
Users can create user-specific clients, which gives them a `client_id` and `client_password` (like before), which any can app use with the `client_credentials` `grant_type` to get directly an `access_token`, **without providing username and password like before**. Callback URIs are no longer required.

### Global clients
Instances admin can create global clients accessible for all of the instance users, with a name, one or multiple callback URIs, a description (**new!**) and a picture (**new!**). Then they must contact the app maintainers to add their instance with the corresponding client_id.
There's now two methods to connect, depending if the app is server-side (or can obfuscate client secrets) or if it's client-side.
#### Server-side
Applications first call `https://wallabag.tld/oauth/v2/auth` with the following parameters : 
  * `client_id`: the one the instance administrator gave
  * `redirect_uri`: the uri (schemes allowed !) to redirect after authorization is made, read below)
  * `response_type`: `code`
  * `scope`: There is `read`, `write` and `user` (ref #3065)

Then, wallabag promps a page (after login if needed) asking the user to allow the app and then redirects to the `redirect_uri` with a code as parameter.
The app now calls `https://wallabag.tld/oauth/v2/auth` with the following parameters : 
  * `code`: the code you just got
  * `client_id`: the client id, same as before
  * `redirect_uri`: the final redirect uri
  * `grant_type`: `authorization_code`

The `redirect_uri` is finally called with the `access_token` (and `refresh_token` and others like before...) as parameter.

#### Client-side
Applications first call `https://wallabag.tld/oauth/v2/auth` with the following parameters : 
  * `client_id`: the one the instance administrator gave
  * `redirect_uri`: the uri (schemes allowed !) to redirect after authorization is made, read below)
  * `response_type`: `token`
  * `scope`: There is `read`, `write` and `user` (ref #3065)

The `redirect_uri` is then called with the `access_token` (and `refresh_token` and others like before...) as parameter.

tcitworld:
Ping @wallabag/contributors @wallabag/core ^

src/Wallabag/ApiBundle/Controller/DeveloperController.php
```diff
@@ -20,9 +22,62 @@ class DeveloperController extends Controller
     public function indexAction()
     {
         $clients = $this->getDoctrine()->getRepository('WallabagApiBundle:Client')->findByUser($this->getUser()->getId());
+        $global_clients = $this->getDoctrine()->getRepository('WallabagApiBundle:Client')->findByUser(null);
```
j0k3r: camelCase please ➡️ `$globalClients`

src/Wallabag/ApiBundle/Controller/DeveloperController.php
```diff
@@ -75,10 +130,16 @@ public function createClientAction(Request $request)
      */
     public function deleteClientAction(Client $client)
     {
-        if (null === $this->getUser() || $client->getUser()->getId() != $this->getUser()->getId()) {
+        if (($client->getUser() !== null) && (null === $this->getUser() || $client->getUser()->getId() != $this->getUser()->getId())) {
             throw $this->createAccessDeniedException('You can not access this client.');
         }
 
+        if ($img = $client->getImage()) {
+            if ($file = $this->getParameter('wallabag_api.applications_icon_path') . '/' . $img) {
```
j0k3r: You can merge these both condition

matteocoder:
@tcitworld should the whole comment be translated or should I leave something out (like the content inside the parentheses)?

tcitworld:
@matteocoder Which comment ? I haven't started the translations.

src/Wallabag/CoreBundle/Resources/translations/messages.fr.yml
```diff
@@ -445,13 +445,17 @@ developer:
     clients:
         title: "Clients"
         create_new: "Créer un nouveau client"
+        create_new_global: Créer un nouveau client global
```
nicosomb: Maybe you can add quotes for each translation.

app/Resources/FOSOAuthServerBundle/views/Authorize/authorize_content.html.twig
```diff
@@ -0,0 +1,52 @@
+<div class="row">
```
nicosomb: Don't forget translations for this file.

matteocoder:
@tcitworld sorry, I got a little confused. You referenced this PR in another one, so I though that some other text needed to be translated.

src/Wallabag/ApiBundle/Controller/DeveloperController.php
```diff
@@ -20,9 +22,62 @@ class DeveloperController extends Controller
...
+            $client->setAllowedGrantTypes(['token', 'authorization_code', 'refresh_token']);
+            $em->persist($client);
+            $em->flush();
+
+            $this->get('session')->getFlashBag()->add(
+                'notice',
+                $this->get('translator')->trans('flashes.developer.notice.client_created', ['%name%' => $client->getName()])
+            );
+
+            return $this->render('@WallabagCore/themes/common/Developer/client_parameters.html.twig', [
```
nicosomb: Maybe you can display the uploaded picture when we just have created the client.

jlnostr:
@tcitworld Call me stupid, but I don't understand the new behavior as there are several questions open.

1. How should I call the given URL? Inside a webbrowser visible for the user or hidden?
2. How is the redirect URI called with the tokens? Is there a scheme for it? How will it connect with the given redirect URI?

It would be nice to clarify the auth behavior and a recommendation which (server or client) behavior an app should use.

jlnostr:
After this PR the QR codes for setting up a client should also be updated.

tcitworld:
> How should I call the given URL? Inside a webbrowser visible for the user or hidden?

Yes, the user must interact to login then allow your app.

> How is the redirect URI called with the tokens?

The redirect URI is called with tokens as `GET` parameters.

Is there a scheme for it? How will it connect with the given redirect URI?

I don't get these points.

jlnostr:
Forget my last point. Assuming I have the following redirect URI:

    wallabag://login_success

How is the final result going to look like? Like this?

    wallabag://login_success?access_token=abcd&refresh_token=1234

tcitworld:
That's right.

Simounet:
I didn't do this part on wallabagger but if I get it well:
- there is no more `oauth/v2/token` but `oauth/v2/auth`
- the user need to interact directly with wallabag (on a popup for example)
- we need to get and store the result of the previous page mentioned

tcitworld:
Okay so I've changed it a bit. I found that admins registering themselves the apps was a pain in the ass so now apps can register themselves by sending a POST request at `/api/apps` with the following parameters : 

* `client_name`: **REQUIRED** the client name, typically your app name, that will be shown to the user
* `redirect_uri`: **REQUIRED** the redirect uris used through the auth process
* `logo_uri`: An URL for your logo that will be displayed on the allow page
* `description`: A short description for your app that will be displayed at the user
* `app_uri`: An URL for your app, it's homepage or help section...

In other changes, the access_token delivered how has a lifetime of **60 days** and refresh_tokens are no longer expired. You also need to pass a state (random string) when requiring permission from the user on `/oauth/v2/auth` and check it afterwards.

Docs incoming. Translations now welcomed.

tcitworld:
> there is no more oauth/v2/token but oauth/v2/auth

No. `/oauth/v2/auth` is used to allow the user to accept the app to use his account. `/oauth/v2/token` is only used in the Server-side to exchange a code for an `access_token`. Otherwise, `/oauth/v2/auth` returns directly the `access_token` in the redirect uri callback.
> the user need to interact directly with wallabag (on a popup for example)

Nope, you need to register a callback to this kind of URIs : https://developer.chrome.com/apps/identity#method-launchWebAuthFlow

> we need to get and store the result of the previous page mentioned

Hmm ?

tcitworld:
Also, added a « Featured apps page » (added wallabagger afterwards).
wallabag_prez

jlnostr:
> now apps can register themselves by sending a POST request
And this doesn't need any kind of auth? If so, then we should worry about a lot of spam clients because everyone can create a dozen of useless clients.

Another question: With the new `oauth/v2/auth` way, refreshing the tokens is only necessary once a year, right? And it still uses the same "old" API endpoints? Or is refreshing not longer needed?

tcitworld:
> And this doesn't need any kind of auth? If so, then we should worry about a lot of spam clients because everyone can create a dozen of useless clients.

Yes, but I plan to have a task/command (called by a cron for instance) to remove all useless clients that didn't get an approval from the user.

> Another question: With the new oauth/v2/auth way, refreshing the tokens is only necessary once a year, right? And it still uses the same "old" API endpoints? Or is refreshing not longer needed?

That's what I was hoping for but the library we use makes refreshing required. So refreshing the tokens is needed once a year. The `/oauth/v2/token` route stays the same.

di72nn:
>> And this doesn't need any kind of auth? If so, then we should worry about a lot of spam clients because everyone can create a dozen of useless clients.

> Yes, but I plan to have a task/command (called by a cron for instance) to remove all useless clients that didn't get an approval from the user.

That's not acceptable. I'm yet to read through the changes to suggest anything.

tcitworld:
> That's not acceptable

Please develop.

di72nn:
Allowing to perform not read-only actions on a server without any authentication is an easy way to introduce DoS or other vulnerabilities.

> a task/command (called by a cron for instance) to remove all useless clients that didn't get an approval from the user

That doesn't sound like a viable solution.


IMO, user-specific apps should be added with at least user access level, global apps should be added with admin user access level. A flow like `wallabag promps a page (after login if needed) asking the user to allow the app` should be fine I guess.

di72nn:
Also I don't see the point in overly increasing tokens' lifetime considering improved auth mechanism.

tcitworld:
The idea is for apps to login users with a good user experience. It shouldn't be needed to create client credentials on the server or even scan a QR code before using them.
On big instances the admins creating the global apps could be fine, but most of our users aren't here anyway.

Thoughts @wallabag/core ?

> Also I don't see the point in overly increasing tokens' lifetime considering improved auth mechanism.
 
That's not a improvement, just a reasonable adjustement for web apps which can't get a `refresh_token` since they use the implicit flow.

di72nn:
> The idea is for apps to login users with a good user experience. It shouldn't be needed to create client credentials on the server or even scan a QR code before using them.

 * Leave the same initial call to `/api/apps`.
 * Require a user to login (if he is not logged in already) before adding a new app.
  Isn't the user supposed to login on the next step anyway (when he authorizes a device for the app)?
 * If the app doesn't exist ask user to create it (simple OK/cancel form). If the app is already registered, return existing app info.
 * Proceed with auth.

In this scenario the `/api/apps` method allows to get credentials for existing or new apps to avoid hassle with QR code scanning (or whatever) and avoids creating duplicates of the same app.

tcitworld:
Doable.

tcitworld:
If I understand correctly.

* the app calls the `api/apps` route with the following parameters : 
   * `client_name`: **REQUIRED**
   * `redirect_uris`**REQUIRED**
   * `logo_uri`
   * `description`
   * `app_uri`
   * `uri_redirect_after_creation`: **NEW** - **REQUIRED** - Callback to transmit informations. *Please suggest a better parameter name*
* The user login if needed
* If the app name already exists, wallabag redirects to callback `uri_redirect_after_creation` with it's parameters
* wallabag asks the user if he wants to create the app
   * ok : callback to `uri_redirect_after_creation` with the new parameters
   * no : callback to `uri_redirect_after_creation` with an error code

However, I'm not sure the `client_name` is enough to identify an app. For instance, a fake app could give the same name and get the credentials of the previous app.
Maybe the best would be to offer the existing app or offer to create a new one, no ? @di72nn

di72nn:
> For instance, a fake app could give the same name and get the credentials of the previous app.

The credentials don't give any access on their own, they are only used to bind clients to specific apps, right?

> Maybe the best would be to offer the existing app or offer to create a new one, no ?

Maybe add a screen that shows details about an existing app (all the fancy stuff: logo, description) with a big "Continue" button, and a subtle "Create new app instead" link/button?

tcitworld:
> The credentials don't give any access on their own, they are only used to bind clients to specific apps, right?

No, but they could ask the user for access, and since they look the same as the real apps, users could give them access.

> Maybe add a screen that shows details about an existing app (all the fancy stuff: logo, description) with a big "Continue" button, and a subtle "Create new app instead" link/button?

Yup.

> Also, I'm not sure what's the purpose of this parameter:

>> redirect_urisREQUIRED

These are an array of the URIs used during the auth process, as the standard describes it.

tcitworld:
I'll let you guys decide if you want this in 2.4 or not

PR #3106, opened 2017-05-11
Allow other fields to be send using API

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/1634
| License       | MIT

Creating entry using the API can now have these new fields:
- content
- language
- preview_picture
- published_at

Re-use the ContentProxy to be able to do the same using the web UI (in the future).
htmLawed is used to clean stuff from content, I hope it’ll be enough to avoid security breach.

Lower content validation when we want to update an entry with content already defined. Before, language & content_type were required. If there weren’t provided, we re-fetched the content using graby. I think these fields aren’t required for an entry to be created. So I removed them.
Which means some import from the v1 export won’t be re-fetched since they provide content, url & title.

Also, remove liberation link from Readability import to avoid overlaping import (from wallabag v1, which had the same link)

- [x] add support for author
- [x] better support for date?
- [x] add many test with crazy html (see https://github.com/j0k3r/graby/pull/105)
- [x] update doc about new fields (and combination that can force parsing, like missing title)


j0k3r:
Doc is displayed like that in swagger:

image

j0k3r:
The PR is now RFR.

Keep in mind that https://github.com/j0k3r/graby/pull/105 needs to be merged first (and then the `composer.json` in that PR will need an update too).

nicosomb:
About `content`, is that all the DOM or just the article's content?

j0k3r:
Oh I got you, it's the html content, with tags etc.
But it can also be simple string without content.

Rurik19:
Will be there some sort of security checking? html it can contain some dangerous scripts

j0k3r:
@Rurik19 reading the whole PR and you'll see that there will be some check applied (in https://github.com/j0k3r/graby/pull/105) but feel free to provide more crazy html so we can improve these checks.

j0k3r:
Graby PR has been merged.
Final RFR.

Rurik19:
Will PATCH method support such functionality?

j0k3r:
Hum not yet.
Good point, I'll add it.

nicosomb:
@j0k3r there are some conflicts.

j0k3r:
@nicosomb :white_check_mark:

j0k3r:
@Rurik19 patch also got new fields.

PR #3110, opened 2017-05-12
Added logger when we match Tagging rules

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      |  no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/RuleBasedTagger.php
+++ b/src/Wallabag/CoreBundle/Helper/RuleBasedTagger.php
@@ -8,6 +8,7 @@ use Wallabag\CoreBundle\Entity\Tag;
 use Wallabag\CoreBundle\Repository\EntryRepository;
 use Wallabag\CoreBundle\Repository\TagRepository;
 use Wallabag\UserBundle\Entity\User;
+use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Helper/RuleBasedTagger.php
```diff
@@ -36,6 +38,11 @@ public function tag(Entry $entry)
                 continue;
             }
 
+            $this->logger->info('Matching rule.', [
+                'rule' => $rule->getRule(),
+                'tags' => $rule->getTags(),
```
j0k3r: Can't you put just string instead of the whole object?
Like the rule name and the tag list

nicosomb: The object is `$rule` and `Rule` and `Tags` are fields in wallabag_tagging_rule table.

j0k3r:
Could you also test that the logger was called (ie: contains records)?
See https://github.com/j0k3r/graby/blob/master/tests/GrabyFunctionalTest.php#L60

Issue #2967, opened 2017-03-06
Tiny UI wording change: rename “Download” to “Export” in article view.


brabalan:
There is a small UI inconsistency in Wallabag: when viewing a collection of articles (for instance a tag), the button at top right says “Export”. In article view, the same button (on the left) says “Download”. It might be better to say “Export” there too.

PR #3130, opened 2017-05-18
unify Download/Export wording.

X-dark:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | #2967 
| License       | MIT

Use export in article view as well.
The translation has been taken from entry.list.export_title when available.


tcitworld:
Instead of modifying only the value of the translation, let's change the key too. And let's change it too in the docs.

Also, Travis, get your shit together.

Issue #2835, opened 2017-02-03
Staying on an article view after removing a tag


foxmask:
### Issue details

When we drop a tag from the view page of an article, we are redirected to the unread articles

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.2
* How did you install wallabag? Via `git clone` or by downloading the package? git
* Last wallabag version that did not exhibit the issue (if applicable): n/a
* php version: 7
* OS: deb 8.5
* type of hosting (shared or dedicated): dedi
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): pgsql

### Steps to reproduce/test case

open an article with tags, drop one tag and see

Kdecherf:
It appears that this behavior is currently handled by _"Where do you want to be redirected after mark an article as read?"_.

@j0k3r @nicosomb Should we create an option for tag removal or just consider that we must always stay on the article?

j0k3r:
I prefer option 2.

nicosomb:
For the tag deletion, I think option 2 too.

PR #3138, opened 2017-05-21
Ignore ActionMarkAsRead when removing tag from entry

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | somewhat?
| Deprecations? | no
| Tests pass?   | -
| Documentation | no
| Translation   | no
| Fixed tickets | #2835 
| License       | MIT

Ignore ActionMarkAsRead to stay on entry when removing a tag.

### DIFF
--- a/tests/Wallabag/CoreBundle/Controller/TagControllerTest.php
+++ b/tests/Wallabag/CoreBundle/Controller/TagControllerTest.php
@@ -134,6 +134,12 @@ class TagControllerTest extends WallabagCoreTestCase
         $this->assertEquals(302, $client->getResponse()->getStatusCode());
         $this->assertEquals($entryUri, $client->getResponse()->getTargetUrl());
 
+        // re-retrieve the entry to be sure to get fresh data from database (mostly for tags)
+        $entry = $client->getContainer()
+            ->get('doctrine.orm.entity_manager')
+            ->getRepository('WallabagCoreBundle:Entry')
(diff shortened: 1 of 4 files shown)


Kdecherf:
yep, i'll work on the test and php-cs-fixer next week

Kdecherf:
I've added two new tests and updated one existing (but this last one is failing). I also fixed files with php-cs-fixer.

tests/Wallabag/CoreBundle/Controller/TagControllerTest.php
```diff
@@ -126,9 +126,13 @@ public function testRemoveTagFromEntry()
             ->getRepository('WallabagCoreBundle:Tag')
             ->findOneByEntryAndTagLabel($entry, $this->tagName);
 
+        // We make a first request to set an history and test redirection after tag deletion
+        $client->request('GET', '/view/'.$entry->getId());
```
Kdecherf: @j0k3r @nicosomb this new request is making the next `assertNotContains` call to fail.

It seems that `$entry` is not correctly updated when accessed during assertion. Do you have any idea?

j0k3r: This should be fixed.
I don't know why it worked before...
I just re-retrieved the entry to be sure to then fetch tags correctly, and it works.

Issue #2502, opened 2016-10-26
Tags - do not differ between upper / lower case


benbeton:
This is a feature request

tags should not be different when written in upper / lower case.

Right now, I see 2 different tags when writing them in upper or lower case.

regards

nicosomb:
Hello @benbeton! 

It's not a feature request, it's a bug in fact.

benbeton:
great, hope that will be fixed in one of the upcoming version.

nicosomb:
We can close this issue thanks to @Kdecherf 🎉

PR #3139, opened 2017-05-21
Ignore tag's case

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | somewhat?
| Deprecations? | no
| Tests pass?   | -
| Documentation | no
| Translation   | no
| Fixed tickets | #2502 
| License       | MIT

Store tag labels in lowercase in order to prevent case-sensitive duplicates.

Non-lowercase tags already existing before this PR should still work as usual but any new tag will be saved in lowercase.

### DIFF
--- a/app/DoctrineMigrations/Version20170719231144.php
+++ b/app/DoctrineMigrations/Version20170719231144.php
@@ -1,0 +1,103 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Changed tags to lowercase
+ */
+class Version20170719231144 extends AbstractMigration implements ContainerAwareInterface
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    private function getTable($tableName)
+    {
+        return $this->container->getParameter('database_table_prefix').$tableName;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->skipIf($this->connection->getDatabasePlatform()->getName() == 'sqlite', 'Migration can only be executed safely on \'mysql\' or \'postgresql\'.');
+        
+        // Find tags which need to be merged
+        $dupTags = $this->connection->query("
+            SELECT LOWER(label)
+            FROM   ".$this->getTable('tag')."
+            GROUP BY LOWER(label)
+            HAVING COUNT(*) > 1"
+        );
+        $dupTags->execute();
+
+        foreach ($dupTags->fetchAll() as $duplicates) {
+            $label = $duplicates['LOWER(label)'];
+
+            // Retrieve all duplicate tags for a given tag
+            $tags = $this->connection->query("
+                SELECT id
+                FROM   ".$this->getTable('tag')."
+                WHERE  LOWER(label) = '".$label."'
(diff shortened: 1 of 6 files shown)


j0k3r:
Is test on MySQL are green locally?

Kdecherf:
@j0k3r will check next week

Kdecherf:
Well... After checking the failing test I think I'm hitting a case sensitivity issue with databases. I will rework the PR.

Kdecherf:
I updated the test for WallabagV1Import. Please review carefully @j0k3r @nicosomb @tcitworld 

Reminder: this PR removes the ability to have upper chars in tag labels.

nicosomb:
Do we need a migration to update current tags?

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -146,4 +146,52 @@ private function validateContent(array $content)
...
+
+    /**
+     * Assign some tags to an entry.
+     *
+     * @param Entry        $entry
+     * @param array|string $tags          An array of tag or a string coma separated of tag
+     * @param array        $entitiesReady Entities from the EntityManager which are persisted but not yet flushed
+     *                                    It is mostly to fix duplicate tag on import @see http://stackoverflow.com/a/7879164/569101
+     */
+    public function assignTagsToEntry(Entry $entry, $tags, array $entitiesReady = [])
```
j0k3r: Hum, in 2.3 `assignTagsToEntry` has moved to `Wallabag\CoreBundle\Helper\TagsAssigner`.
Please update the correct one 🙂

Kdecherf: Well I blindly rebased my work without seeing that the whole method was missing...

Kdecherf:
I need to add a specific test case for checking backward compatibility.

j0k3r:
Go ahead 🙂

Kdecherf:
I broke something really hard while trying to fix some tests. It's way more
complicated than I initially thought when dealing with same tags with
different cases. The better way will be to migrate all previous tags in
lowercase. I'll handle that on the next week.

Sorry for the delay :(

Sent from my phone

Kdecherf:
I am going to resume work on this PR this evening.

Kdecherf:
The migration of tags is mostly done but I found another issue regarding unicode characters: PHP function `strtolower()` and SQL function `LOWER()` don't handle them. In this case wallabag can't merge `héllo` and `hÉllo` tags together.

Should we improve the support of unicode chars on the app side or should we just convert label like slug instead?

Kdecherf:
Well... Correction: only SQLite has issues with LOWER-ing unicode characters.

Kdecherf:
Please check the message of the commit introducing migration for detail. Note also that SQLite is unsupported by this migration.

Extensive test is required for migrating existing databases on MySQL and PostgreSQL.

app/DoctrineMigrations/Version20170719231144.php
```diff
@@ -0,0 +1,103 @@
...
+            if (count($ids) > 0) {
+                // Merge tags
+                $this->addSql("
+                    UPDATE '".$this->getTable('entry_tag')."'
+                    SET    tag_id = ".$newId."
+                    WHERE  tag_id IN (".implode(',', $ids).")"
+                );
+
+                // Delete unused tags
+                $this->addSql("
```
j0k3r: What if the tag is associated to some entries?
We should handle the relation table `entry_tag` too 😕

Kdecherf: This is the goal of the SQL block just above it

j0k3r: :rofl: true.

j0k3r:
Could you run a php-cs-fixer? And I think we'll be good to go :slightly_smiling_face:

j0k3r:
👏

Issue #3157, opened 2017-05-28
Consider sending hash of URL to check if pages already saved


greigdp:
Just a suggestion, rather than a bug. The current implementation appears to send the full URL of pages visited to the server (https://github.com/wallabag/wallabagger/blob/master/wallabagger/js/wallabag-api.js#L218). There may be some risks as a result of this (for example, URLs containing auth tokens or other secrets etc, getting sent to the server).

If you run your own server, this should not be a massive risk (and there's always the option to disable it), although it would appear that the sending of the URL of each visited page could be avoided. One way would be to send the (e.g. SHA256/512) hash of the page URL (and store this on the server side to do a look-up). If there was any "fancy" processing of the URL to remove query parameters etc, this would need done at both client and server side to get the hash to match.

This would appear to preserve privacy of pages visited which are not added to wallabag (and ensuring minimum information is disclosed to the server which could be held in access logs), while preserving the UX benefits of allowing the user to tell if the page is already saved (and giving the green icon).

tcitworld:
I totally agree on this.

cc @wallabag/contributors

tcitworld:
Easy enough.

greigdp:
Just thinking about this and looking over the patch (a long time since I used PHP though), I wonder if it's also worth incorporating a per-user salt to the hash. I am not familiar enough with the wallabag data model to know if such a salt exists (not the password salt). The idea here would be to attempt to mask data beyond passive traffic analysis (if users ignore the advice to use HTTPS). It would complicate the system though, as each logged in session would need to know this salt, and it would need to be stored in the database for new clients.

Maybe this is overkill? The idea would be to use HMAC_SHA512(salt, url), or something similar. That way, there's no way to generate a lookup table of hashes to URLs at the server-side, or to correlate visited URLs between users, since user 1 and user 2 will both generate a different hash for the same URL.

Rurik19:
I am disagree. The hashe computing is quite heavy operation. Is it possible to implement in javascript? How lond will user wait after he clicked on the icon in browser? Will this work at all if computer doesn't have enought memory?

di72nn:
// This should be a wallabagger issue (issue for wallabag is to create new API to implement it with hashes or something).

Sending every visited URL to server (it is what wallabagger does, right?) is very bad for a couple of reasons.
[...]

PR #3158, opened 2017-05-28
Hash exist url

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #3157 
| License       | MIT

### DIFF
--- a/app/DoctrineMigrations/Version20190401105353.php
+++ b/app/DoctrineMigrations/Version20190401105353.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Wallabag\CoreBundle\Doctrine\WallabagMigration;
+
+/**
+ * Add hashed_url in entry.
+ */
+class Version20190401105353 extends WallabagMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $entryTable = $schema->getTable($this->getTable('entry'));
+
+        $this->skipIf($entryTable->hasColumn('hashed_url'), 'It seems that you already played this migration.');
+
+        $entryTable->addColumn('hashed_url', 'text', [
+            'length' => 32,
+            'notnull' => false,
+        ]);
+
+        // sqlite doesn't have the MD5 function by default
+        if ('sqlite' !== $this->connection->getDatabasePlatform()->getName()) {
+            $this->addSql('UPDATE ' . $this->getTable('entry') . ' SET hashed_url = MD5(url)');
+        }
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $entryTable = $schema->getTable($this->getTable('entry'));
+
+        $this->skipIf(!$entryTable->hasColumn('hashed_url'), 'It seems that you already played this migration.');
+
+        $entryTable->dropColumn('hashed_url');
+    }
+}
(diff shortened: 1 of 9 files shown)


tcitworld:
Godammit Travis.

j0k3r:
Rebase needed

nicosomb:
Just FYI, this feature will also be useful to add an index on entries. 
It's complicated to add an index on URL (because URL can have so many letters), but it will be easier with hashes.

shtrom:
Hey there,

Not sure if I'm reading this correctly, but does this mean that the client is now in charge of the hashing before hitting the URL? This seems to introduce an unnecessary coupling, where client and API now need to agree on the hash to use. It would seem more flexible to not change the API, and do the hashing entirely in the backend, just before hitting the DB.

Speaking of hash agility, I wonder if SHA-512 is not a bit overkill here. Noting that this is not a security feature, I think we could just use MD5, which would be faster. The only drawback might be the risk of increase collision, but I guess this could be handled with searching by hash, then confirming a hit by doing a full-text comparison. The hash would however not be usable as an primary key, but I don't think that would be an issue.

Better yet, with all the hashing in the backend, what algorithm to use could be a simple configuration option!

Kdecherf:
> Not sure if I'm reading this
> [wallabag/src/Wallabag/ApiBundle/Controller/EntryRestController.php](https://github.com/wallabag/wallabag/blob/642528bb15293405ccf0160f06aff342f165714f/src/Wallabag/ApiBundle/Controller/EntryRestController.php#L24-L25)
> 
> Lines 24 to 25 in [642528b](/wallabag/wallabag/commit/642528bb15293405ccf0160f06aff342f165714f)
> 
>       *          {"name"="hashedurl", "dataType"="string", "required"=true, "format"="An url", "description"="SHA512 Url to check if it exists"}, 
>       *          {"name"="hashedurls", "dataType"="string", "required"=false, "format"="An array of urls (?urls[]=http...&urls[]=http...)", "description"="SHA512 Urls (as an array) to check if it exists"} 
> 
> correctly, but does this mean that the client is now in charge of the hashing before hitting the URL? This seems to introduce an unnecessary coupling, where client and API now need to agree on the hash to use. It would seem more flexible to not change the API, and do the hashing entirely in the backend, just before hitting the DB.

The main concern in the linked issue is a privacy one. The idea is to provide a way for the client to quickly check if a given address exists in the database without leaking the actual searched address in the webservers access logs. In this case the hash must be done on the client side.

> Better yet, with all the hashing in the backend, what algorithm to use could be a simple configuration option!

The hashing algorithm may request a database change to update the length of the corresponding field in the table: we should always provide a fixed char field for better indexation.

j0k3r:
Good point @Kdecherf. I was _all-in_ with @shtrom until I re-read the privacy issue.
I also think the algorithm won't be a parameter, it'll be hardcoded to md5.

Kdecherf:
> I also think the algorithm won't be a parameter, it'll be hardcoded to md5.

Maybe we could at least use sha1?

j0k3r:
> Maybe we could at least use sha1?

Okay

shtrom:
> Good point @Kdecherf. I was _all-in_ with @shtrom until I re-read the privacy issue.
> I also think the algorithm won't be a parameter, it'll be hardcoded to md5.

Ah, yeah, I missed the privacy point. Good one, yeah!

tl;dr: ignore my previous comment, it's mainly invalid if you have privacy in
mind.

If we have some security/privacy concerns, then I think my argument about MD5
is a bit moot. We should probably use a more secure algorithm so it's
difficult to reverse the hash (which is not that much with MD5).

In addition, I think I saw a mention of a salt by username somewhere; I think
it would be a good idea.

On related matters, did you see how Troy Hunt implemented k-anonimyt for HaveIBeenPwend? https://www.troyhunt.com/were-baking-have-i-been-pwned-into-firefox-and-1password/#enablinganonymoussearcheswithkanonymity

j0k3r:
While it might eventually be a good idea to enable salt I think it’s really overkill for the feature.
I do not think it’s important to preserve that much the url we want to check. We are not dealing with password, email or a real sensible information here FMPOV.

Issue #2822, opened 2017-02-01
2FA via third party apps


n00b12345:
I am having a bit of trouble getting 2FA emails delivered to me. 

Wouldn't it be more convenient from a user's perspective to have codes delivered locally via Google authentication app or Authy etc?

Thanks!

nicosomb:
Hello @n00b12345. With the library that we use for 2FA by email, we can handle Google Authenticator. 
But we need to add a new field in database, so I planned this issue for 2.3.0.

n00b12345:
@nicosomb awesome! Thanks!

hvhaugwitz:
It would be very useful to be able use an authentication app instead of email. So I second this feature request.

What is the state of this issue? Why the 'add-google-authenticator' pull request has been closed without being merged?

j0k3r:
Nobody is working on it atm.
And I guess it was closed because of this comment https://github.com/wallabag/wallabag/pull/3170#issuecomment-307323576

hvhaugwitz:
The comment seems just to be about the wording.

So what one can do to get this feature into wallabag?

74cmonty:
Nextcloud, a fork of Owncloud, is using [this](https://github.com/nextcloud/twofactor_totp#readme) working 2FA OTP.
It is open source, and why should this not work with wallabag after some adaption?

PR #3170, opened 2017-05-31
Added Google authenticator

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #2822 
| License       | MIT

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -237,11 +237,17 @@ scheb_two_factor:
         cookie_lifetime: 2592000
 
     email:
-        enabled: "%twofactor_auth%"
+        enabled: false
         sender_email: "%twofactor_sender%"
         digits: 6
         template: WallabagUserBundle:Authentication:form.html.twig
         mailer: wallabag_user.auth_code_mailer
+
+    google:
+        enabled: true
+        server_name: wallabag
+        issuer: wallabag
+        template: WallabagUserBundle:Authentication:form.html.twig
 
 kphoen_rulerz:
     targets:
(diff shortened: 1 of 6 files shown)


app/config/config.yml
```diff
@@ -237,12 +237,18 @@ scheb_two_factor:
...
-        enabled: "%twofactor_auth%"
+        enabled: false
         sender_email: "%twofactor_sender%"
         digits: 6
         template: WallabagUserBundle:Authentication:form.html.twig
         mailer: wallabag_user.auth_code_mailer
 
+    google:
+        enabled: true
+        server_name: wallabag
```
tcitworld: One of these should depend on the instance's domain.

tcitworld: Issuer can be wallabag but server_name should be a new `%domain_name%`.

tcitworld: See https://github.com/wallabag/wallabag/pull/3173

j0k3r: Regarding #3173 I was wondering if the Google auth needs a _domain name_ with or without the protocol?

tcitworld: AFAIK it doesn't care about that, it's just an information to generate crypto.

tcitworld:
Maybe refer to it as OTP instead of Google Auth.

childnode:
@nicosomb can you clarify why this has been closed / aborted?

Issue #2766, opened 2017-01-18
Drop SQLite support


nicosomb:
All is in the title. SQLite is not nice for migrations. 🤢

- [x] Defining MySQL as the default database https://github.com/wallabag/wallabag/pull/3171
- [ ] remove SQLite build from Travis
- [ ] remove all migrations related to SQLite
- [ ] remove documentation about SQLite (in doc.wallabag.org)
- [ ] remove choice of SQLite when installating wallabag
- [ ] remove all tests related to SQLite
- [ ] communicate about dropping FULL support of SQLite in wallabag (using twitter, etc.)
- [ ] create a dumping script (in whatever language) to migrate a production running instance of wallabag from SQLite to MySQL or PostgreSQL

tcitworld:
(officially, as it's quite handy to develop on quicky)

Kdecherf:
As I said earlier on Twitter, SQLite is quite useful for development when you don't have any rdbms available or you don't want any local rdbms (and it's my case).

Could we simply stop supporting migrations for SQLite and add warnings saying that this engine is reserved for fresh installs in dev mode?

j0k3r:
I agree with @Kdecherf.
We should keep SQLite because it's faster to test wallabag. But we should really warn user to not use SQLite in production mostly because we don't support migration for it.

nicosomb:
And we need to change the default database type in parameters.yml.

j0k3r:
Also, 3.0.0 seems to far for me about that point.

nicosomb:
FYI, I removed the question mark in the issue title.

tcitworld:
This means a lot.

n00b12345:
Thanks for retaining SQLite. I'm the sole user for my setup on a raspberry pi. And SQLite is well, light on resources.

j0k3r:
The main problem is: it's really **really** hard to generate and maintain migration scripts for SQLite.

We'll drop support for migration in the coming version. But I think we'll let user use SQLite if they want to test wallabag. But we'll definitely warn user and not recommend SQLite to be used in production

zertrin:
What's the plan for all of us who just installed wallabag with the default option?

Because let's face it, SQLite is not only the dev-friendly option, but also a very user friendly one, it's only unfriendly for those writing migrations ^^.

I mean, normal users may not really understand this decision (what's a migration? why should I care?).

What's the transition plan for them? 

[...]

PR #3171, opened 2017-05-31
Defined MySQL as the default rdbms for wallabag

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes ⚠️
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #2766
| License       | MIT

In fact, maybe we only need to changes parameters.yml.dist. @wallabag/core I want your opinion. 
We also need to update http://github.com/wallabag/doc/ I will do that after when this PR will be merged.

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -20,7 +20,7 @@ parameters:
     database_path: null
     database_table_prefix: wallabag_
     database_socket: null
-    # with PostgreSQL and SQLite, you can set "utf8"
+    # with PostgreSQL and SQLite, you must set "utf8"
     database_charset: utf8mb4
 
     mailer_transport: smtp
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -10,17 +10,18 @@ parameters:
     # database_user: %env.database_user%
     # database_password: %env.database_password%
 
-    database_driver: pdo_sqlite
+    database_driver: pdo_mysql
     database_host: 127.0.0.1
     database_port: ~
     database_name: symfony
     database_user: root
     database_password: ~
-    database_path: "%kernel.root_dir%/../data/db/wallabag.sqlite"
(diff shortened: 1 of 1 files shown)


app/config/parameters.yml.dist
```diff
@@ -10,17 +10,18 @@ parameters:
...
     database_user: root
     database_password: ~
-    database_path: "%kernel.root_dir%/../data/db/wallabag.sqlite"
+    # For SQLite, database_path should be "%kernel.root_dir%/../data/db/wallabag.sqlite"
+    database_path: null
     database_table_prefix: wallabag_
     database_socket: null
-    # with MySQL, use "utf8mb4" if you got problem with content with emojis
-    database_charset: utf8
+    # with PostgreSQL and SQLite, you can set "utf8"
```
tcitworld: must

tcitworld:
If you'd ask me, I would have set Postgres as default. :P

nicosomb:
utf8mb4 was introduced in MySQL 5.5.3: 2010-03-24 https://dev.mysql.com/doc/relnotes/mysql/5.5/en/news-5-5-3.html We can consider that everyone has recent MySQL server, right?

app/config/parameters.yml.dist
```diff
@@ -10,17 +10,18 @@ parameters:
     # database_user: %env.database_user%
     # database_password: %env.database_password%
 
-    database_driver: pdo_sqlite
+    database_driver: pdo_mysql
     database_host: 127.0.0.1
     database_port: ~
     database_name: symfony
```
j0k3r: Can we also rename the default database to `wallabag`?

tcitworld:
We could also change the `root` user to `wallabag`.

nicosomb:
I already updated documentation repository (https://github.com/wallabag/doc/commit/c6610a3c163c7fb3398fe1ba9e1c2c9f7998fd33)

Issue #2964, opened 2017-03-05
Download images locally


ngosang:
### Issue details

1) Enable "Download images locally"
2) Add => https://www.tvaddons.ag/realdebrid-kodi-jarvis/
3) Most images are not stored locally.

![2017-03-05 16_43_04-configuring the real-debrid service for kodi 16 1 jarvis or below tvaddons ag](https://cloud.githubusercontent.com/assets/10577978/23588805/200ce608-01c3-11e7-9e6c-496c0f9f4d3b.png)

### Environment

* wallabag 2.2.2

j0k3r:
Thanks for the report, we'll have a look

prolibre:
idem for that

https://theconversation.com/conversation-avec-gerald-bronner-ce-nest-pas-la-post-verite-qui-nous-menace-mais-lextension-de-notre-credulite-73089

thanks,

Anthony

Alwaysin:
Could reproduce.

j0k3r:
@Alwaysin could you please share the related url to reproduce?

j0k3r:
@ngosang the problem was related to the symbol `&` in each url. This'll be fixed in https://github.com/wallabag/wallabag/pull/3176

@prolibre the problem with this url is we can't determine if it's a real image so we can't find the extension. The server respond with an empty header instead of some information about image (if it's a jpg, gif, etc..) http://hurl.eu/hurls/45edd6b5e31d8460c152cab9f85ffa1e33b51422/256e61fdf52426ab47f213789804fedd912edefb
Yes we can check about the final extension of the url but we can't 100% sure it's an image. It can be a zip file with a jpg at the end. So I guess, we won't be able to fix the problem in that particular case.

@Alwaysin could you please share the related url to reproduce? So I can try myself to see what's wrong. Otherwise you can also check `var/log/prod.log` and look for `DownloadImages`. It might have a line with an error about that.

Alwaysin:
Example with: https://cdn.theconversation.com/files/164136/width1356x668/image-20170405-14593-pvu9r5.JPG

j0k3r:
@Alwaysin I mean the article url not the image 🙂

Alwaysin:
Sorry, edited my answer. Here is the log:

[2017-06-01 18:36:29] app.DEBUG: DownloadImages: working on image: https://cdn.theconversation.com/files/164136/width1356x668/image-20170405-14593-pvu9r5.JPG [] []
[...]

PR #3176, opened 2017-06-01
Replace images with & in url

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2964
| License       | MIT

Images with `&` in the path weren’t well replaced because they might be with `&amp;` in the html instead.

Replacing `&` with `&amp;` fix the problem.
It might not fix all problem related to images download so we shouldn't close the related issue yet.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/DownloadImages.php
+++ b/src/Wallabag/CoreBundle/Helper/DownloadImages.php
@@ -66,6 +66,12 @@ class DownloadImages
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


j0k3r:
That's a good question, I don't know for now.

Issue #3056, opened 2017-04-25
Add api endpoint for create new user


bourvill:
- [x] Is it possible to create a new API to add a user?
- [x] In the response I suggest the addition of a previously created client_id and client_secret

tcitworld:
I'm not sure the new password should pass through the API, what do you think ?

bourvill:
This suggestion is to save time on creating a new token for the customer.

Typically after the creation of the user this one will directly connect. 
But can we put it on another endpoint?

nicosomb:
The second part of your issue seems to be a duplicate of #3056 by @jlnostr, right?

tcitworld:
\o/ \o/
/o/ \o\
/oo\
|o|
\o/
Fusion !

nicosomb:
Yah ! 

> When a user wants to use the API from his third-party application, it's required to create an application in the backend. Maybe we can add an API endpoint that automatically creates a client with a specific name (related to #1765) when the login is successful. This way the user only needs to enter the URL of his wallabag instance and the credentials. The rest would be done by the API.

from @jlnostr

j0k3r:
First step done.

nicosomb:
@bourvill @di72nn @jlnostr  (and also @Rurik19 @Simounet maybe): you can now create a user through the API in your applications.

j0k3r:
Well, it'll be available in the 2.3.0 release :slightly_smiling_face:

nicosomb:
Sure, but they can prepare their applications, to be OK when we'll release 2.3 (I dream of a release of each of our applications when we publish a minor release of the web app)

PR #3177, opened 2017-06-02
API user creation behing a toggle

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | https://github.com/wallabag/doc/pull/5
| Translation   | yes
| Fixed tickets | https://github.com/wallabag/wallabag/issues/3056
| License       | MIT

Following https://github.com/wallabag/wallabag/pull/3065

I've added a toggle feature (in internal settings) so that user api creation can be disabled while form registration still can be enabled.
Also, the /api/user endpoint shouldn't require authentication. Even if we check the authentication when sending a GET request, to retrieve current user information.

I've moved all the internal settings definition to config to avoid duplicated place to define them.
I don't know why we didn't did that earlier.

@wallabag/translators :globe_with_meridians: can I get the translation for:

> Enable user to be registered using the API

Thanks :+1:


tcitworld:
> I've moved all the internal settings definition to config to avoid duplicated place to define them.

Maybe a dedicated config file for them ? Otherwise the `config.yml` file is painful to read.

src/Wallabag/ApiBundle/Controller/UserRestController.php
```diff
@@ -43,16 +43,16 @@ public function getUserAction()
      */
     public function putUserAction(Request $request)
     {
-        if (!$this->container->getParameter('fosuser_registration')) {
+        if (!$this->getParameter('fosuser_registration') || !$this->get('craue_config')->get('api_user_registration')) {
```
tcitworld:
```php
(!($this->getParameter('fosuser_registration') && $this->get('craue_config')->get('api_user_registration')))
```
?

j0k3r: I found it more clear like I did

matteocoder:
@j0k3r  "Abilita la registrazione dell'utente attraverso l'API".

tcitworld:
Since the `wallabag_core` settings were moved to `wallabag.yml` file, we may have to check docs for changes.

tcitworld:
`putUserAction` should return HTTP 201 or HTTP 202.

src/Wallabag/CoreBundle/DataFixtures/ORM/LoadSettingData.php
```diff
@@ -6,173 +6,27 @@
...
-            ],
-            [
-                'name' => 'restricted_access',
-                'value' => '0',
-                'section' => 'entry',
-            ],
-        ];
-
-        foreach ($settings as $setting) {
+        foreach ($this->container->getParameter('wallabag_core.default_internal_settings') as $setting) {
```
nicosomb: 👍

PR #3184, opened 2017-06-05
Use an alternative way to detect images

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/2964
| License       | MIT

When parsing content to retrieve images to save locally, we only check for the content-type of the image response.
In some case, that value is empty.
Now we’re also checking for the first few bytes of the content as an alternative to detect if it’s an image wallabag can handle.
We might get higher image supports using that alternative method.


tcitworld:
Could this bring security issues ?

j0k3r:
If a binary resource starts with the exact same bytes as an image it could be a security issue.
But I think `imagecreatefromstring` (used right after) might fail if the string isn't a real image and in that case, we won't save the binary

PR #3187, opened 2017-06-07
Create (and return) a client after creating a new user using the API

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/3056
| License       | MIT

While creating a new user using the API, we also create a new client for the current user.
So the app which just create the user can use its newly created client to configure the app.

That new client is only return after creating the user.
When calling the endpoint /api/user to get user information, the new client information won’t be return.

Here is a response when creating a user:

```json
{
  "id": 5,
  "username": "username",
  "email": "email@google.io",
  "created_at": "2017-06-07T23:23:19+0200",
  "updated_at": "2017-06-07T23:23:19+0200",
  "default_client": {
    "client_secret": "3srtm3ign4ow84oocss0c4s4gscswo44kwkowocs8ggwgowk88",
    "client_id": "2_1t041dsqpae8kg4swo8gksscwks8kw4g0wwso84ocgk8kwkk8w",
    "name": "Default client"
  }
}
```

And when you request a logged in user:

```json
{
  "id": 5,
  "username": "username",
  "email": "email@google.io",
  "created_at": "2017-06-07T23:23:19+0200",
  "updated_at": "2017-06-07T23:23:19+0200"
}
```

Poke @jlnostr @bourvill

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/UserRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/UserRestController.php
@@ -9,6 +9,7 @@ use Nelmio\ApiDocBundle\Annotation\ApiDoc;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Wallabag\UserBundle\Entity\User;
+use Wallabag\ApiBundle\Entity\Client;
 
 class UserRestController extends WallabagRestController
 {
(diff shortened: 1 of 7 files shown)


Rurik19:
> When calling the endpoint /API/user to get user information, the new client information won’t be return.

Why not? If user reinstalls the application or just install it, already having a wallabag user - it will enter the client ID + secret by hand? Application should have more control about client via API

j0k3r:
Well at the moment a fully authenticated user means:
- a username
- a password
- a client_id
- a client_secret

The endpoint `/api/user` required a fully authenticated user. So you won't be able to retrieve user information without client_id & client_secret.

In your case, user will have to re-type the client_id & client_secret after re-installing the app.

As far as I understand this could change in the future, see https://github.com/wallabag/wallabag/pull/3068.

tcitworld:
> If user reinstalls the application or just install it, already having a wallabag user - it will enter the client ID + secret by hand?

Just like right now.

src/Wallabag/ApiBundle/Controller/UserRestController.php
```diff
@@ -97,29 +98,38 @@ public function putUserAction(Request $request)
                 ->setStatusCode(JsonResponse::HTTP_BAD_REQUEST);
         }
 
+        // create a default client
+        $client = new Client($user);
+        $client->setName('Default client');
```
tcitworld: Instead of putting `Default client`, we could require the app to provide a name in the request.

Rurik19:
Hi all!
in current branch 2.3 this path enabled?  I got 404 when I try to call /api/user

j0k3r:
Calling using PUT?

Rurik19:
Yeah. @di72nn said here https://github.com/wallabag/wallabag/pull/3227 that 2.3 branch was rolled back. maybe /api/user was rolled back as well?

Issue #3186, opened 2017-06-07
Validate the fetched item properties


jlnostr:
When I checked the database of @JeffMelton to determine what caused the windows app to crash, I found the issues in the database. I found json entries like:

    "preview_picture":"https:"

or

     "language":"{{locale}}"

The first one isn't a valid uri according to the RFC standards, and the second one isn't a language code. A basic check if the fetched data is correct would be nice.

j0k3r:
Good suggestion @jlnostr !

PR #3192, opened 2017-06-08
Validate language & preview picture fields

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/3186
| License       | MIT

_This is a proposal.
Any suggestions on which fields might be a good to validate is welcome._

Instead of saving the value of each field right into the content without any validation, it seems better to validate them **before**.
This might sounds obvious now we say that but we didn't do it before that PR.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -19,6 +19,7 @@ class ContentProxy
 {
     protected $graby;
     protected $tagger;
+    protected $validator;
     protected $logger;
     protected $mimeGuesser;
     protected $fetchingErrorMessage;
@@ -127,7 +128,7 @@ class ContentProxy
             isset($content['open_graph']['og_image']) ? $content['open_graph']['og_image'] : ''
         );
 
-        // if content is an image define as a preview too
+        // if content is an image, define it as a preview too
         if (!empty($content['content_type']) && in_array($this->mimeGuesser->guess($content['content_type']), ['jpeg', 'jpg', 'gif', 'png'], true)) {
             $this->validateAndSetPreviewPicture(
                 $entry,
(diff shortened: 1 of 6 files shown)


nicosomb:
Nice 👍

nicosomb:
> Any suggestions on which fields might be a good to validate is welcome.

For me, it looks OK.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -7,6 +7,9 @@
 use Wallabag\CoreBundle\Entity\Entry;
 use Wallabag\CoreBundle\Tools\Utils;
 use Symfony\Component\HttpFoundation\File\MimeType\MimeTypeExtensionGuesser;
+use Symfony\Component\Validator\Constraints\Language as LanguageConstraint;
```
tcitworld: Language is `en` and not `en_GB` for instance. However, the OpenGraph `og:locale` format is `language_TERRITORY`, so it should be `Symfony\Component\Validator\Constraints\Locale` instead.

Handling merging `en_GB` and `en_US` into `en` should be done somewhere else.

nicosomb: You can also do this for `fr_FR`, `fr_*`, etc.

j0k3r: Hum, I'll check that.
We don't use the `og:locale` as a fallback for the language field return by graby.

tcitworld:
This has a breaking change : if language or preview_picture are not found, it returns `''` instead of `null`. Therefore it saves an empty string in the database for nothing, templates show empty pictures (since we check for `null` and not `empty`) and filters don't work anymore.

Issue #3032, opened 2017-04-09
RSS feed for tags


jamietanna:
Certain tags, such as `work` would be useful to consume via some other application such that I have clear separation of recommended reading for work, and personal reading. Having a separate RSS feed with the tag would be useful.

Issue #2769, opened 2017-01-19
RSS Support for all Articles


maitrenageur:
Actual support is for unread, starred and archived articles.
It is also intesting to have all articles in one view.

Issue #626, opened 2014-04-09
Public feeds with different tokens


shtrom:
Currently, Wallabag exports a reading list, favourites and archive as RSS feeds hidden by a single random token.

It might be desirable to share some of these feeds (e.g., favourites) publicly. However, doing so exposes the secret token, and would let the public access the other feeds too.

It would be good to be able to differentiate both. My thoughts at the moment are that this could be done by either
1. providing one separate token per feed or,
2. using a public token with limited access.

Option (1) might be the most flexible, and perhaps easiest to implement.

supermamie:
Why not adding a "second token" in exports ?
This would maybe limit the impact on the code ?
For example, now it is :
https://my-instance.fr/user/ToKeN/all.xml
https://my-instance.fr/user/ToKeN/tags/mytag.xml
https://my-instance.fr/user/ToKeN/tags/mytag2.xml
https://my-instance.fr/user/ToKeN/tags/mytag3.xml

Could it be somoething like
https://my-instance.fr/user/ToKeN/tOkEnAll/all.xml
https://my-instance.fr/user/ToKeN/tOkEnMyTaG/tags/mytag.xml
https://my-instance.fr/user/ToKeN/tags/tOkEnMyTaG2/mytag2.xml
https://my-instance.fr/user/ToKeN/tags/tOkEnMyTaG3/mytag3.xml

So if I want to publish a rss feed for one specific tag, people won't be able to guess the url for all my posts.

(Writing this, I am not sure anymore that the "ToKen" part is still needed in the url, it could be determined serverside based on the user/newToken couple, and avoid to give any more information than needed)

And this feature could be totally different than what you talk about on a legal way (another issue can be open for this)

PR #3207, opened 2017-06-10
Add RSS for tags

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | not needed
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/3032 & https://github.com/wallabag/wallabag/issues/2769
| License       | MIT

Adding RSS for tags.
Url are `https://wallabag.instance/admin/43kIG075GUoLZiU/tags/foot.xml`

Also, I'm wondering where should we display RSS link for tag.
- put a sample url in the config page next to other RSS link (like "for tag please use that pattern `/username/token/tags/tag-slug.xml`")
- put a link on the tag page


nicosomb:
👍  for `put a link on the tag page`, or near the tag name on full tags page.

tcitworld:
Can't choose between the two, will have a look at how it feels.

src/Wallabag/CoreBundle/Controller/ExportController.php
```diff
@@ -57,16 +56,17 @@ public function downloadEntriesAction(Request $request, $format, $category)
...
+        $epository = $this->get('wallabag_core.entry_repository');
 
         if ($category == 'tag_entries') {
-            $tag = $this->getDoctrine()->getRepository('WallabagCoreBundle:Tag')->findOneBySlug($request->query->get('tag'));
+            $tag = $this->get('wallabag_core.tag_repository')->findOneBySlug($request->query->get('tag'));
 
-            $entries = $this->getDoctrine()
-                ->getRepository('WallabagCoreBundle:Entry')
-                ->findAllByTagId($this->getUser()->getId(), $tag->getId());
+            $entries = $epository->findAllByTagId(
```
tcitworld: typo

src/Wallabag/CoreBundle/Controller/ExportController.php
```diff
@@ -57,16 +56,17 @@ public function downloadEntriesAction(Request $request, $format, $category)
...
+                $this->getUser()->getId(),
+                $tag->getId()
+            );
         } else {
-            $entries = $this->getDoctrine()
-                ->getRepository('WallabagCoreBundle:Entry')
+            $entries = $epository
```
tcitworld: same

j0k3r:
Feeds icon screen:

image

Shouldn't we add an icon somewhere in a tag page (I mean not only on the tag**s** page)

tcitworld:
Ideally, yes. And possibly a <link> tag to it in the headers.

jamietanna:
How will the URLs be generated? Will it be possible to regenerate them ie if you want to stop someone being able to access the feed? Also, will it be possible to have multiple URLs for the same feed?

j0k3r:
Like I said in the PR description:

> https://wallabag.instance/USERNAME/RSS_TOKEN/tags/TAG_SLUG.xml

If someone found your rss token and then can discover your feed, you just need have to regenerate your RSS token from your config panel and update your RSS reader with the new feed url.

What do you mean by multiple URLs for the same feed? Please describe what you need.

jamietanna:
Ah sorry I missed that. That makes sense.

> What do you mean by multiple URLs for the same feed?

Say I want to share the `work` tag with my work email, and my personal phone (via an RSS reader). But I want the chance to revoke the access to the work URL at any point. Having multiple, separate URLs would help this.

j0k3r:
You mean multiple tokens for each url I think.

jamietanna:
Yes, exactly

j0k3r:
@jamietanna there is already an issue about that https://github.com/wallabag/wallabag/issues/626.

tcitworld:
- Added « all entries » RSS feed
- Added rss button on all views
- Added tags rss on baggy
- Fixed some translation issues

j0k3r:
@tcitworld great, an other fixed tickets.
I've tested it on both baggy & material and it looks good to me :+1:

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
```diff
@@ -1,9 +1,16 @@
 {% extends "WallabagCoreBundle::layout.html.twig" %}
 
+{% block head %}
```
tcitworld: This part is missing on baggy.

j0k3r: Added

Issue #966, opened 2014-12-09
To support account management via LDAP


klorydryk:
Hi,
I would like to offer this wonderfull tool to my colleagues, in my company. But I can't manage the accounts (creation, deletion) manually. Neither ask to create a new password...
It would be great to support a directory service protocol as LDAP to do that.
Thanks!

conditionsofexistence:
+1

smknstd:
+1

stevenroose:
+1 or just a bump, sadly this one is delayed to 2.0. What's the eta for 2.0?

tcitworld:
> What's the eta for 2.0?

The upcoming weeks should see an alpha version be released. You may be interested by this page : https://www.wallabag.org/pages/roadmap.html

stevenroose:
Thanks! Good to see v2 coming along! Hopefully this feature will still make it to the milestone :)

nicosomb:
see http://symfony.com/blog/new-in-symfony-2-8-ldap-component

nebulade:
Hi, is there any update on the ldap front?

j0k3r:
Well, regarding the features list, I think this feature will more likely coming in the v2.1. What do you think @nicosomb ?

nicosomb:
Yes, we need to plan it for 2.1. We have to be focused on main features to release v2.0 as soon as possible. 
But we don't forget this feature.

jeromelebleu:
Hi! As we need it for the [YunoHost app](https://github.com/YunoHost-Apps/wallabag2_ynh), I tried to make something regarding @nicosomb link - thanks for that by the way - but I'm facing into knowledge lake to improve it... To not pollute this issue, I've created a specific one for that here: https://github.com/YunoHost-Apps/wallabag2_ynh/issues/1. Thank you in advance to those who can help!

nerzhul:
i agree that LDAP integration would be a must have. Self hosting nowadays needs centralized auth like LDAP.

JimboJoe:
Hi, do you see any chance to have LDAP integration implemented any time soon?
We would love to welcome Wallabag v2 as an official YunoHost app ;-)

tYYGH:
Hi! If I understand both issues’ comments correctly, both LDAP authentication and proxy-provided identity are now available in Wallabag, but not yet present in a release.
Would it be possible to create a new release to include LDAP authentication? Latest release is from long ago ;-)

j0k3r:
[...]

PR #3263, opened 2017-07-05
WIP: Add LDAP authentication

immae:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | https://github.com/wallabag/doc/pull/31
| Translation   | no
| Fixed tickets |  #966
| License       | MIT

This PR adds LDAP authentication to wallabag.
It is still a work in progress.
- *Need to find a way to make it optional* (in particular not making ldap extension mandatory if the user doesn't need it)
- Need to write some tests
- Need to write some documentation

Features:
- Find, create and update users based on LDAP informations
- Handles both kinds of users (non-LDAP and LDAP)
- Optionnally adds admin role based on LDAP information
- Works with the API too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ branches:
 
 before_script:
     - PHP=$TRAVIS_PHP_VERSION
+    - echo "extension=ldap.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - if [[ ! $PHP = hhvm* ]]; then echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini; fi;
     # xdebug isn't enable for PHP 7.1
     - if [[ ! $PHP = hhvm* ]]; then phpenv config-rm xdebug.ini || echo "xdebug not available"; fi
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -42,6 +42,10 @@ class AppKernel extends Kernel
             new OldSound\RabbitMqBundle\OldSoundRabbitMqBundle(),
         ];
 
+        if (class_exists('FR3D\\LdapBundle\\FR3DLdapBundle')) {
+          $bundles[] = new FR3D\LdapBundle\FR3DLdapBundle();
+        }
+
         if (in_array($this->getEnvironment(), ['dev', 'test'], true)) {
             $bundles[] = new Symfony\Bundle\DebugBundle\DebugBundle();
             $bundles[] = new Symfony\Bundle\WebProfilerBundle\WebProfilerBundle();
--- a/app/DoctrineMigrations/Version20170710113900.php
+++ b/app/DoctrineMigrations/Version20170710113900.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Added dn field on wallabag_users
+ */
+class Version20170710113900 extends AbstractMigration implements ContainerAwareInterface
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
+    private function getTable($tableName)
+    {
+        return $this->container->getParameter('database_table_prefix').$tableName;
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $usersTable = $schema->getTable($this->getTable('user'));
+
+        $this->skipIf($usersTable->hasColumn('dn'), 'It seems that you already played this migration.');
+
+        $usersTable->addColumn('dn', 'text', [
+            'default' => null,
+            'notnull' => false,
+        ]);
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        $usersTable = $schema->getTable($this->getTable('user'));
+        $usersTable->dropColumn('dn');
+    }
+}
+
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -60,3 +60,23 @@ parameters:
 
     # sites credentials
     sites_credentials: {}
+
+    # ldap configuration
+    # To enable, you need to require fr3d/ldap-bundle
+    ldap_enabled: false
+    ldap_host: localhost
+    ldap_port: 389
+    ldap_tls:  false
+    ldap_ssl:  false
+    ldap_bind_requires_dn: true
+    ldap_base: dc=example,dc=com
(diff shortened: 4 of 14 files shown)


j0k3r:
Looks interesting!
Do not forget to run `php bin/php-cs-fixer fix src/`

tcitworld:
> in particular not making ldap extension mandatory if the user doesn't need it

It shouldn't be required in the composer.json file but checked when the LdapHydrator is used.

tcitworld:
Ccing some YunoHost wallabag 2 package contributors here @jeromelebleu @jimbojoe @lapineige

immae:
@tcitworld : the problem I have is not removing it from composer.json, but handling the rest of the configuration. The LdapHydrator happens too late in the process

tcitworld:
You may add a listener to the `kernel.request` event and check it here, but it may be bad in terms of performance.

lapineige:
Thanks @tcitworld :)
It would indeed be really cool to integrate that in our wallabag v2 package :)

immae:
@tcitworld I spent a few days trying to learn more of the internal ways of symfony (especially the kernel). Now I "just have" to write it correctly so that it happens at "compilation" time, but it's doable :)

j0k3r:
Maybe you can hook inside the `WallabagUserExtension`?

immae:
Ok, now LDAP is completely optional:
- If you want to use ldap, you need to "composer require fr3d/ldap-bundle"
- Otherwise, no ldap is needed (not even the php extension)

@j0k3r : That's what I was trying to do, but services and parameters have different priority rules so not everything can happen at load time.

j0k3r:
> If you want to use ldap, you need to "composer require fr3d/ldap-bundle"

This'll be a no go for the final user.
Composer might fail to run in the final user server because it'll need more RAM, etc.

We should provide a package with the bundle in it and a single option to enable ldap during installation (for example in parameters.yml) or not. This should be a dead simple as possible for the final user.

immae:
@j0k3r : it's not possible, the fr3d/ldap-bundle requires php ldap extension. Unless you want to make it mandatory for everyone, even people who don't want ldap ?

immae:
@j0k3r that was my main problem in the last days: making ldap extension not mandatory if the user doesn't need it, so that someone who needs ldap can simply activate it on demand

immae:
Note that the "composer require fr3d/ldap-bundle" can be integrated in the "make install" script if necessary, which already contains composer commands.

Would that be suitable for you?

immae:
I changed the install script so that it can be enabled simply with:

`make install LDAP_ENABLED=true`

+ updating configuration in parameters.yml

immae:
documenation added

nicosomb:
I think that we need a separated bundle for this feature, like I did for https://github.com/nicosomb/wallabag-carbondate-bundle and https://github.com/nicosomb/wallabag-camo-bundle.

immae:
@nicosomb I cannot qualify that as a "simple" setup...

I’m sorry, but I’m not willing to spent more time on that. This PR works and I'm already using is successfuly in production. I made it "dead simple as possible for the final user" as required, as well as making it completely optional. I’m okay to do some work to add unit and functional tests, but if the rest doesn’t suit you I’ll just close the PR and let someone else do it the right way.

ViViDboarder:
It seems a shame to throw away a pr for a 3 year old feature request.

Is there guidance as to why some features should be separate bundles vs included? It seems like many features and integration are included by default. Is there something that makes this a candidate for exclusion? I don't see anything in the developer documentation as to when or how to make external bundles...

immae:
ViViDboarder: As far as I know, they don’t really care.

But you can just cherry-pick the patch and use it as is, I use it successfully for months now without any issue.

almet:
Hi everyone!

First and more importantly, thanks @immae for taking the time to write this. Don't be mistaken: this is really appreciated!

Now, what we need is to find an acceptable way to get this merged upstream. My understanding of the issue is that the fact this code relies on a php-module makes it hard to have in the core. I can see two outcomes for it:

- Package it as a separate plugin, as @nicosomb suggested. It probably makes it harder to install for the end user but relaxes the expectations to merge this in wallabag directly;
- Detect if the module is enabled, and if it's the case propose an option to enable LDAP support in the settings.

Wallabag team: would option 2 be an acceptable approach?

ViViDboarder:
I have very little (read: no) experience with PHP, however if there is any kind of guidance as to how one would bundle it, I'd take a stab.

However, I am having trouble understanding why. It seemed like the original objection was that it was too burdensome on the users to run `composer require fr3d/ldap-bundle` before installing. However that seems to be counter to the bundle methodology since all those bundles also seem to require an additional command plus additional configuration.

This probably stems from me having no clue how Symphony works or the design guidelines for Wallabag. Symphony seems huge, so not sure where a relevant place to begin would be and the developer docs for Wallabag seem pretty sparse.

Like I said, happy to help, but would appreciate a point in the right direction!

DMW007:
Any news on this? Wallabag is a nice piece of software, but would be the only app without LDAP on my server. This prevents me from using it in production. I saw feature requests for LDAP that were created years ago. It would be nice if we can find a solution for this issue.

immae:
Hey there,
Since the discussion seems impossible at the moment, but I still get people contacting me to do something to help having LDAP in wallabag, I created a branch on a fork of this repository, containing the ldap feature.

It’s currently rebased above 2.3.2 (last version), and I’ll try to follow up when there are new releases:

https://git.immae.eu/?p=github/wallabag/wallabag.git;a=shortlog;h=refs/heads/gitolite_local/ldap

Hope it will be useful.

joshp23:
Just some input. It seems appropriate to simply bundle this in and allow the admin to switch LDAP on or off in the GUI. Am I missing something about this that makes that a no-go?

j0k3r:
https://github.com/wallabag/wallabag/pull/3263#issuecomment-314438486

j0k3r:
Just to let you know that we might publish a new bundle to enable LDAP on wallabag.
It'll require some manual step to install it (and for sure won't work on Raspberry because of lack of ram).
We'll publish it under the wallabag organisation.
We'll let you know when it's available.

ismaelbouyaf:
Thanks for the news @j0k3r . In the meantime, my branch https://git.immae.eu/?p=github/wallabag/wallabag.git;a=shortlog;h=refs/heads/gitolite_local/ldap is rebased on top of last release (2.3.6), and requires less RAM to prepare: I added the lines in the composer.lock / composer.json, which was the main problem of the previous method (The `composer require` don’t need to be done on the target host, it can be done at buildtime)

Someone asked me yesterday help to enable ldap on their instance, I found out it was not as trivial as I thought and detailed steps here: https://forum.chatons.org/t/wallabag-authentification-avec-ldap/241/2 (in French)

j0k3r:
@immae @ismaelbouyaf if you want I create a dedicated repo in the wallabag organization so you can export your work into a bundle (like `WallabagLdapBundle`)?

immae:
Hi @j0k3r, sure it would be great. However I don’t know how to package it as a bundle, since I modify some existing entities to add elements (like "User" for instance), do you have example so that I can find my way?

j0k3r:
We can accept some changes into the core too after reviewing them.
About bundle, you can check https://github.com/topics/wallabag-plugin

Issue #2821, opened 2017-02-01
make the EPUB prettier


anarcat:
### Issue details

Now that I have a nice EPUB importer for my ebook reader ([wallabako](https://gitlab.com/anarcat/wallabako/)), I am getting EPUBs from Wallabag, and it's awesome. Thanks for that feature so much!!

They do look a bit ... well, they're ugly. :) The first page says:

> Produced by wallabag with PHPePub
>
> Please open [an issue](https://github.com/wallabag/wallabag/issues) if you have trouble with the display of this E-Book on your device.

This could be much better. First off, this text is great, but should be at the end of the ebook, not the beginning.

Second, the first page should be the title of the book, with the author. Right now, the author is always "wallabag" - how about putting the domain name in there or some other metadata?

Third, a table of contents would be nice. We could parse `<H[1-9]>` tags in the document and generate a table of contents with that?

Finally, how about some images? We do have some image data for articles, we display it on the website - that would make for an awesome cover page!

So, in short:

- [ ] put debug notes at the end
- [ ] set epub author to the article domain or some other metadata
- [ ] generate table of contents from the headings in the document
- [ ] add a cover page based on the article main image
- [ ] `sudo make unicorn` ;)

Thanks!

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.1.5
* How did you install wallabag? Via `git clone` or by downloading the package? unknown
* Last wallabag version that did not exhibit the issue (if applicable): N/A
* php version: unkonwn
* OS: Debian GNU/Linux
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): N/A

### Steps to reproduce/test case

Download an EPUB version. Look at it. Just look at it! ;)

Quent-in:
Could we also externalize the text strings so it can be translated ?
To me, it will help epubs to look prettier :)

j0k3r:
@Quent-in of course!

pmartin:
I was about to create an issue, but I think it would be kind of a duplicate of this one and it's just my opinion and not an "issue", so posting here:

 * The 1st page of EPUB exports ("produced by wallabag") is not really *useful*: I have to skip it each time I open an EPUB and never read it as it's always the same. It could be either removed, or placed at the end of the EPUB file, +1 for me.
[...]

PR #3266, opened 2017-07-08
Use the article publisher as author for exported files

egilli:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | maybe
| Documentation | no
| Translation   | no
| License       | MIT

As mentioned in #2821, the author of an epub file is always "wallabag".

This PR introduces another author:
- When exporting a single article to epub, mobi, and pdf, use the article domain for the "author" metadata
- When exporting multiple articles in one file (all or category), the author is set to "_All authors_" or "_&lt;category> authors_", similar to the title ("_All articles_" or "_&lt;category> articles_"). Please tell me if you would rather display a list of all the authors of the exported articles; I can do that instead.

N.B.: I changed the author from array to string, because for now, there is always only one author.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/EntriesExport.php
+++ b/src/Wallabag/CoreBundle/Helper/EntriesExport.php
@@ -18,7 +18,7 @@ class EntriesExport
     private $logoPath;
     private $title = '';
     private $entries = [];
-    private $authors = ['wallabag'];
+    private $author = 'wallabag';
     private $language = '';
     private $footerTemplate = '<div style="text-align:center;">
         <p>Produced by wallabag with %EXPORT_METHOD%</p>
@@ -67,6 +67,24 @@ class EntriesExport
 
         if ('entry' === $method) {
             $this->title = $this->entries[0]->getTitle();
+        }
+
+        return $this;
+    }
+
+    /**
+     * Sets the author for just one entry.
+     *
+     * @param string $method Method to get articles
+     *
+     * @return EntriesExport
+     */
+    public function updateAuthor($method)
+    {
+        $this->author = $method.' authors';
+
+        if ('entry' === $method) {
+            $this->author = $this->entries[0]->getDomainName();
         }
 
         return $this;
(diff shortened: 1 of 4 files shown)


anarcat:
i think this is a great idea, but i wonder if there's article metadata that could be used to indicate the author, instead of the domain name?

there is, for example, such a thing as the [meta author](https://www.w3.org/TR/html5/document-metadata.html#meta-author) tag.

j0k3r:
For a given entry, you have the [`published_by` field](https://github.com/wallabag/wallabag/blob/2.3/src/Wallabag/CoreBundle/Entity/Entry.php#L133-L140) which can contains authors of the entry. It that field is empty, then you can fallback to the domain name.

But that field is only available in the 2.3 branch for now. So you change your targeted branch and rebase your branch against the 2.3

egilli:
I didn't know about this field, that's much better!
I rebased on 2.3 and used the publishedBy.

(Sorry for the branch deletion and close/reopen PR, I think it's ok now.)

src/Wallabag/CoreBundle/Helper/EntriesExport.php
```diff
@@ -73,6 +73,31 @@ public function updateTitle($method)
...
+    public function updateAuthor($method)
+    {
+        $this->author = $method.' authors';
+
+        if ('entry' === $method) {
+            $publishedBy = $this->entries[0]->getPublishedBy();
+            if (!empty($publishedBy)) {
+                $this->author = implode(', ', $this->entries[0]->getPublishedBy());
+            } else {
+                $this->author = $this->entries[0]->getDomainName();
```
j0k3r: You can avoid the `else` by redefining `$this->author = $this->entries[0]->getDomainName();` before the `if`.

Also, I think we can return early to avoid cascading spaces by doing:

```php
if ('entry' !== $method) {
    $this->author = $method.' authors';
    return $this;
}

// then do the job about the entry method
```

j0k3r:
Also, do not forget `php bin/php-cs-fixer fix`

egilli:
Thanks, done.

egilli:
Yes, it can be reviewed, thanks.

j0k3r:
Thanks @egilli !

egilli:
Thank you!

anarcat:
we just updated our host to 2.3 and i must emphasize that i really apppreciate this change! it makes a big difference in managing the epubs on [wallabako](https://gitlab.com/anarcat/wallabako), so thanks again!

Issue #3268, opened 2017-07-08
Article export API endpoint fails to return HTML


di72nn:
### Issue details

`/api/entries/{entry}/export.html` responds with `HTTP 500`.
Works fine with other (pdf, txt, etc.) formats.

Server log message:
```
request.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "The format "html" is not yet supported." at /usr/share/webapps/wallabag/src/Wallabag/CoreBundle/Helper/EntriesExport.php line 89 {"exception":"[object] (InvalidArgumentException(code: 0): The format \"html\" is not yet supported. at /usr/share/webapps/wallabag/src/Wallabag/CoreBundle/Helper/EntriesExport.php:89)"} []
```

Tested with 2.2.3 and with the 2.3 branch.

Mentioned here: wallabag/android-app#587.

j0k3r:
We never implemented the HTML export from the beginning. The error is normal to me.

Only supported export are: epub|mobi|pdf|json|xml|txt|csv

j0k3r:
You can follow https://github.com/wallabag/wallabag/issues/3272 to now when the html export will be available.

PR #3270, opened 2017-07-09
Disallow html for API templating formats

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | https://github.com/wallabag/wallabag/issues/3268
| License       | MIT

Using html template format will then put the html format in the allowed list for the api doc which we don’t want since the api doesn’t response for html format.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -110,7 +110,7 @@ fos_rest:
             epub: true
             mobi: true
         templating_formats:
-            html: true
+            html: false
         force_redirects:
             html: true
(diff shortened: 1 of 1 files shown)


tcitworld:
We could also support html export. :/

nicosomb:
> We could also support html export. :/

See #3272.

PR #3271, opened 2017-07-10
Add `given_url` in Entry table to check if a redirected url has already added

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

Some URLs are added twice (or more): wallabag should say that the URL is already in database. 
Example `https://fashionunited.com/news/business/jigsaw-posts-8-percent-increase-in-annual-sales/2017070316404` because, through wallabag, the saved URL is `https://fashionunited.com/news/business/jigsaw-posts-8-percent-increase-in-annual-sales/2017070316404?PageSpeed=noscript`.

Thanks to this PR, we now have 2 fields: the URL (the URL after the redirection) and the given URL (the URL given in the form or via the addon). 

We now check if a new URL is equal to url or given url.


src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -78,6 +78,7 @@ public function addEntryFormAction(Request $request)
                 return $this->redirect($this->generateUrl('view', ['id' => $existingEntry->getId()]));
             }
 
+            $entry->setGivenUrl($entry->getUrl());
```
j0k3r: You should think about the API and import too and use ContentProxy instead

tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
```diff
@@ -260,6 +260,41 @@ public function testPostNewOkUrlExistWithAccent()
         $this->assertContains('/view/', $client->getResponse()->getTargetUrl());
     }
 
+    public function testPostNewOkUrlExistWithRedirection()
+    {
+        $this->logInAs('admin');
+        $client = $this->getClient();
+
+        $url = 'https://fashionunited.com/news/business/jigsaw-posts-8-percent-increase-in-annual-sales/2017070316404';
```
j0k3r: Better use a bit.ly url or a t.co so it'll be obvious that the url will be redirected and given_url won't be the same as url.

nicosomb:
In fact, maybe we can merge this one to embed it into 2.3...

@wallabag/core what's your opinion?

src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -298,8 +298,11 @@ public function findAllByTagId($userId, $tagId)
      */
     public function findByUrlAndUserId($url, $userId)
     {
+        $url = urldecode($url);
+
         $res = $this->createQueryBuilder('e')
-            ->where('e.url = :url')->setParameter('url', urldecode($url))
+            ->where('e.url = :url')->setParameter('url', $url)
```
j0k3r: Shouldn't we add an index on url, givenUrl & user_id? Can you test it locally (hopefully on an heavy database)?

j0k3r:
I'm ok for 2.3 :shipit:

nicosomb:
Testsuite fails because we need to specify a length for the `url` field. 
But it's impossible to do that with Doctrine Migration (see https://stackoverflow.com/a/32540801/178414)
I think that I need to write the index creation manually...

nicosomb:
I need help/opinion about the SQLite part. 
Because we'll remove SQLite in 2.4, do I need to write migration for SQLite?

j0k3r:
Nope, you don't need migration for SQLite since you target 2.4.0

docljn:
Hello @j0k3r and @nicosomb I am participating in codetriage, and wondered if this PR was still an active issue, or what the next steps were if it was not to be closed? Thank you.

j0k3r:
Thanks for the codetriage, I think this PR is still relevant.
We just don't have time to work on it

docljn:
Thanks @j0k3r - sympathies on the lack of time.

j0k3r:
Following a bad rebase (against the master instead of the 2.4 branch) I force pushed.
Also, fixing tests so the PR is now ready to be reviewed.

j0k3r:
@Kdecherf @tcitworld This is ready.
I fucked up while wanting to use the `origin_url` instead of creating the `given_url` but it's not the same purpose and `origin_url` might not be always save (in case of ignore list).

I re-used the hashing system to ensure we can have an index on the query.
I haven't tested the perf for now.

app/DoctrineMigrations/Version20190601125843.php
```diff
@@ -0,0 +1,74 @@
...
+        $entryTable->addIndex(
+            [
+                'user_id',
+                'hashed_url',
+                'hashed_given_url',
+            ],
+            'hashed_urls_user_id',
+            [],
+            [
+                // specify length for index which is required by MySQL on text field
```
tcitworld: :see_no_evil:

app/DoctrineMigrations/Version20190601125843.php
```diff
@@ -0,0 +1,74 @@
...
+                'lengths' => [
+                    // user_id
+                    null,
+                    // hashed_url
+                    40,
```
tcitworld: Can you point out why this specific value ?

src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -368,6 +368,7 @@ public function findByHashedUrlAndUserId($hashedUrl, $userId)
     {
         $res = $this->createQueryBuilder('e')
             ->where('e.hashedUrl = :hashed_url')->setParameter('hashed_url', $hashedUrl)
+            ->orWhere('e.hashedGivenUrl = :hashed_given_url')->setParameter('hashed_given_url', $hashedUrl)
             ->andWhere('e.user = :user_id')->setParameter('user_id', $userId)
```
tcitworld: I can't remember if Doctrine respects the order of `WHERE` statements ? I think it may be faster if we have the condition on `user_id` first.

tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
```diff
@@ -266,6 +265,41 @@ public function testPostNewOkUrlExistWithAccent()
         $this->assertContains('/view/', $client->getResponse()->getTargetUrl());
     }
 
+    public function testPostNewOkUrlExistWithRedirection()
```
tcitworld: Add
```php
    /**
     * This test will require an internet connection.
     */
```

app/DoctrineMigrations/Version20190601125843.php
j0k3r: Because hash is a sha1 which is 40 length

src/Wallabag/CoreBundle/Repository/EntryRepository.php
j0k3r: This code generates this query:

```sql
SELECT e FROM Wallabag\CoreBundle\Entity\Entry e WHERE (e.hashedUrl = :hashed_url OR e.hashedGivenUrl = :hashed_given_url) AND e.user = :user_id
```
And I can't find a way to handle the `user_id` first and then the `OR` without _almost_ write the query by hand.
Also there is an index on each field of the query, it should fast enough

j0k3r: But the index is not used because we use a `OR`.
I'll split the query in two and create two indexes so they'll be used

PR #3280, opened 2017-07-19
Fix updatePublishedAt on already parsed article's date

Simounet:
Hi,
Don't know if it's the best way to do this but it a blocker to me because we're using it on wallabagger.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -122,7 +122,7 @@ class ContentProxy
      */
     public function updatePublishedAt(Entry $entry, $value)
     {
-        $date = $value;
+        $date = $value instanceof \DateTime ? $value->date : $value;
 
         // is it a timestamp?
         if (filter_var($date, FILTER_VALIDATE_INT) !== false) {
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -122,15 +122,21 @@ class ContentProxy
      */
     public function updatePublishedAt(Entry $entry, $value)
     {
-        $date = $value instanceof \DateTime ? $value->date : $value;
+        $date = $value;
 
         // is it a timestamp?
         if (filter_var($date, FILTER_VALIDATE_INT) !== false) {
-            $date = '@' . $value;
+            $date = '@' . $date;
         }
 
         try {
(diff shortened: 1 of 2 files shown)


src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -184,7 +184,8 @@ private function stockEntry(Entry $entry, array $content)
         }
 
         if (!empty($content['date'])) {
-            $this->updatePublishedAt($entry, $content['date']);
+            $contentDate = $content['date'] instanceof \DateTime ? $content['date']->date : $content['date'];
+            $this->updatePublishedAt($entry, $contentDate);
```
j0k3r: Yeah this is a good start but you should add that condition in `updatePublishedAt` instead

Simounet:
That was obvious. I updated it. Thanks.

j0k3r:
I've tested your problem with the same link from the previous issue: http://www.lemonde.fr/m-perso/article/2017/06/25/antoine-de-caunes-je-veux-avoir-le-droit-de-tatonner_5150728_4497916.html

And while patching the entry to star it I don't have the problem of the date (I haven't applied your patch).

Could you please re-describe step by step a way to reproduce the error on the up-to-date 2.3 branch?

Simounet:
Launch this with your server and credentials twice:
`curl 'https://wallabag.server.com/api/entries.json' -H 'Host: wallabag.server.com' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:55.0) Gecko/20100101 Firefox/55.0' -H 'Accept: application/json' -H 'Accept-Language: fr,fr-FR;q=0.8,en-US;q=0.5,en;q=0.3' --compressed -H 'content-type: application/json' -H 'authorization: Bearer xxxxxxxxxxxxxxxxxxxx' -H 'origin: moz-extension://be8e9b8a-3710-4b4a-8b1e-60d0d6d72663' -H 'Connection: keep-alive' --data '{"url":"https://www.simounet.net/goaccess-pour-analyser-ses-logs-apache/"}'`
The second time, I get a 500.

j0k3r:
Second time it works too.

```
[Mon Jul 24 13:57:10 2017] 127.0.0.1:33114 [200]: /api/entries.json
[Mon Jul 24 13:57:23 2017] 127.0.0.1:33130 [200]: /api/entries.json
```

:confused:

j0k3r:
Ok the problem was still here but was catched.
I've updated your fix to be more readable and also added a test to ensure there won't be a regression on that point.

PR #3288, opened 2017-07-25
Add an add-tag-shortcut to entry view

janLo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| Fixed tickets | 0
| License       | MIT


src/Wallabag/CoreBundle/Resources/views/themes/common/Static/howto.html.twig
```diff
@@ -147,6 +147,10 @@
                                 <td>{{ 'howto.shortcuts.toggle_favorite'|trans }}</td>
                             </tr>
                             <tr>
+                                <td><code>t</code></td>
+                                <td>{{ 'entry.view.left_menu.add_a_tag'|trans }}</td>
```
j0k3r: This should better be `howto.shortcuts.add_a_tag`.
Which means a new line in each translated file.

janLo:
I thought its better to keep the files short and reuse the translations.
Otherwise it creates redundancy in the translation files, makes them longer
and introduces the risk that a function is renamed without renaming it in
the howto.

If really necessary I can change this, but its not really DRY.

j0k3r:
I agree but what if we want to add more text in the shortcut page than in the menu?

janLo:
Then we might also add this longer text as a tooltip to the menu entry? I
think its better to have the name of functionality in the howto and the
actual ui in sync. The tooltip does not provide its own function. It only
describes the one it refers to?

j0k3r:
We do not have tooltip yet in the menu so it that extra text won't be in the menu for now.
Text in the menu must be as short as possible because of the size of the menu.
In that help page, we can add more text about that functionality (what will happen after hitting the shortcut, how it will happen, etc.). That's why I prefer a different entry in the translation message.

janLo:
The text is exactly the same as in the menu. As it is with the other shortcut descriptions. Are you sure you want to duplicate it?

j0k3r:
Yes I am.

janLo:
Ok.

j0k3r:
Since you don't agree with us you are closing the PR instead of updating it?

janLo:
I accept your point of view, it's totally ok for me. I just don't want to be the one who duplicates code by copy/pasting it. You're free to take and modify my work :).

On the other hand I think the hotkeys should be handled somehow more centrally and independent from the theme. Also the solution I provided is not perfect as it triggers while the 't' is typed, causing a 't' to appear in the tag field.

Issue #3156, opened 2017-05-28
make the CLI API powerful


oupala:
### Issue details

I want to be able to set language of a user via CLI.

To be more general, I find that [FreshRSS's CLI API](/FreshRSS/FreshRSS/blob/master/cli/README.md) is really well made and could be an inspiration for Wallabag's CLI.

You can create users with many configuration options (such as language, tokens). You can delete users. You can list users. You can import and export data for a user. And you can get info from a user (admin flag, username, last activity datetime, disk usage of a user, number of articles, tags, unread, favorites...).

It would be a very good improvement if we have the same level of API in Wallabag.

### Environment

n/a

### Steps to reproduce/test case

n/a

j0k3r:
This is a major enhancement.
So I guess it won't come early.
But listing commands is a good start.

oupala:
I do agree that it is a major enhancement if it was not planned from the beginning.

But offering web hooks and cli hooks is nowadays a must-have to get integration with other tools.

Keep that in Wallabag's roadmap and I'll be happy, even if it doesn't get into Wallabag as early as tomorrow.

Rurik19:
I am creating wallabag-cli npm package, which will be able to use wallabag API from client machines console.

tcitworld:
Do we have an issue to expose config settings through the API ?

oupala:
I'm now examining and installing Wallabag 2.4.2 and the issue is still valid nowadays.

PR #3301, opened 2017-07-30
Add list users command

nclsHart:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| Fixed tickets | #3156
| License       | MIT

wallabag-list-users

### DIFF
--- a/src/Wallabag/CoreBundle/Command/ListUserCommand.php
+++ b/src/Wallabag/CoreBundle/Command/ListUserCommand.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Wallabag\CoreBundle\Command;
+
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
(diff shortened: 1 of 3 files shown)


tcitworld:
I think it shouldn't display the whole user list by default, maybe paginate it to the `x` (for instance, 100) first users. I have ~13k users on framabag, it would be better to have a `-a` option to display them all.

Also it would be nice that you could filter results (with a `LIKE` request) through an argument to this command.

j0k3r:
Thanks for adding that command.
I'm :+1: with @tcitworld about paginating & filtering.

nclsHart:
Great idea.

I added a `limit` option and a `search` argument.

 I didn't implemented a `all` option because of potential conflict with the `limit` option.

PR #3309, opened 2017-08-03
Multiple tag search, which was broken from API

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

First, the `setParameter()` were done on the same parameter which in fact just duplicated the condition in the SQL query (like `where t.label = 'test' and t.label = 'test'`).

Changed the parameter doesn't help because the query was then wrong.

Changing the way to match associated tags for an entry and it worked.

poke @Zlitus

### DIFF
--- a/src/Wallabag/CoreBundle/Repository/EntryRepository.php
+++ b/src/Wallabag/CoreBundle/Repository/EntryRepository.php
@@ -156,7 +156,7 @@ class EntryRepository extends EntityRepository
                 $entryAlias = 'e' . $i;
                 $tagAlias = 't' . $i;
 
-                // Complexe queries to ensure multiple tags is associated to an entry
+                // Complexe queries to ensure multiple tags are associated to an entry
                 // https://stackoverflow.com/a/6638146/569101
                 $qb->andWhere($qb->expr()->in(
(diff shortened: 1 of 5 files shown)


src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -152,8 +152,23 @@ public function findEntries($userId, $isArchived = null, $isStarred = null, $isP
         }
 
         if ('' !== $tags) {
-            foreach (explode(',', $tags) as $tag) {
-                $qb->andWhere('t.label = :label')->setParameter('label', $tag);
+            foreach (explode(',', $tags) as $i => $tag) {
+                $entryAlias = 'e' . $i;
+                $tagAlias = 't' . $i;
+
+                // Complexe queries to ensure multiple tags is associated to an entry
```
tcitworld: are

Zlitus:
👍

Issue #2381, opened 2016-10-06
Reorder favorites page by date of favorite action and not by date of article import


yoannsculo:
Currently, when we tag an article as favorite it is displayed on the favorite list and this list is ordered by date of article import in database.
But I have hundreds of articles in database and I don't read them in a chronological order. When I mark an article as favorite, it will not be displayed on the top of the favorite list.

I'm using the favorite list to see afterwards the most interesting articles I read since the last time I checked the page. It would be great to be able to change the order of the list to display articles by time of favorite action and not by date of article import in Wallabag.

nicosomb:
Today, this feature is not possible. 
We don't store the date of the `favorite` action. So we need to update the schema database for that.

yoannsculo:
Ah ok I understand the reason. I just had a look at the database. is_starred is a field from wallabag_entry. Too bad it's not a easy request.
Anyway, I would be very happy if we could have this feature in the future :)

yoannsculo:
Nice!!! Thank you very much! :+1:

PR #3330, opened 2017-08-23
Add starred_at field which is set when an entry is starred

franek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #2381
| License       | MIT

This field is used to sort starred entries by starred_at field instead of createdAt field which is often not relevant (i.e. RSS exposed by wallabag).

TODO
===

- [x] Add a migration script in order to set ```starred_at``` field

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/Entry.php
+++ b/src/Wallabag/CoreBundle/Entity/Entry.php
@@ -143,6 +143,15 @@ class Entry
     private $publishedBy;
 
     /**
+     * @var \DateTime
+     *
+     * @ORM\Column(name="starred_at", type="datetime", nullable=true)
+     *
+     * @Groups({"entries_for_user", "export_all"})
+     */
+    private $starredAt = null;
+
+    /**
      * @ORM\OneToMany(targetEntity="Wallabag\AnnotationBundle\Entity\Annotation", mappedBy="entry", cascade={"persist", "remove"})
      * @ORM\JoinTable
      *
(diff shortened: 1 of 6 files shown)


src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -465,6 +494,12 @@ public function timestamps()
         }
 
         $this->updatedAt = new \DateTime();
+
+        if ($this->isStarred()) {
+            $this->setStarredAt(new \DateTime());
+        } else {
+            $this->setStarredAt(null);
```
j0k3r: I would rather move that line before the `if` and then remove the `else`.

franek:
@j0k3r done. Hope it could be merged in the next release ;)

j0k3r:
Postgres doesn't seems to like your migration https://travis-ci.org/wallabag/wallabag/jobs/267957103#L670

tests/Wallabag/ApiBundle/Controller/EntryRestControllerTest.php
```diff
@@ -497,6 +499,7 @@ public function testPostArchivedAndStarredEntry()
         $this->assertSame('http://www.lemonde.fr/idees/article/2016/02/08/preserver-la-liberte-d-expression-sur-les-reseaux-sociaux_4861503_3232.html', $content['url']);
         $this->assertSame(1, $content['is_archived']);
         $this->assertSame(1, $content['is_starred']);
+        $this->assertGreaterThanOrEqual(new \DateTime($content['starred_at']), $now);
```
j0k3r: This should be in the opposite way: first the value you expect and then the expected value

franek:
\o/ tests are green ;)

tests/Wallabag/ApiBundle/Controller/EntryRestControllerTest.php
```diff
@@ -497,6 +499,7 @@ public function testPostArchivedAndStarredEntry()
         $this->assertSame('http://www.lemonde.fr/idees/article/2016/02/08/preserver-la-liberte-d-expression-sur-les-reseaux-sociaux_4861503_3232.html', $content['url']);
         $this->assertSame(1, $content['is_archived']);
         $this->assertSame(1, $content['is_starred']);
+        $this->assertGreaterThanOrEqual($now->getTimestamp(), (new \DateTime($content['starred_at']))->getTimestamp());
```
franek: I have updated this test on my last push force sync (not a good idea but I would like to avoid too many commits). I don't know your rules concerning PR (when do we need to squash our commits, ...).

j0k3r: It's ok.
Usually we let the user push commits and when everything is ok and if there are too much useless commits we ask to squash. Otherwise it's ok to have more than one commit and sometimes it's better when there are some reflexion between each commit.

PR #3368, opened 2017-10-04
CORS repare for api, fix #3367

### LINKED ISSUE #3367: CORS doesn't work?
Maybe I am wrong, but it seems that CORS in wallabag doesn't work?
OPTIONS preflight request returns an error 400. It happends because of default settings of the **nelmioCors** bundle in config.yml: 
`allow_headers: ['X-Custom-Auth']`
but we need the `Authorization` header to be allowed. 

When I changed that setting to `allow_headers: ['Authorization']` CORS request begin working.

Does somebody need 'X-Custom-Auth' header?
Should I make a PR?

Rurik19:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| Fixed tickets | #3367

### DIFF
--- a/src/Wallabag/UserBundle/Resources/views/TwoFactor/email_auth_code.html.twig
+++ b/src/Wallabag/UserBundle/Resources/views/TwoFactor/email_auth_code.html.twig
@@ -74,7 +74,7 @@
 
                 <table cellpadding="0" cellspacing="0" border="0" align="center" id="card">
                     <tr>
-                        <td style="padding: 20px;" width="96px" valign="top"><img class="image_fix" src="{{ absolute_url(asset('wallassets/themes/_global/img/logo-other_themes.png')) }}" alt="logo" title="{{ wallabag_url }}" style="width: 96px; height: 96px;" /></td>
+                        <td style="padding: 20px;" width="96px" valign="top"><img class="image_fix" src="{{ absolute_url(asset('wallassets/themes/_global/img/logo-square.svg')) }}" alt="logo" title="{{ wallabag_url }}" style="width: 96px; height: 96px;" /></td>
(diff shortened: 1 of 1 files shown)


j0k3r:
Why not updating others `allow_headers`?

j0k3r:
Could you rebase your branche again recent changes from the 2.3 branch?

Rurik19:
done

nicosomb:
@Rurik19 Do you still work on this PR.

Rurik19:
@nicosomb  for now, this last commit is enough,

app/config/config.yml
```diff
@@ -157,11 +157,11 @@ nelmio_cors:
             max_age: 3600
         '^/':
             #origin_regex: true
-            allow_origin: ['^http://localhost:[0-9]+']
+            allow_origin: ['*']
             allow_headers: ['Authorization','content-type']
             allow_methods: ['POST', 'PUT', 'GET', 'DELETE']
             max_age: 3600
-            hosts: ['^api\.']
+            #hosts: ['^api\.']
```
j0k3r: Why should we allow cors on the UI? 🤔

Rurik19: As we still don't have API endpoints for clients information, I want to get application info right from the ui, like native applications do. If you thind it is unsecuse, let us kill last commit.

j0k3r: By client info, you mean the credentials?
I don't really like that way to do. It'll work for you, but it'll also open other security issues.

Rurik19: not password/username but clientId and clientSecret

Rurik19: And then - CORS restriction works only for web apps, any local programs, even **curl** or **wget** can grab pages easily.

Issue #3286, opened 2017-07-25
2.3: reading_time in embedded can be NULL?


janLo:
### Issue details

The current 2.3 branch seems to pass the database contents directly to an api client when `/api/entries.json` is called. This causes the Android app to raise an error if no reading time is set:
![20170725_113927](https://user-images.githubusercontent.com/96146/28566445-73f0c5ce-712f-11e7-9e4f-065ac7ce2dbb.png)

Is this an error in the API or should the android app deal with that case? Is there any spec how the returned objects look like? The doc at `yourwallabag/api/doc` only describes the requests.

### Environment

* wallabag version (or git revision) that exhibits the issue:
  * revision f0cd12f
* How did you install wallabag? Via `git clone` or by downloading the package?
  * Clone in a docker container
* Last wallabag version that did not exhibit the issue (if applicable):
  * Not sure.
* php version:
  * 7.x (from alpine)
* OS:
  * Alpine-Linux
* type of hosting (shared or dedicated):
  * Docker container, behind reverse proxy
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):
  * Postgres

j0k3r:
Could you please tell us which article got a reading_time at `NULL`? It shouldn't be.

janLo:
From a tcpdump I got this:
```
....
{
    "_links": {
        "self": {
            "href": "/api/entries/922"
        }
    },
    "annotations": [],
    "content": null,
    "created_at": "2017-02-22T14:02:56+0000",
    "domain_name": null,
    "headers": null,
    "http_status": null,
    "id": 922,
    "is_archived": 0,
    "is_public": false,
    "is_starred": 0,
    "language": null,
    "mimetype": null,
    "preview_picture": null,
    "published_at": null,
    "published_by": null,
    "reading_time": null,
    "tags": [],
    "title": "Die Angst der Machteliten vor dem Volk",
    "uid": null,
    "updated_at": "2017-02-22T14:03:17+0000",
    "url": "http://www.uni-kiel.de/psychologie/mausfeld/pubs/Mausfeld_Die_Angst_der_Machteliten_vor_dem_Volk.pdf",
    "user_email": "xxx@yyy.zz",
    "user_id": 2,
    "user_name": "xxx"
}
,...
```

The db-entry is:
```
wallabag=# select * from wallabag_entry where id=922;                                                                                                                                                                                                                                                                                                                     
[...]

PR #3373, opened 2017-10-08
Changed reading_time field to prevent null values

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes 🤞 
| Documentation | no
| Translation   | no
| Fixed tickets | #3286
| License       | MIT

Not sure about Postgres migration. I don't install PG on my new laptop, so if someone can help me here 😗 ...

### DIFF
--- a/app/DoctrineMigrations/Version20171008195606.php
+++ b/app/DoctrineMigrations/Version20171008195606.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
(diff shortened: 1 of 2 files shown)


Kdecherf:
There's a chance that migration on PostgreSQL will eventually fail:
```
You can only use SET NOT NULL when the column contains no null values.
```

You may need to add an additional statement before the `ALTER TABLE` to update field `reading_time` for all rows with `null` value.

nicosomb:
Something like my last commit?

j0k3r:
I think so

Kdecherf:
Seems ok :+1:

Kdecherf:
Another note but I'm unsure: as long as you don't define any default value, any future `INSERT` in this table will need to explicitly provide a value for `reading_time`.

j0k3r:
This is the default value: https://github.com/wallabag/wallabag/pull/3373/files#diff-a7f29fe65dccca2b9551e2380b1a9340R187

PR #3390, opened 2017-10-24
Use httplug and graby 2.0

aaa2000:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| Fixed tickets | #...
| License       | MIT

Try wallabag with a branch of Graby which supports httplug

@see https://github.com/j0k3r/graby/issues/107

- [ ] Add Guzzle 6 instead of Guzzle 5
- [x] Manually create a client to handle timeout

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -34,6 +34,7 @@ class AppKernel extends Kernel
             new FOS\JsRoutingBundle\FOSJsRoutingBundle(),
             new BD\GuzzleSiteAuthenticatorBundle\BDGuzzleSiteAuthenticatorBundle(),
             new OldSound\RabbitMqBundle\OldSoundRabbitMqBundle(),
+            new Http\HttplugBundle\HttplugBundle(),
 
             // wallabag bundles
             new Wallabag\CoreBundle\WallabagCoreBundle(),
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -375,6 +375,9 @@ httplug:
     clients:
         wallabag_core:
             factory: 'wallabag_core.http_client_factory'
+            config:
+                defaults:
+                    timeout: 10
             plugins: ['httplug.plugin.logger']
         wallabag_core.entry.download_images:
             factory: 'httplug.factory.auto'
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -370,3 +370,22 @@ jms_serializer:
 sensio_framework_extra:
    router:
         annotations: false
+
+httplug:
+    clients:
+        wallabag_core:
+            factory: 'wallabag_core.http_client_factory'
+            plugins: ['httplug.plugin.logger']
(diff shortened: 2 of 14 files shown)


aaa2000:
not yet functionally tested, and there are maybe others `guzzle` reference to replace

Example: `wallabag_core.entry.download_images.client`

https://github.com/wallabag/wallabag/blob/1953a872932a63792293b4aec087880265ba89f7/src/Wallabag/CoreBundle/Resources/config/services.yml#L213

aaa2000:
Do you think that I should replace all `GuzzleHttp\Client ` references by `Httplug` ? Or should it be limited to `Graby` ? 

Are you interested by the [HttpBundle](http://docs.php-http.org/en/latest/integrations/symfony-bundle.html#web-debug-toolbar) in order to have debug information in the web debug toolbar?

aaa2000:
We can't use Httpbundle only on dev, but the profiling is automatically turned off when `kernel.debug = false`. I prefer using always `httplug` because it will be homogeneous in the code and be useful to debug.

aaa2000:
I checked that the pocket import is OK, it seems that it is the only import impacted.

I also upgrade the code to be compatible with Graby 2.0 https://github.com/j0k3r/graby/commit/b5d8ad48c7d01505c53c80382ccf11f2acb90148

I didn't check that the articles behind a paywall can be retrieved

@j0k3r In order to use Guzzle 6 instead of Guzzle 5, it is necessary to convert BD\GuzzleSiteAuthenticator\Guzzle\AuthenticatorSubscriber in Guzzle5 middleware https://github.com/guzzle/guzzle/blob/master/UPGRADING.md#migrating-to-middleware or maybe a Http Plugin http://docs.php-http.org/en/latest/plugins/build-your-own.html. I will try with a HTTP plugin

j0k3r:
If you can succeed in converting it in a http plugin, it should be a better solution.
Otherwise, since I think we are the only one to use that middleware, converting it to Guzzle 6 can be a fallback.
We can do whatever we want: https://github.com/wallabag/guzzle-site-authenticator
Maybe it should be better to put the solution in the same repo (if it's possible)

kour1er:
Just wondering if there's been any progress on this addition - I was thinking it might be a good workaround for [#3656](https://github.com/wallabag/wallabag/issues/3656) to proxy to an American ip?

j0k3r:
I think it's time to review that PR.
Few things:
- won't jump to Guzzle 6 for now because the paywall plugin only works with Guzzle 5 atm (we'll see that later)
- Graby 2.0 isn't release yet but it's almost done, will be release before the 2.4.0 version

j0k3r:
HUGE thanks @aaa2000 :pray: :clap:

Issue #2053, opened 2016-05-09
Links without titles can't be opened


jlnostr:
### Issue details

The article below provides no title. This means I can it only open by accessing the internal URL directly.
### Steps to reproduce/test case
- Save https://github.com/Fody/PropertyChanged/wiki/NotificationInterception
- Try to open it from the main page

j0k3r:
Works fine for me: http://v2.wallabag.org/view/746

jlnostr:
Yep, by typing the url.

> This means I can it only open by accessing the internal URL directly.

But there's no way to open it from the main page. Or is there another way instead of clicking the title to open an article?
![image](https://cloud.githubusercontent.com/assets/5391936/15176955/e820a85a-176d-11e6-9e1e-e11d057c97dd.png)

nicosomb:
Is that a bug in wallabag or in graby?

(about the https://github.com/Fody/PropertyChanged/wiki/NotificationInterception link, I managed to save it and the title is available).

j0k3r:
@nicosomb it needs investigation. If the title isn't found it should display "No title found"

Zlitus:
Maybe the title is found but empty?

Kdecherf:
Well, I have 'some' cases on current master:
![2017-11-18-160039_1920x1080](https://user-images.githubusercontent.com/226063/32981820-a83d1f0c-cc7a-11e7-864a-8019c6215a16.png)

Kdecherf:
The issue could be linked to wallabagger as I don't see it when adding links via the web interface.

PR #3442, opened 2017-11-26
Fix empty title and domain_name when exception is thrown during fetch

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Documentation | no
| Translation   | no
| Fixed tickets | #2053 
| License       | MIT

Entries will now have a default title when it's empty (*because of an exception during the fetch*):

- basename part of entry's path, if any
- or domain name

This default title does not replace the "No title found" of Graby when it fails to extract a title.

This PR still needs a test on `EntryControllerTest` using a mock.

### DIFF
--- a/tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
+++ b/tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
@@ -6,6 +6,7 @@ use Tests\Wallabag\CoreBundle\WallabagCoreTestCase;
 use Wallabag\CoreBundle\Entity\Config;
 use Wallabag\CoreBundle\Entity\Entry;
 use Wallabag\CoreBundle\Entity\SiteCredential;
+use Wallabag\CoreBundle\Helper\ContentProxy;
 
 class EntryControllerTest extends WallabagCoreTestCase
 {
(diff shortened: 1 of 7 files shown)


tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
```diff
@@ -1429,4 +1430,55 @@ public function testRestrictedArticle()
 
         $client->getContainer()->get('craue_config')->set('restricted_access', 0);
     }
+
+    /**
+     * This test will require an internet connection.
```
j0k3r: Are you sure?

tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
```diff
@@ -1429,4 +1430,55 @@ public function testRestrictedArticle()
...
+        $form = $crawler->filter('form[name=entry]')->form();
+
+        $data = [
+            'entry[url]' => $url,
+        ];
+
+        /**
+         * We generate a new client to be able to use Mock ContentProxy
+         * Also we reinject the cookie from the previous client to keep the
+         * session.
```
j0k3r: Wow hard stuff :hushed:

tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php
Kdecherf: Nice one, will fix it

nicosomb:
Could you just edit `CHANGELOG.md` please? (new process ;-) )

PR #3449, opened 2017-12-02
Share page improved (mobile + header's padding)

Simounet:
Few changes to the `share` page. I added a doctype and a meta viewport. I reordered head infos and generalized the padding to everything in the page except the preview image.
❄️


Simounet:
Before:
wallabag-share-page-before
After:
wallabag-share-page-after-1
wallabag-share-page-after-2

Issue #3457, opened 2017-12-04
List view is broken on mobile


nicosomb:
See the screenshot: 

![img_0226](https://user-images.githubusercontent.com/121870/33553156-e9f582de-d8f7-11e7-868d-366b32ca60af.PNG)

On the classic view, all is OK. 
On the list view on desktop, all is OK.

wallabag version: master branch
iOS 11.1.2
Browser: Safari.

I can reproduce it with Firefox on desktop with the adaptive view.

nicosomb:
Fixed thanks to @Simounet.

PR #3459, opened 2017-12-04
Pagination improved

Simounet:
Hello there,
I fixed a wrong behavior when the pagination was on in the list mode. I improved the pagination handling (bigger space on small screens) and fixed the wrong page number alignment.
wallabag-pagination-before
wallabag-pagination-after


nicosomb:
👍 
It seems that your PR fixes https://github.com/wallabag/wallabag/issues/3457.

nicosomb:
About the desktop view, all is still OK with your changes?

Simounet:
Yep, I checked both views on desktop and mobile.

Simounet:
If you want. I thought about it but I didn't modified it. What's the point of living this theme around? I know you might have good memories with it but it's not so handy to maintain multiple officials themes. It might deserves another topic.

j0k3r:
I don't know, maybe we can then say "hey you can switch to a different theme" :slightly_smiling_face: 
+1 for an other topic

PR #3463, opened 2017-12-06
Footer improved

Simounet:
Howdy,
I fixed the issue with the `line-height` on Firefox and the white space after the footer on Firefox for Android. I removed the ellipsis on the footer's stats and it is always displayed, even on mobile. This is not shown on the entry view anyway and I think it's a cool stuff to display. :)

Before:
wallabag-firefox-footer-before

After:
wallabag-firefox-footer-after


Simounet:
Ah ah, nice joke! :man_dancing: 
I've got few other things to do in my list. I hope I'm not bothering you with my little PR. I think this is easier to validate than a "big" one.

j0k3r:
I totally agree!

Issue #3480, opened 2017-12-12
Problem upgrading from 2.2.3 to 2.3.0


markwaters:
### Issue details

I am upgrading from a new installation of 2.2.3 to latest 2.3.0
Seeing this error on the website.
500: Internal Server Error
An exception has been thrown during the rendering of a template ("An exception occurred while executing 'SELECT w0_.id AS id_0 FROM `wallabag_entry` w0_ WHERE w0_.user_id = ? AND w0_.is_starred = 1 GROUP BY w0_.id ORDER BY w0_.starred_at DESC' with params [1]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'w0_.starred_at' in 'order clause'").

On the command line , make update shows some errors.

Migrating up to 20171125164500 from 0               
                                                           
  ++ migrating 20160401000000
                                   
     -> CREATE TABLE wallabag_craue_config_setting (name VARCHAR(255) NOT NULL, value VARCHAR(255) DEFAULT NULL, section VARCHAR(255) DEFAULT NULL, UNIQUE INDEX UNIQ_5D9649505E237E06 (name), PRIMARY KEY(name))
 DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;
Migration 20160401000000 failed during Execution. Error An exception occurred while executing 'CREATE TABLE wallabag_craue_config_setting (name VARCHAR(255) NOT NULL, value VARCHAR(255) DEFAULT NULL, section V
ARCHAR(255) DEFAULT NULL, UNIQUE INDEX UNIQ_5D9649505E237E06 (name), PRIMARY KEY(name)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;':

SQLSTATE[42S01]: Base table or view already exists: 1050 Table 'wallabag_craue_config_setting' already exists
08:45:47 ERROR     [console] Error thrown while running command "doctrine:migrations:migrate --no-interaction --env=prod". Message: "An exception occurred while executing 'CREATE TABLE wallabag_craue_config_se
tting (name VARCHAR(255) NOT NULL, value VARCHAR(255) DEFAULT NULL, section VARCHAR(255) DEFAULT NULL, UNIQUE INDEX UNIQ_5D9649505E237E06 (name), PRIMARY KEY(name)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unic
ode_ci ENGINE = InnoDB;':

[...]

PR #3487, opened 2017-12-12
Change the way to check for initial migration

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | :confused: 
| Fixed tickets | #3480
| License       | MIT

Instead of checking if some migrations were run we just check if the main table exist.
- if it exists, we can skip the initial migration
- if not, run the initial migration

Poke @aaa2000

### DIFF
--- a/app/DoctrineMigrations/Version20160401000000.php
+++ b/app/DoctrineMigrations/Version20160401000000.php
@@ -4,20 +4,36 @@ namespace Application\Migrations;
 
 use Doctrine\DBAL\Migrations\AbstractMigration;
 use Doctrine\DBAL\Schema\Schema;
+use Doctrine\DBAL\Schema\SchemaException;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Initial database structure.
  */
-class Version20160401000000 extends AbstractMigration
+class Version20160401000000 extends AbstractMigration implements ContainerAwareInterface
 {
+    /**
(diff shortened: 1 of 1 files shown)


aaa2000:
I don't understand why  the migration_versions table is empty.  Do you have an idea ? 

The initial migration will not failed but the next migration will fail, no ?

aaa2000:
Maybe, I understand. The next migration use ` $this->skipIf(..`, so if all migrations use the same behaviour, the migration_versions table will be empty.

https://github.com/wallabag/wallabag/blob/f209798368de66b4acd8b892fe71018ebb30c2da/app/DoctrineMigrations/Version20160410190541.php#L32

app/DoctrineMigrations/Version20160401000000.php
```diff
@@ -4,20 +4,36 @@
...
+    }
+
     /**
      * @param Schema $schema
      */
     public function up(Schema $schema)
     {
-        if ($this->version->getConfiguration()->getNumberOfExecutedMigrations() > 0) {
-            $this->version->markMigrated();
+        try {
```
aaa2000: You can mayble use `$schema->hasTable($this->getTable('entry'));` instead of `$schema->getTable($this->getTable('entry'));`

j0k3r:
:point_up: that doc linking is a mistake.

j0k3r:
@aaa2000 you're right, sh*t

Issue #3484, opened 2017-12-12
link share with "nothing"


foxmask:
### Issue details

when you disable all the shares from the "article" part of the page http://domain/settings
the link "share" should not be displayed at all

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.3.0
* How did you install wallabag? Via `git clone` or by downloading the package? make update
* Last wallabag version that did not exhibit the issue (if applicable):
* php version:7
* OS:debian 9
* type of hosting (shared or dedicated):dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):PostgreSQL

### Steps to reproduce/test case

disable all the shares from the "article" part of the page http://domain/settings
then open an article, and click on the "share" link, nothing happened.
That give the feeling something is broken.
May be we could drop the share link when we disable any share from the settings ?

PR #3492, opened 2017-12-12
Fix 3484: hide the share toggle button if no service is enabled

bdunogier:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | aye
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #3484
| License       | MIT

Hides the "Share" toolbar entry completely if all the share services are disabled.

I'm not really happy with the big "if" block, but I'd say it'd be overdoing it to change it without having a more modulare sharing system (e.g. be able to add/remove supported services). The list is hardcoded in the template, it can be hardcoded in the condition for it.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
@@ -97,6 +97,15 @@
             <div class="collapsible-body"></div>
         </li>
 
+        {% if craue_setting('share_public')
+           or craue_setting('share_twitter')
+           or craue_setting('share_shaarli')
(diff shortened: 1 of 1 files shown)


nicosomb:
It looks OK for me. 
If you want to improve that, you can maybe create a Twig function. What do you think?

bdunogier:
Yeah, it would be one way to do it. As said in the description, I think it's fine as it is. But you're the maintainers, you tell me :)

nicosomb:
If you can take some time to do it and if you want to do it, let's go. I can merge right now if you prefer ;-)

PR #3500, opened 2017-12-14
WIP on automatic migrations.

iammichiel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | 
| License       | MIT

Hi,

The whole point of this PR is to be able, to run doctrine-migrations on every request. It can be usefull for those who don't have a ssh access to their servers. 

To enable the behaviour all one has to do, is edit the `wallabag.yml` file and set the `run_migration_on_request` to `true`.

### DIFF
--- a/app/config/wallabag.yml
+++ b/app/config/wallabag.yml
@@ -23,6 +23,7 @@ wallabag_core:
     reading_speed: 1
     cache_lifetime: 10
     action_mark_as_read: 1
+    run_migration_on_request: false
     list_mode: 0
     fetching_error_message_title: 'No title found'
     fetching_error_message: |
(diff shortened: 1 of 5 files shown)


iammichiel:
Displaying it on an admin page might actually not work. For example, if you declare a new field on the `User` entity. You might not be able to display that page if the migrations are not up to date. Doctrine will crash complaining about the field missing in database.

This was actually a solution for people who don't have ssh access on their servers and need to run the migrations manually. This is a simpler solutions. And it is definitely disabled by default. 

As for the performance overhead, it is actually a single SQL request when everything is up to date.

Issue #3503, opened 2017-12-14
Database error during first time installation


virse:
Hey there,

I get an Nginx error when I try to use the forum, so I hope it is okay to ask for help...

I try to do a fresh installation of Wallabag. I get the following errors, when the database should be filled/created:

### Issue details
```
  [RuntimeException]                                                           
  The command "doctrine:migrations:migrate" generates some errors:             
                                                                               
                                                                               
                      Application Migrations                                   
                                                                               
                                                                               
  Migrating up to 20171125164500 from 0                                        
                                                                               
    ++ migrating 20160401000000                                                
                                                                               
       -> CREATE TABLE wallabag_craue_config_setting (name VARCHAR(255) NOT N  
  ULL, value VARCHAR(255) DEFAULT NULL, section VARCHAR(255) DEFAULT NULL, UN  
  IQUE INDEX UNIQ_5D9649505E237E06 (name), PRIMARY KEY(name)) DEFAULT CHARACT  
  ER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;                         
       -> CREATE TABLE `wallabag_entry` (id INT AUTO_INCREMENT NOT NULL, user  
  _id INT DEFAULT NULL, title LONGTEXT DEFAULT NULL, url LONGTEXT DEFAULT NUL  
  L, is_archived TINYINT(1) NOT NULL, is_starred TINYINT(1) NOT NULL, content  
   LONGTEXT DEFAULT NULL, created_at DATETIME NOT NULL, updated_at DATETIME N  
  OT NULL, mimetype LONGTEXT DEFAULT NULL, language LONGTEXT DEFAULT NULL, re  
  ading_time INT DEFAULT NULL, domain_name LONGTEXT DEFAULT NULL, preview_pic  
  ture LONGTEXT DEFAULT NULL, is_public TINYINT(1) DEFAULT '0', INDEX IDX_F4D  
  18282A76ED395 (user_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLAT  
  E utf8_unicode_ci ENGINE = InnoDB;                                           
       -> CREATE TABLE wallabag_entry_tag (entry_id INT NOT NULL, tag_id INT   
  NOT NULL, INDEX IDX_C9F0DD7CBA364942 (entry_id), INDEX IDX_C9F0DD7CBAD26311  
[...]

Issue #3486, opened 2017-12-12
Wallabag 2.3.0 DB prefix issue


GaneshKandu:
### Issue details

When i try to install wallabag by changing "database_table_prefix" it throws following error :
1 . There is no table with name 'walla230.prewb_entry' in the schema.

I have attached log file to debug.

[prod.log](https://github.com/wallabag/wallabag/files/1551025/prod.log)

### Environment

* wallabag version : 2.3.0
* How did you install wallabag? Via `git clone` or by downloading the package? wget [https://wllbg.org/latest-v2-package](https://wllbg.org/latest-v2-package) && tar xvf latest-v2-package ( mentioned on [https://doc.wallabag.org/en/admin/installation/installation.html#on-shared-hosting](https://doc.wallabag.org/en/admin/installation/installation.html#on-shared-hosting) )
* php version: 5.6.23
* OS: Apache/2.2.31 (Unix) Server
* type of hosting (shared or dedicated):
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MySQL

### Steps to reproduce/test case

* Steps which i have followed
	* Unziped wallabag 2.3.0 package iin sub-directory
	* Created blank database for wallabag
	* Manually filled the database details in "app/config/parameters.yml" with "prewb_" as table prefix
	* Given writable permission to var folder.
	* Then run this command "php bin/console wallabag:install --env=prod"
	* Installer says "It appears that your database already exists. Would you like to reset it? (yesno) [no]:"
	* After hitting "no". Getting error which I mentioned above

![untitled2](https://user-images.githubusercontent.com/12833130/33878581-d9244282-df51-11e7-808d-0948c25e57a9.png)
![untitled](https://user-images.githubusercontent.com/12833130/33878582-d961459c-df51-11e7-9bd5-36d3fc875b16.png)

nicosomb:
@aaa2000 any idea?

aaa2000:
PR #3504 should  resolve that.

nicosomb:
@GaneshKandu Could you try the @aaa2000's fix please?

GaneshKandu:
Hi,
Wallabag 2.3.0 database prefix issue is solved working fine
but getting issue with font getting 404 error created new issue #3518

PR #3504, opened 2017-12-14
Fix initial migration when using custom table prefix

aaa2000:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #3503, #3486
| License       | MIT

The initial migration doesn't take into account of the `database_table_prefix` then the next migration fails.

But the postgresql sequence are not prefixed, is it a problem ? How to fix that ?


app/DoctrineMigrations/Version20160401000000.php
```diff
@@ -29,42 +29,44 @@ public function up(Schema $schema)
...
-CREATE UNIQUE INDEX UNIQ_1D63E7E592FC23A8 ON "wallabag_user" (username_canonical);
-CREATE UNIQUE INDEX UNIQ_1D63E7E5A0D96FBF ON "wallabag_user" (email_canonical);
-CREATE UNIQUE INDEX UNIQ_1D63E7E5C05FB297 ON "wallabag_user" (confirmation_token);
-CREATE TABLE wallabag_annotation (id INTEGER NOT NULL, user_id INTEGER DEFAULT NULL, entry_id INTEGER DEFAULT NULL, text CLOB NOT NULL, created_at DATETIME NOT NULL, updated_at DATETIME NOT NULL, quote VARCHAR(255) NOT NULL, ranges CLOB NOT NULL, PRIMARY KEY(id), CONSTRAINT FK_A7AED006A76ED395 FOREIGN KEY (user_id) REFERENCES "wallabag_user" (id) NOT DEFERRABLE INITIALLY IMMEDIATE, CONSTRAINT FK_A7AED006BA364942 FOREIGN KEY (entry_id) REFERENCES "wallabag_entry" (id) NOT DEFERRABLE INITIALLY IMMEDIATE);
-CREATE INDEX IDX_A7AED006A76ED395 ON wallabag_annotation (user_id);
-CREATE INDEX IDX_A7AED006BA364942 ON wallabag_annotation (entry_id);
+                $sql = <<<SQL
+CREATE TABLE ${databaseTablePrefix}craue_config_setting (name VARCHAR(255) NOT NULL, value VARCHAR(255) DEFAULT NULL, section VARCHAR(255) DEFAULT NULL, PRIMARY KEY(name));
```
j0k3r: Why don't you use `getTable()` (defined at the end of that class)?

aaa2000: I think because that I believed that could not use method in heredoc (I was wrong) and there are not only table.

nicosomb:
@aaa2000 do you need some help here?

PR #3525, opened 2017-12-21
Fix empty card title link

Simounet:
Hello there,
Long time annoying (me) fix with empty title on not fetched content. The only way (I know) to make it fetched again was to get the previous item and change manually the URL. Now, I using an existing translation to put a default title.
Before:
wallabag-title-before

After:
wallabag-title-after

I couldn't resist sharing the card and tags templates. No differences with that.

🎄


src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/Card/_content.html.twig
```diff
@@ -0,0 +1,15 @@
+<div class="card-content">
+    {% if withPreview is defined %}
+        <i class="grey-text text-darken-4 activator material-icons right">more_vert</i>
+    {% endif %}
+    <a href="{{ path('view', { 'id': entry.id }) }}" title="{{ entry.title| striptags | e('html_attr') }}" class="card-title dot-ellipsis dot-resize-update">
+        {{ entry.title | striptags | truncate(80, true, '…') | raw | default('config.form_rules.faq.variable_description.title'|trans) }}
```
nicosomb: Do we need to create a new translation key or use this one?

j0k3r: I think we should create a new translation key :+1:

Simounet: Ok, I'll do that with the same translations.

Simounet: Because there is a lot of files to modify, I would like having a validation before editing everything. I thought about adding this to the `entry` section, like this:
```yml
entry:
    default_title: "xxxxxxxx"
```
Is it ok for you?

j0k3r: It's ok for me

Simounet:
I created the new key but I encountered some strange translation files (da, fa, ro) with no translation for the previous key I used so I put it as comments.

nicosomb:
Do we need to update `entry.html.twig`?

PR #3526, opened 2017-12-22
Added random feature

### LINKED ISSUE #2285: Randomized order by default of unread articles
### Issue details

Let's say I have 24 unread articles in my Wallabag. First, I only see the 12 of them, the ones I add last. Meaning there are more chances I'll read the 12 most recent than the others. Then I read 5 of them and I add 6 new articles. The oldest article will be on page 3. It's clear that it will be more and more forgotten because it's not on the main page.  
Now, let's say I have 500 unread articles. The article on page 28 (on 42) will never be read because new ones will come and so on... (plus, I can't click on "page 28" because the option is simply not here, the only solution is to click "next next next" or type "28" in the address bar which is not a good solution imho)

I think that the possibility to have a randomized order option is important. With this kind of option, there are less chances an article can be forgotten forever because it has a chance to appear on the main page. (I'm not talking about the "read a random article" button of Wallabag v1)

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.0.8
* How did you install wallabag? Via `git clone` or by downloading the package?
* php version:
* OS:
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):

Please provide necessary steps for reproduction of this issue, or better the
reduced test case (without any external dependencies, if possible).

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| Fixed tickets | #2285
| License       | MIT

Fix #2285 

Need help about tests, because I don't know how I can test a random result.

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/EntryController.php
+++ b/src/Wallabag/CoreBundle/Controller/EntryController.php
@@ -2,6 +2,7 @@
 
 namespace Wallabag\CoreBundle\Controller;
 
+use Doctrine\ORM\NoResultException;
 use Pagerfanta\Adapter\DoctrineORMAdapter;
 use Pagerfanta\Exception\OutOfRangeCurrentPageException;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\Cache;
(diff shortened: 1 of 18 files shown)


j0k3r:
That PR doesn't solves the referenced issue.
It redirects the user to a random article from starred, unread, etc.

j0k3r:
@Kdecherf :wave:

Kdecherf:
Here's a (non-blocking) review of the SQL part:

The current behavior fetches min and max entry id for the user and the given status (_e.g. unread entries of user 1_) and gives a random id number between these two numbers. This behavior may favor some entries on others regarding their id and the possible holes in the table (_deleted entries, entries from other users_). Also the randomness may be weak for users with a few entries.

Here's a example raw SQL query which returns a random entry insensitive to table holes:

``` sql
with w1_ as (select w1_.title, ROW_NUMBER() OVER (ORDER BY id) "rn" from "wallabag_entry" w1_ WHERE w1_.user_id = 1 AND w1_.is_archived = false)                                                                                    
select w1_.title from w1_ join (SELECT ceil(random() * count(w0_.id)) "rn" FROM "wallabag_entry" w0_ WHERE w0_.user_id = 1 AND w0_.is_archived = false) w0_ using(rn);
```

This query basically uses the window feature of PostgreSQL to give an incremental number to each row regarding the current query. It joins the result to a subquery which gives a random row number (_caped to the count of entries for the user and the given status_).

Sadly MySQL does not offer a native way to do the same thing so we may have difficulty to write an equivalent query.

Finally, this query may lead to a greater query time compared to the former behavior but it seems to remain acceptable: I'm seeing a 2 ms delay for 650 entries.

Kdecherf:
Here's the pgsql query plan of my query:

```
                                                                             QUERY PLAN                                                                                                                                                                                         
--------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Hash Join  (cost=314.83..331.01 rows=3 width=32) (actual time=1.694..2.564 rows=1 loops=1)
   Hash Cond: ((w1_.rn)::double precision = (ceil((random() * (count(w0_.id))::double precision))))
   CTE w1_
     ->  WindowAgg  (cost=166.01..177.31 rows=646 width=65) (actual time=0.931..1.475 rows=646 loops=1)
           ->  Sort  (cost=166.01..167.62 rows=646 width=57) (actual time=0.923..1.117 rows=646 loops=1)
                 Sort Key: w1__1.id
                 Sort Method: quicksort  Memory: 101kB
                 ->  Index Scan using idx_entry_archived on wallabag_entry w1__1  (cost=0.28..135.85 rows=646 width=57) (actual time=0.028..0.607 rows=646 loops=1)
                       Index Cond: (is_archived = false)
                       Filter: ((NOT is_archived) AND (user_id = 1))
   ->  CTE Scan on w1_  (cost=0.00..12.92 rows=646 width=40) (actual time=0.934..1.860 rows=646 loops=1)
   ->  Hash  (cost=137.50..137.50 rows=1 width=8) (actual time=0.580..0.581 rows=1 loops=1)
         Buckets: 1024  Batches: 1  Memory Usage: 9kB
         ->  Aggregate  (cost=137.47..137.49 rows=1 width=8) (actual time=0.572..0.572 rows=1 loops=1)
               ->  Index Scan using idx_entry_archived on wallabag_entry w0_  (cost=0.28..135.85 rows=646 width=4) (actual time=0.017..0.466 rows=646 loops=1)
                     Index Cond: (is_archived = false)
                     Filter: ((NOT is_archived) AND (user_id = 1))
 Planning time: 0.324 ms
 Execution time: 2.725 ms
```

j0k3r:
I'll go in a really different way. Instead of killing ourself with finding the best way to do, I'll go in a much simpler way: fetch all entries id, random select one in PHP and fetch that entry.
It'll use the index to retrieve the data so I guess it might become slow if user got hundred of millions of entries, which is almost impossible 😨

tcitworld:
A solution I found is
```sql
SELECT Max(e.id) FROM entity e;
$randomId = rand(1, $maxId);
SELECT e FROM entity e WHERE e.id > $randomId LIMIT 1;
```

j0k3r:
Same problem as @Kdecherf explained https://github.com/wallabag/wallabag/pull/3526#issuecomment-437685336

Kdecherf:
This review is still relevant:

> This icon may mislead the user to think that it will re-sort the page. Maybe we should move this icon in the top menu bar **with an icon like a magic wand** annotated 'go to a random entry'.

lgtm apart that.

j0k3r:
I’ll let you find the material icon then 😉

tcitworld:
`swap_calls`, `all_inclusive`, `whatshot` are ideas.

We can also have a label behind a kebab menu icon.

j0k3r:
I like `swap_calls`
image

I already added a label when you override the icon:
> Jump to a random entry from that list

Kdecherf:
> `swap_calls`, `all_inclusive`, `whatshot` are ideas.
> 
> We can also have a label behind a kebab menu icon.

My preference goes to `all_inclusive`, `whatshot` maybe.
I think that `swap_calls` is too similar to `shuffle` and it may be an issue if we use the latter for another feature.

`explore` may be a good alternative, I think.

tcitworld:
There's also good old `casino`, which is what we used before. It doesn't imply that we're sorting the list in any way.

image png 96 x 96 pixels

Also maybe we could just find another place for the icon so it changes its meaning?

Kdecherf:
Any thoughts, @notabene @gnppn?

PR #3541, opened 2017-12-28
Nav actions updated

Simounet:
Hi,
Long time no see. I'm pretty sure you missed my UI PR. 🐰
I fixed and issue with the `new entry` and `search` inputs alignment and I replaced the label icons of these two actions by button so we can submit forms without en `return` key. It's good for accessibility, mobile or mouse mode. I also simplified the boxes into the nav for better alignment. I gave some love to the nav mobile view, reducing the buttons' margin so we have more space for inputs.
I worked on the JS actions to share the toggle/close methods between `new entry` and `search`.

## Desktop
Before:
add-entry-before

After:
add-entry-after

## Mobile
Before:
add-entry-mobile-before

After:
add-entry-mobile-after

⛄️


Simounet:
Hello,
Is the WIP label for me or for reviewers?

j0k3r:
Hum I think it's because you changed html stuff and then some tests are now failing (those which retrieve information from the html to validate the page rendering is ok)

Simounet:
Ok, I didn't get that. I'll fix that.

Simounet:
I can be wrong but I looked at the failing tests but I don't think that's my fault.

PR #3546, opened 2018-01-03
add `set -eu` to update.sh

### LINKED ISSUE #2957: Update fails, but continues the update process
### Issue details

When doing an update and a step in the update script `scripts/update.sh` fails, the script does not abort. It continues to run. 

excerpt of `scripts/update.sh`:
```
SYMFONY_ENV=$ENV $COMPOSER_COMMAND install --no-dev -o --prefer-dist
php bin/console doctrine:migrations:migrate --no-interaction --env=$ENV
php bin/console cache:clear --env=$ENV
```

When the script fails in the first line due to my special composer installation, it just goes on to the next line and continues the update. So far nothing worse has happend to my wallabag database, but i fear this can harm. And also it is annoying to see the error and to wait until the script finishes.

My suggestion is to `set -e` to the bash script. And while we are in the script, we can also introduce `set -u`. 
Does that make sense to you?

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.2.1
* php version:
* OS:
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):

`make update` and composer.phar being not in the typical locations.

Strubbl:
fixes #2957

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | 
| Deprecations? | no
| Tests pass?   | 
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #2957
| License       | MIT

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -2,6 +2,9 @@
 # You can execute this file to update wallabag
 # eg: `sh update.sh prod`
 
+set -e
+set -u
+
 COMPOSER_COMMAND='composer'
 
 DIR="${BASH_SOURCE}"


Kdecherf:
Thanks 👍

PR #3553, opened 2018-01-07
Swap entry url with origin url if graby provides an updated one

### LINKED ISSUE #3529: Added link is different than the original one
:warning: If your issue is about an error during fetching a link, please read: http://doc.wallabag.org/en/user/errors_during_fetching.html#how-can-i-help-to-fix-that

### Issue details

Content fetching alters the link user added thus the source address stored in reading list may end up being completely different than what did the user originally intended.

For example:
* Add https://www.reddit.com/r/linux/comments/7lgnho/llvm_501_release/ to Wallabag
* http://lists.llvm.org/pipermail/llvm-announce/2017-December/000076.html gets added instead

fetched content is content of the article posted to reddit (in other words not the comments page the user submitted to Wallabag). This may or may not be the desirable outcome for fetched text but I don't want to discuss this as this is not the real issue here.

I think the real issue here is that the original link the user has submitted is lost, even if the user clicks the original link he'll go to the content page instead of reddit discussion page.

I checked that link in https://f43.me/feed/test and the result is same. I don't know if this is the intended behavior of the fetcher or a bug, but I think it is another issue which should not distract from this one. 

I think that the original link submitted by the user should not be changed under any circumstances regardless of what result comes from fetching.

Thanks

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.2.3
* How did you install wallabag? Via `git clone` or by downloading the package? Package
* Last wallabag version that did not exhibit the issue (if applicable): not checked
* php version: 5.6.30
* OS: Debian 8.0
* type of hosting (shared or dedicated): Dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): SQLite

provided above

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | _sort of_
| New feature?  | _sort of_
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #3529 
| License       | MIT

Fixes #3529 

This PR is a partial fix for #3529: it still set the final url as the actual entry url, but now it sets origin_url field to the former url.

Ex:
Current behavior:
- User saves a reddit.com to Wallabag
- Graby fetchs the page and finds a new link to follow (llvm.org)
- Entry is saved with llvm.org as the entry url

| Given url | Fetched url | Stored url | Stored origin url |
| - | - | - | - |
| `https://www.reddit.com/r/li[…]` | `http://lists.llvm.org/pip[…]` | `http://lists.llvm.org/pip[…]` | - |

New behavior:
- User saves a reddit.com to Wallabag
- Graby fetchs the page and finds a new link to follow (llvm.org)
- Entry is saved with llvm.org as the entry url and reddit.com as the origin url

| Given url | Given origin url | Fetched url | Stored url | Stored origin url | Diff |
| - | - | - | - | - | - |
| `https://www.reddit.com/r/li[…]` | - | `http://lists.llvm.org/pip[…]` | `http://lists.llvm.org/pip[…]` |  `https://www.reddit.com/r/li[…]` | url |
| `https://www.lemonde.fr/tiny[…]` | - | `https://www.lemonde.fr/m-deco[…]` | `https://www.lemonde.fr/m-deco[…]` | - | lemonde regex |
| `http://www.example.org/blah[…]` | - | `https://www.example.org/blah[…]` | `https://www.example.org/blah[…]` | - | scheme |
| `https://www.medium.com/hello[…]` | - | `https://www.medium.com/hello#world[…]` | `https://www.medium.com/hello[…]` | - | fragment
| `https://www.mylink.com/foo[…]` | `https://twitter.com[…]` | `https://www.anotherlink.com/bar[…]` | `https://www.anotherlink.com/bar[…]` | `https://twitter.com[…]` | origin url already set |
| `http://feedproxy.google.com/~r/Wa[…]` | - | `https://example.org/hello-wallabag[…]` | `https://example.org/hello-wallabag[…]` | - | host ignore list |

Here is the hardcoded ignore list for specific patterns : 
- `feedproxy.google.com`
- `feeds.reuters.com`
- `https?://www\.lemonde\.fr/tiny.*`

Todo
- [x] leave origin_url unchanged if difference is an ending slash
- [x] leave origin_url unchanged if difference is a fragment
- [x] leave origin_url unchanged if new value matches the ignore list

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -68,9 +68,8 @@ class ContentProxy
 
         // In one case (at least in tests), url is empty here
         // so we set it using $url provided in the updateEntry call.
-        // Not sure what are the other possible cases where this property is empty 
-        if (empty($entry->getUrl()) && !empty($url))
-        {
+        // Not sure what are the other possible cases where this property is empty
+        if (empty($entry->getUrl()) && !empty($url)) {
             $entry->setUrl($url);
         }
 
@@ -247,7 +246,15 @@ class ContentProxy
      */
     private function stockEntry(Entry $entry, array $content)
     {
-        $entry->setUrl($content['url']);
+        // When a redirection occurs while fetching an entry
+        // we move the original url in origin_url property if empty
+        // and set the entry url with the final value
+        if (!empty($content['url']) && $entry->getUrl() !== $content['url']) {
+            if (empty($entry->getOriginUrl())) {
+                $entry->setOriginUrl($entry->getUrl());
+            }
+            $entry->setUrl($content['url']);
+        }
 
         $this->setEntryDomainName($entry);
 
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -328,6 +328,18 @@ class ContentProxy
         $parsed_entry_url = parse_url($entry->getUrl());
         $parsed_content_url = parse_url($url);
 
+        /**
+         * The following part computes the list of part changes between two
+         * parse_url arrays.
+         *
+         * As array_diff_assoc only computes changes to go from the left array
+         * to the right one, we make two differents arrays to have both
+         * directions. We merge these two arrays and sort keys before passing
+         * the result to the switch.
+         *
+         * The resulting array gives us all changing parts between the two
+         * urls: scheme, host, path, query and/or fragment.
+         */
         $diff_ec = array_diff_assoc($parsed_entry_url, $parsed_content_url);
         $diff_ce = array_diff_assoc($parsed_content_url, $parsed_entry_url);
 
@@ -340,6 +352,17 @@ class ContentProxy
             return false;
         }
 
+        /**
+         * This switch case lets us apply different behaviors according to
+         * changing parts of urls.
+         *
+         * As $diff_keys is an array, we provide arrays as cases. ['path'] means
+         * 'only the path is different between the two urls' whereas
+         * ['fragment', 'query'] means 'only fragment and query string parts are
+         * different between the two urls'.
+         * 
+         * Note that values in $diff_keys are sorted.
+         */
         switch ($diff_keys) {
             case ['path']:
                 if (($parsed_entry_url['path'] . '/' === $parsed_content_url['path']) // diff is trailing slash, we only replace the url of the entry
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -246,15 +246,7 @@ class ContentProxy
      */
     private function stockEntry(Entry $entry, array $content)
     {
-        // When a redirection occurs while fetching an entry
-        // we move the original url in origin_url property if empty
-        // and set the entry url with the final value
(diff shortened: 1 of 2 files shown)


Kdecherf:
Well, after using this PR for few days, there's a new behavior for which I don't know if it'll annoy users:

I fetch news using a rss feed reader and share articles that I'm interested in into wallabag (Android share).

Some RSS providers like Le Monde and Slashdot use dedicated rss links which redirects to the actual content.
Until now the initial link (rss.slashdot.org for slashdot, lemonde.fr/tiny for Le Monde) is not kept. But with this PR this former link is now visible, e.g.:

2018-01-13-150229_1920x1080

Kdecherf:
New change: this PR does not override `origin_url` if it's not empty

Kdecherf:
Hey there,

Regarding my last comment I think I will add a custom parameter (in `parameters.yml`) to ignore some domains like feedproxy.google.com, rss.slashdot.org. Toughts @wallabag/core?

j0k3r:
Please, re-explain 🙂 
I've re-read you few last comments, and I don't get it.

Kdecherf:
This PR fills empty `origin_url` with former url in case of redirection (`single_page_link` or HTTP redirection).

When you use a RSS reader and share articles to Wallabag, they often come from an intermediate url, e.g.:
- PetaPixel articles have feedproxy.google.com/~r/PetaPixel/ as former url
- LeMonde.fr articles have lemonde.fr/tiny as former url

Imo it's somewhat annoying and not very useful. That's why I suggest adding an ignore list in `parameters.yml` to prevent these links to be set as `origin_url`.

j0k3r:
Can you rebase your branch against the master?

Kdecherf:
I consider this PR ready for review. I'll update the PR message with a table to explain the different behaviors introduced with it.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -305,6 +312,90 @@ private function stockEntry(Entry $entry, array $content)
...
+            $diff_ce = array_diff_assoc($parsed_content_url, $parsed_entry_url);
+
+            $diff = array_merge($diff_ec, $diff_ce);
+            $diff_keys = array_keys($diff);
+            sort($diff_keys);
+
+            if ($this->ignoreUrl($entry->getUrl())) {
+                $entry->setUrl($url);
+            } else {
+                switch ($diff_keys) {
```
j0k3r: Could you explain that `swtich`? It's a bit unclear what is expected here and also what the `case`s means.

j0k3r:
Be careful while removing query parameters.
If you have `site.com?pageid=123` and `site.com?pageid=456` they should be considered as two separate entry and not a single one.

src/Wallabag/CoreBundle/Helper/ContentProxy.php
Kdecherf: I'll add a comment to explain this part

tcitworld:
Maybe ignored hosts and patterns should be inside a dedicated file and not hardcoded ?

Kdecherf:
@tcitworld that was my original thought but I decided to go faster by hardcoding it in the first version. @j0k3r are you still okay with that?

j0k3r:
Yeah we agree to hardcode them in 2.3 and to add them in the database in the 2.4 (because we'll need a migration. And also, people will be able to manage them

Issue #3571, opened 2018-01-22
update.sh uses BASH_SOURCE to determine directory


Strubbl:
### Issue details

In [update.sh in line 10 ](https://github.com/wallabag/wallabag/blob/master/scripts/update.sh#L10) we use BASH_SOURCE to try to find the directory of update.sh script:

```bash
DIR="${BASH_SOURCE}"
```

But according to documentation, this variable is an array.

> BASH_SOURCE
>
>    An array variable whose members are the source filenames where the corresponding shell function names in the FUNCNAME array variable are defined. The shell function ${FUNCNAME[$i]} is defined in the file ${BASH_SOURCE[$i]} and called from ${BASH_SOURCE[$i+1]}

https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#index-BASH_005fSOURCE

If we want to find the directory of the update.sh script, we should follow a different approach, e.g.:

```bash
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```

https://stackoverflow.com/a/246128/709697

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.3.2
* How did you install wallabag? Via `git clone` or by downloading the package?git
* Last wallabag version that did not exhibit the issue (if applicable):2.3.1
* php version:
* OS:debian
* type of hosting (shared or dedicated):dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):postgresql

roots84:
I suppose I already saw an issue resulting from this. While trying to upgrade the setup found a missing PHP extension: 

> roots@roots-server /var/www/x22889/wallabag $ sudo make update
> [sudo] password for roots:
> composer.phar not found, we'll see if composer is installed globally.
> remote: Counting objects: 777, done.
> remote: Compressing objects: 100% (69/69), done.
> remote: Total 777 (delta 456), reused 499 (delta 444), pack-reused 255
> Receiving objects: 100% (777/777), 154.48 KiB | 0 bytes/s, done.
> Resolving deltas: 100% (510/510), completed with 151 local objects.
> From https://github.com/wallabag/wallabag
>  * [new branch]      2.4        -> origin/2.4
>  * [new branch]      add-random-article -> origin/add-random-article
>  + e4dcc1b...9d5f120 change-locale-register -> origin/change-locale-register  (forced update)
>  * [new branch]      fix-migrations-dash -> origin/fix-migrations-dash
[...]

PR #3572, opened 2018-01-22
call set -eu only after env vars are tested

Strubbl:
This commit fixes a race condition where we search for variable
BASH_SOURCE, but it is not set. In that case the update.sh would abort.
But it should not abort, because in the following lines we are going to
set DIR variable by another way.
To avoid this case where BASH_SOURCE is not set, the `set -u` command is
moved after the code, where we expect all variables to be valid.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/
| New feature?  | /no
| BC breaks?    | /no
| Deprecations? | /no
| Tests pass?   | yes/no
| Documentation | /no
| Translation   | /no
| CHANGELOG.md  | /no
| Fixed tickets | #...
| License       | MIT

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -2,15 +2,14 @@
 # You can execute this file to update wallabag
 # eg: `sh update.sh prod`
 
-set -e
-
 COMPOSER_COMMAND='composer'
 
 DIR="${BASH_SOURCE}"
 if [ ! -d "$DIR" ]; then DIR="$PWD/scripts"; fi
 . "$DIR/require.sh"
 
-# starting from here, update shall abort if a variable is not set
+# starting from here, update shall abort if a variable is not set or command errors
+set -e
 set -u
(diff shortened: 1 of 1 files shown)


Strubbl:
The questions to the reviewer are: 

- After set -e, is any command allowed to fail and the update.sh shall continue?
- After set -u, is it okay if any of the following variables is not set?

kissgyorgy:
This is a very bad idea IMO, because it doesn't solve the problem, just hides it. The BASH_SOURCE should be set a default value should be used so `set -u` would not fail.
It is NOT a race condition at all (it's a totally different thing), but the problem is that the shell which is running is not actually bash when running through Makefile. See my comments here: https://github.com/wallabag/wallabag/issues/3571#issuecomment-399228821

Strubbl:
Yes @kissgyorgy , you are right. thank you

Issue #720, opened 2014-06-05
Renaming tags


tcitworld:
Renaming tags. 
How to do it : 
- create a new tag for new name if not already exists
- assign all of the articles for the old tag name to the new one
- delete old tag

This should also allow merging two different tags in a single one.

shulard:
Hello guys, I've just faced this issue and I'll be glad to help adding this feature.

I haven't contributed to Wallabag at the moment but I'll try to do my best here 😄.

shulard:
Hello guys,

I've just installed a Wallabag instance on my dev environment. Have you already discussed implementation details here ?

I've some ideas but I need help to refine them :

* I think adding an "edit" icon in tag tag list on the right of each label ;
* When user click on that label two options :
  * Create an inline field in place of the label and update the edit icon to a save one ;
  * Go to a new page where the field is displayed and the tag renamed (like the edit icon on the view article page).

Any though here ?
Maybe I'm not really clear, please ask questions if it's needed 😄.

nicosomb:
I'd prefer an inline field. 
@j0k3r @tcitworld @Kdecherf any opinion?

Kdecherf:
Go for the inline way 👍

shulard:
Hello,

I've started working on the implementation (#3574). I've searched inside the codebase and found a lot of server side logic (clic -> action -> reload -> display). Is there a "standard" way to handle the inline field generation inside JavaScript ? 

Do you prefer having a `<form>` inside each tag element that can be submitted or handle it with an Ajax request ?

j0k3r:
Not at the moment.

ewoks:
@shulard what are missing steps that need to be completed? Do you still need some feedback to make decision? Is there any specific blocker?

Kdecherf:
@ewoks a review has been requested, trying to do it today :)

timokau:
@j0k3r why the close? Has this been implemented or was it given up on?

shulard:
@timokau, this was implemented in the #3574 MR.

timokau:
Great, thank you!

PR #3574, opened 2018-01-23
Allow to rename tags from the web interface.

shulard:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| CHANGELOG.md  | no
| Fixed tickets | #720
| License       | MIT

This PR will add the ability to rename tags. As discussed in #720 the process to rename is tag is :

* Create a new tag with the new label;
* Attach the new tag to all the entries attached to the old one;
* Drop the old one.

We need to discuss further about UX implementation.


shulard:
Hello,

I've finished to implement the tag renaming process. Actually it works only inside the "material" theme but it can be implemented easily in the other one.

It use a form to allow the user to edit the label. Then it's submitted as a POST request and handled with a Symfony form.

I've some questions about implementation details :

* I've added a class and some style on the form element, I hope it was the right place to do it 😄 ;
* I use a single RenameTagType form instance to generate all the rename tag form in the list view, is it ok for you ?
* I've added some translations in the fr / en dictionaries, do I need to add them on all the languages ?
* JavaScript interactions is actually defined in the `tags.html.twig` file and injected in the footer. I can't find the right file to add my functions, is it ok for you ?

shulard:
Also, if I need to add some details in the documentation please tell me where !

src/Wallabag/CoreBundle/Controller/TagController.php
```diff
@@ -130,4 +132,46 @@ public function showEntriesForTagAction(Tag $tag, $page, Request $request)
...
+            foreach ($entries as $entry) {
+                $this->get('wallabag_core.tags_assigner')->assignTagsToEntry(
+                    $entry,
+                    $form->get('label')->getData()
+                );
+            }
+
+            $em = $this->getDoctrine()->getManager();
+            $em->remove($tag);
+            $em->flush();
```
j0k3r: Be careful, a tag can be associated to entries from other users.
IMHO you shouldn't delete it from the database but rather deassociate it from each entry of the current logged in user.

src/Wallabag/CoreBundle/Resources/views/themes/material/Tag/tags.html.twig
```diff
@@ -13,7 +13,16 @@
         <ul class="card-tag-labels">
             {% for tag in tags %}
                 <li title="{{tag.label}} ({{ tag.nbEntries }})" id="tag-{{ tag.id }}">
-                    <a href="{{ path('tag_entries', {'slug': tag.slug}) }}" class="card-tag-link">{{tag.label}} ({{ tag.nbEntries }})</a>
+                    <a href="{{ path('tag_entries', {'slug': tag.slug}) }}" class="card-tag-link">
+                        {{tag.label}}&nbsp;({{ tag.nbEntries }})
```
j0k3r: I know it was already there like this, but I think `{{ tag.label }}` (adding spaces around) might be better

j0k3r:
> I've added some translations in the fr / en dictionaries, do I need to add them on all the languages ?

You should add them in other langages too (add the english one and comment the line)

>  I can't find the right file to add my functions, is it ok for you ?

I think you can add them here: `app/Resources/static/themes/_global/index.js`

> Also, if I need to add some details in the documentation please tell me where !

About the doc, you can update that file https://github.com/wallabag/doc/blob/master/fr/user/tags.md

src/Wallabag/CoreBundle/Controller/TagController.php
shulard: Ok, I've updated the code to remove the tag from associated entries.

src/Wallabag/CoreBundle/Resources/views/themes/material/Tag/tags.html.twig
shulard: Ok, done 😄.

shulard:
Hello @j0k3r,

I've updated the translation files to add lines even if they are commented. I also moved my JavaScript code inside `index.js` file and added compatibility with the baggy theme.

I think that the code part is done for me 😄, I'll create another PR for the documentation on the specific repository.

j0k3r:
Looks like tests are failing, you should fix them.
And also run `php bin/php-cs-fixer fix` to fix CS issue.

shulard:
I've updated the CS issues and fixed the tests. Travis status will validate these updates.

ewoks:
Hi @j0k3r , any progress on this? Looks like it is ready for merge, but stayed stale. Can you take a look? thanks

ewoks:
friendly poke to @j0k3r please don't forget review

Kdecherf:
@ewoks I will review it in the following days

j0k3r:
Thanks @shulard :clap:

PR #3606, opened 2018-03-18
add manifest.json for android pwa

mritzmann:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | untested
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | no
| License       | MIT

---

[PWA](https://developers.google.com/web/progressive-web-apps/) would be an enrichment for Android users who only want to use the WebApp. Later, this feature will also be available [[1](https://medium.com/@kennethrohde/progressive-web-apps-coming-to-all-chrome-platforms-80e31272e2a8)] [[2](https://www.androidpolice.com/2018/02/21/google-improving-progressive-web-apps-chrome-chrome-os/)] in Chrome Desktop.

---

- [x] add manifest.json with settings for android
- [x] linkt to manifest.json in html meta
- [x] add icon with [192x192px or bigger for splash screen](https://developers.google.com/web/updates/2015/10/splashscreen)


src/Wallabag/CoreBundle/Resources/views/base.html.twig
```diff
@@ -38,6 +38,8 @@
 
             <link rel="shortcut icon" type="image/x-icon" href="{{ asset('wallassets/themes/_global/img/appicon/favicon.ico') }}">
 
+            <link rel="manifest" href="/manifest.json">
```
j0k3r: Better use:

```twig
<link rel="manifest" href="{{ asset('manifest.json') }}">
```

mritzmann: Thank you! I corrected it.

mritzmann:
In my opinion, works well.

* When adding to the home screen, the name is correct https://i.imgur.com/R5h5s9b.png
* The WebApp starts with a splash screen https://i.imgur.com/4Vmjtz5.png
* Runs in full screen https://i.imgur.com/npCGS0c.png
* Displayed in the App Drawer as a standalone app https://i.imgur.com/wGlwKIS.jpg

Now it is possible to build a version with [Service Worker](https://developers.google.com/web/ilt/pwa/introduction-to-service-worker) that also works offline... But I'm not familiar with that. This pull request is only a small change.

@j0k3r: I tried to test it with `make test`. I then terminated because it requires too many dependencies. Is that still okay, or do I still have to deal with it?

web/manifest.json
```diff
@@ -0,0 +1,48 @@
+{
+  "short_name": "wallabag",
+  "name": "Save and classify articles. Read them later. Freely.",
```
tcitworld: Ideally this would be translated.

j0k3r: Yeah but it'll be hard to translate it inside the json.

j0k3r:
Thanks @mritzmann and sorry for the delay!

PR #3614, opened 2018-03-25
Spelling: GitHub, Log out, of the dev

comradekingu:
| Q             | A
| ------------- | ---
| Translation   | yes
| License       | MIT

Not completely sure what "Thought of the developers" alludes to.
Maybe it should be just "Thought of developers" ?

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/translations/messages.en.yml
+++ b/src/Wallabag/CoreBundle/Resources/translations/messages.en.yml
@@ -265,7 +265,7 @@ about:
     who_behind_wallabag:
         developped_by: 'Developed by'
         website: 'website'
-        many_contributors: 'And many others contributors ♥ <a href="https://github.com/wallabag/wallabag/graphs/contributors">on Github</a>'
+        many_contributors: 'And many others contributors ♥ <a href="https://github.com/wallabag/wallabag/graphs/contributors">on GitHub</a>'
         project_website: 'Project website'
         license: 'License'
         version: 'Version'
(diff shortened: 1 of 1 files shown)


src/Wallabag/CoreBundle/Resources/translations/messages.en.yml
```diff
@@ -323,7 +323,7 @@ howto:
         go_import: Go to import
         go_developers: Go to developers
         go_howto: Go to howto (this page!)
-        go_logout: Logout
+        go_logout: Log out
```
tcitworld: If I understand https://whitelightconcepts.com/login-vs-log-in-logout-vs-log-out/ correctly, I think it should stay in one word.

comradekingu: How is it used in wallabag? You click the button and what happens?
"Login" makes sense for talking about the process of logging in, the portal as it were. Clicking to do something however, is two words. I can't grasp logging out being more complex than clicking the "Log out"-button…?

Kdecherf: There's no intermediate page when you click on this link, you're directly logged out. 'Log out' seems more accurate here.

comradekingu: @tcitworld It is indicative of an action occuring as a result of clicking it, so "Log out" applies.
You got to the logout page (or "Log out"-page) and click "Log out" to perform an action.

PR #3653, opened 2018-05-24
Fix/firefox mobile unneeded resize

Simounet:
Hi there,
If you use wallabag and Firefox Mobile, you know what this means. If you don't, open wallabag on Firefox Mobile and go up and down on an entry page.
This PR fix that.
The problem was the resize method calling an animation that was stopping the scrolling move.
Cheers.


j0k3r:
👍

PR #3684, opened 2018-07-04
Fix image downloading on null image path

Simounet:
Hi,
Image path could be null resulting in this error `PHP Warning:  imagecreatefromstring(): Data is not in a recognized format in /wallabag/src/Wallabag/CoreBundle/Helper/DownloadImages.php on line 121`.
Encountered with https://framablog.org/2018/06/30/engagement-atypique/ .
Have a nice day.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/DownloadImages.php
+++ b/src/Wallabag/CoreBundle/Helper/DownloadImages.php
@@ -85,6 +85,10 @@ class DownloadImages
      */
     public function processSingleImage($entryId, $imagePath, $url, $relativePath = null)
     {
+        if (null === $imagePath) {
+            return false;
+        }
+
         if (null === $relativePath) {
             $relativePath = $this->getRelativePath($entryId);
         }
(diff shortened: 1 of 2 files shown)


src/Wallabag/CoreBundle/Helper/DownloadImages.php
```diff
@@ -85,6 +85,9 @@ public function processHtml($entryId, $html, $url)
      */
     public function processSingleImage($entryId, $imagePath, $url, $relativePath = null)
     {
+        if (null === $imagePath) {
+            return false;
+        }
```
j0k3r: Could you add an extra empty line below?

Simounet:
I agree but this method shouldn't handle null strings anyway. I'll investigate a little bit further.

Simounet:
It's coming from the `DownloadImagesSubscriber` when `$entry->getPreviewPicture()` returns `null`. https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Event/Subscriber/DownloadImagesSubscriber.php#L59
I could add a test if no preview image detected.

j0k3r:
Thanks for the investigation!
It's ok with what you did.
Could you just fix the CS? https://travis-ci.org/wallabag/wallabag/jobs/400059586#L1031

Simounet:
I updated it.

PR #3690, opened 2018-07-11
More robust srcset image attribute handling

Simounet:
Hello,
Linked to HTMLawed PR https://github.com/kesar/HTMLawed/pull/17


src/Wallabag/CoreBundle/Helper/DownloadImages.php
```diff
@@ -193,9 +193,11 @@ protected function getSrcsetUrls(Crawler $imagesCrawler)
         while ($iterator->valid()) {
             $srcsetAttribute = $iterator->current()->getAttribute('srcset');
             if ('' !== $srcsetAttribute) {
-                $srcset = array_map('trim', explode(',', $srcsetAttribute));
+                $pattern = "/(?:[^\"'\s]+\s*(?:\d+[wx])+)/";
```
j0k3r: Could you add a comment to basically explain what it does?
I'm not regex fluent 🙂

Simounet:
Yep, we'll. It seems like cloudinary using commas into URL frequently thought. :/

j0k3r:
Can this be merged without waiting for your other PR to be merged?

Simounet:
Yep, there is no grip between those two. We will do a better job on wallabag with it but it doesn't change the fack that HTMLawed will break some links…

j0k3r:
Ok great. I forgot to ask for a unit test on that with case which aren't handled by the previous code. Can you add some?

Simounet:
I did it with a rebase as a cherry on the cake. 🍰

Issue #3633, opened 2018-04-18
Web UI completely blocked with: The string to escape is not a valid UTF-8 string.


kedare:
### Issue details

When accessing the Wallabag web interface, instead of the actual interface, it shows

500: Internal Server Error
The string to escape is not a valid UTF-8 string.

### Environment

* wallabag version (or git revision) that exhibits the issue: Docker image wallabag/wallabag:2.3.2
* How did you install wallabag? Via `git clone` or by downloading the package? Official docker image
* Last wallabag version that did not exhibit the issue (if applicable):
* php version: PHP 7.1.12 (cli) (built: Dec 27 2017 14:24:15) ( NTS ) (From docker image)
* OS: Docker (Fedora host)
* type of hosting (shared or dedicated): Docker over dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): SQLite

### Steps to reproduce/test case

Access the web interface ? I don't know exactly what could cause this, this was working fine before, I know it started after a day I added a few new pages to wallabag, maybe it's related to that ? (A page containing invalid encoding could break the whole web UI?)

When accessing the web interface showing this error (any page basically), prod.log shows this :

```
[2018-04-18 19:26:18] request.INFO: Matched route "homepage". {"route":"homepage","route_parameters":{"_controller":"Wallabag\\CoreBundle\\Controller\\EntryController::showUnreadAction","page":1,"_route":"homepage"},"request_uri":"http://wallabag.xxx.net/","method":"GET"} []
[2018-04-18 19:26:18] security.DEBUG: Read existing security token from the session. {"key":"_security_secured_area"} []
[2018-04-18 19:26:18] security.DEBUG: User was reloaded from a user provider. {"username":"kedare","provider":"Symfony\\Bridge\\Doctrine\\Security\\User\\EntityUserProvider"} []
[2018-04-18 19:26:18] request.CRITICAL: Uncaught PHP Exception Twig_Error_Runtime: "The string to escape is not a valid UTF-8 string." at /var/www/wallabag/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig line 43 {"exception":"[object] (Twig_Error_Runtime(code: 0): The string to escape is not a valid UTF-8 string. at /var/www/wallabag/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig:43)"} []
[...]

PR #3725, opened 2018-09-18
Bugfix: Sanitize the title of a saved webpage from invalid UTF-8 characters.

Tobi823:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes (my changes doesn't cause any new failures, 8013f35d96c42b15c1da28cdff40e97289ad4e25 already cause 2 Errors)
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #3633
| License       | MIT

The issue described in https://github.com/wallabag/wallabag/issues/3633 was caused by <code>e('html_attr')</code> in [_content.html.twig](https://github.com/wallabag/wallabag/blob/e586d65b64089fc1cc230a18c470aae3f45f91a6/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/Card/_content.html.twig#L5). By replacing it with <code>e('html')</code> is a quick but not secure solution.

The real issue was caused by the library j0k3r/graby returning the title on a webpage in different encodings (like UTF-8 or ISO-8859-1). But wallabag always treaded the title as UTF-8 without converting the encoding.

For some webpages (like [https://www.impots.gouv.fr/portail/files/media/10_conventions/espagne/espagne_convention-avec-l-espagne-impot-sur-le-revenu-impot-sur-la-fortune_fd_1824.pdf](https://www.impots.gouv.fr/portail/files/media/10_conventions/espagne/espagne_convention-avec-l-espagne-impot-sur-le-revenu-impot-sur-la-fortune_fd_1824.pdf)) graphy returns the title (like "Convention avec l'Espagne - Impôt sur le revenu - Impôt sur la fortune - 1995") in the ISO-8859-1 encoding (can also be Windows-1252 because they are very simmilar).

But <code>ô</code> has the value 0xF4 [ISO_8859-1 table](https://de.wikipedia.org/wiki/ISO_8859-1#ISO/IEC_8859-1). For UTF-8 the value 0xF4 for a single character is invalid (because of the [algorithm](https://en.wikipedia.org/wiki/UTF-8#Description) of UTF-8).

My solution is to test if the title of a webpage contains only valid UTF-8 characters. If not, the title is treaded as ISO_8859-1, converted to UTF-8 and checked for invalid UTF-8 characters. If the titel still contains invalid UTF-8 characters, then the invalid characters will be replaced.

I hope that my code is good enough. I dont have much experience in open-source contributions, I'm not an expert in PHP and have no knowledge about Laravel.

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -53,6 +53,7 @@ class ContentProxy
 
         if ((empty($content) || false === $this->validateContent($content)) && false === $disableContentUpdate) {
             $fetchedContent = $this->graby->fetchContent($url);
+            $fetchedContent['title'] = $this->sanitizeUTF8Text($fetchedContent['title']);
 
             // when content is imported, we have information in $content
             // in case fetching content goes bad, we'll keep the imported information instead of overriding them
@@ -66,6 +67,28 @@ class ContentProxy
         $content['url'] = !empty($content['url']) ? $content['url'] : $url;
 
         $this->stockEntry($entry, $content);
+    }
+
+    /**
+     * Remove invalid UTF-8 characters from the given string in following steps:
+     * - try to interpret the given string as ISO-8859-1, convert it to UTF-8 and return it (if its valid)
+     * - simply remove every invalid UTF-8 character and return the result (https://stackoverflow.com/a/1433665)
+     * @param String $rawText
+     * @return string
+     */
+    private function sanitizeUTF8Text(String $rawText) {
+        if (mb_check_encoding($rawText, 'utf-8')) {
+            return $rawText; // return because its valid utf-8 text
+        }
+
+        // we assume that $text is encoded in ISO-8859-1 (and not the similar Windows-1252 or other encoding)
+        $convertedText = utf8_encode($rawText);
+        if (mb_check_encoding($convertedText, 'utf-8')) {
+            return $convertedText;
+        }
+
+        // last resort: simply remove invalid UTF-8 character because $rawText can have some every exotic encoding
+        return iconv("UTF-8", "UTF-8//IGNORE", $rawText);
     }
 
     /**
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -90,8 +90,8 @@ class ContentProxy
      * @return string (maybe contains invalid UTF-8 character)
      */
     private function convertPdfEncodingToUTF8($title) {
-        // first try UTF-16 (then UTF-8) because its easier to detect its present/absence
-        foreach (array('UTF-16BE', 'UTF-16LE', 'UTF-8', 'WINDOWS-1252') as $encoding) {
+        // first try UTF-8 because its easier to detect its present/absence
+        foreach (array('UTF-8', 'UTF-16BE', 'WINDOWS-1252') as $encoding) {
             if (mb_check_encoding($title, $encoding)) {
                 return mb_convert_encoding($title, 'UTF-8', $encoding);
             }
--- a/src/Wallabag/CoreBundle/Helper/ContentProxy.php
+++ b/src/Wallabag/CoreBundle/Helper/ContentProxy.php
@@ -76,7 +76,7 @@ class ContentProxy
      * @param String $rawText
      * @return string
      */
-    private function sanitizeUTF8Text(String $rawText) {
+    private function sanitizeUTF8Text($rawText) {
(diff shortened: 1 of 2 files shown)


src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -68,6 +69,28 @@ public function updateEntry(Entry $entry, $url, array $content = [], $disableCon
...
     }
 
+    /**
+     * Remove invalid UTF-8 characters from the given string in following steps:
+     * - try to interpret the given string as ISO-8859-1, convert it to UTF-8 and return it (if its valid)
+     * - simply remove every invalid UTF-8 character and return the result (https://stackoverflow.com/a/1433665)
+     * @param String $rawText
+     * @return string
+     */
+    private function sanitizeUTF8Text(String $rawText) {
```
j0k3r: Don't type parameter, it comes with PHP 7, wallabag should still be compatible with PHP 5

Tobi823:
I've removed the type parameter. The code should be compatibel with PHP5.

I gain further understanding in the issue:
- Websites are not effected by the bug #3725 because nearly all are encoding in UTF-8.
- PDFs are the problem because they are often encoded in WINDOWS-1252 or in other encoding (as far as I know is UTF-16 possible). 

I add code for trying to detect the encoding (first check for UTF-8, then UTF-16, then WINDOWS-1252). These order is necessary because the "smiling cat face with heart-shaped eyes" encoded in UTF-8 is valid UTF-16 (颻). And text encoded in UTF-8 is also valid WINDOWS-1252.

The underlying PdfParser https://github.com/smalot/pdfparser does not (or cant?) return the encoding and that why we have to guess it. This solution is not perfect but at least wallabag stays usable (when the title of saved websites contains some strange characters9.

(For testing I use special unicode character like 😻 or ℤ instead of chinese or russian.)

Tobi823:
Cant the Travis CI build be restarted? I think the EntryControllerTest (line 200) failed because the website http://www.liberation.fr/planete/2017/04/05/donald-trump-et-xi-jinping-tentative-de-flirt-en-floride_1560768 was slow/down.

j0k3r:
Could you instead rebase your branch against the master? I fixed few tests

src/Wallabag/CoreBundle/Helper/ContentProxy.php
```diff
@@ -68,6 +69,48 @@ public function updateEntry(Entry $entry, $url, array $content = [], $disableCon
...
 
+    /**
+     * Try to sanitize the title of the fetched content from wrong character encodings and invalid UTF-8 character.
+     * @param $title
+     * @param $contentType
+     * @return string
+     */
+    private function sanitizeContentTitle($title, $contentType) {
+        if ('application/pdf' === $contentType) {
+            $convertedTitle = $this->convertPdfEncodingToUTF8($title);
```
j0k3r: You should instead override the `$title` value and use only one `return $this->sanitizeUTF8Text($title);`:

```php
if ('application/pdf' === $contentType) {
    $title = $this->convertPdfEncodingToUTF8($title);
}

return $this->sanitizeUTF8Text($title);
```

tests/Wallabag/CoreBundle/Helper/ContentProxyTest.php
```diff
@@ -531,6 +531,242 @@ public function testWithImageAsContent()
...
+        $expectedTitle = 'F09F98BB' . 'E284A4' . '7A';
+        $this->assertSame($expectedTitle, $this->strToHex($entry->getTitle()));
+    }
+
+    /**
+     * https://stackoverflow.com/a/18506801
+     * @param $string
+     * @return string
+     */
+    function strToHex($string){
```
j0k3r: You should define `strToHex` & `hexToStr` as private as you don't use them outside of this tests

Tobi823:
I rebased my fork to the current master in this repository and I override the value of the method parameter with the new PDF title value.

It should work but I'm going to test it today. (I'm currently working on my laptop without PHP, Apache etc. installed)

tests/Wallabag/CoreBundle/Helper/ContentProxyTest.php
Tobi823: Ok

Kdecherf:
Hello @Tobi823, there's one last failing test, could you please run php-cs-fixer?

Tobi823:
I've tested the code and it should work now.

j0k3r:
Could you run `php bin/php-cs-fixer fix` without any other parameters? Looks like there is still issue on the updated test file.

Tobi823:
I forgot to fix the ContentProxyTest file with php-cs-fixer.

Issue #3590, opened 2018-02-10
abort in case make update runs as root


Strubbl:
### Issue details

reported by @francis-barban in the gitter chat

![grafik](https://user-images.githubusercontent.com/97055/36061082-5eaa570c-0e55-11e8-99ae-0e02c6dff26f.png)

detect when make update runs as root and abort in that case

### Environment

* wallabag version (or git revision) that exhibits the issue:
* How did you install wallabag? Via `git clone` or by downloading the package?
* Last wallabag version that did not exhibit the issue (if applicable):
* php version:
* OS:
* type of hosting (shared or dedicated):
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL):

### Steps to reproduce/test case

Please provide necessary steps for reproduction of this issue, or better the
reduced test case (without any external dependencies, if possible).

francis-barban:
Not the real cause. Warning because security lost but runs well. Anyway you have to CHOWN your wallabag after cache clear command (which rebuild the cache files with root owner so no access for wallabag).

Strubbl:
Yes, you are right. That's the reason I opened this second issue. This issue should respect the composer's warning. Having no git installed, is a different issue and handled by #3589

PR #3733, opened 2018-10-06
Abort running install and update script if root

Deuchnord:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #3590 
| License       | MIT

This pull request makes the `make install` and `make update` scripts stop if they are run as root.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -2,6 +2,12 @@
 # You can execute this file to install wallabag
 # eg: `sh install.sh prod`
 
+# Abort running this script if root
+if [ "$EUID" == "0" ]; then
+    echo "Do not run this script as root!" >&2
+    exit 1
+fi
+
 COMPOSER_COMMAND='composer'
 
 DIR="${BASH_SOURCE}"
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -2,6 +2,12 @@
 # You can execute this file to update wallabag
 # eg: `sh update.sh prod`
 
+# Abort running this script if root
+if [ "$EUID" == "0" ]; then
+    echo "Do not run this script as root!" >&2
+    exit 1
+fi
+
 set -e
 set -u
 


j0k3r:
Maybe we should still _allow_ it by giving a flag like `--ignore-root-warning` to the make command.

I know we shouldn't use root to install wallabag (as composer says) but sometimes you don't have the choice and blocking it might no be the best solution. People might come and says they can't install because of that message.

The message should say sth like that:

> Do not run this script as root! Use `--ignore-root-warning` if you really know what you do.

Deuchnord:
Sorry, didn't see your comment.
Actually, it is a bad idea to use root to host a Web application as root, because if that application has a security issue, it could in certain circumstance lead to compromise the whole system. If you have root access, you'll probably prefer to create a separate user that contains the application, so that if it gets compromised, the whole system cannot. In this case, you don't need to run Composer with root access. I propose we add a note in the README instead to explain this (even if this is not only for Wallabag).

But if needed, we could add the given option, indeed.

Kdecherf:
To be honest I don't see any use case requiring a root access. @j0k3r do you think of something in particular?

j0k3r:
I'm thinking about a dummy user who installed it's own wallabag into an old computer to test it.
S(He) made everything using root because (s)he doesn't care. And after that change (s)he'll come here and yell at us because the script is not working anymore.
So I prefer to prevent wasting my time of explaining how to bypass that restriction or how to not use root to install wallabag, etc.

I guess composer doesn't block the install/update process when you are using root because of that kind of particular reason.

Deuchnord:
Okay, let's add this `--ignore-root-warning` argument :)

Kdecherf:
@j0k3r fair enough

j0k3r:
Also, don't forget to rebase your branch against the master to fix tests.

Deuchnord:
OK, I'm doing this this evening :)

j0k3r:
Thanks @Deuchnord!

Glandos:
Some tests were missing here. The `IGNORE_ROOT_ARG` is taken from `$1`, as well as the `ENV`. So it is impossible to run the script as root right now.

j0k3r:
@Glandos feel free to provide a PR for that

j0k3r:
@Glandos @Deuchnord can you check https://github.com/wallabag/wallabag/pull/3885?

PR #3758, opened 2018-10-24
Dropping PHP < 7.1

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | kind of
| BC breaks?    | kind of
| Deprecations? | no
| Tests pass?   | yes
| Documentation | will need to be updated
| Translation   | no
| CHANGELOG.md  | no
| Fixed tickets | #...
| License       | MIT

- Sunsetting PHP 5.6 & 7.0 👋
- Also cleanup some files.
- And using @Daniel-KM PHPePub which remove some deprecated stuff and seems to be more up to date than the _official_ version.
- Upgrade `wallabag/tcpdf` to fix all issues we have (+ fixing [CVE-2018-17057](https://nvd.nist.gov/vuln/detail/CVE-2018-17057))
- Use our own fork of CraueConfigBundle to fix utf8mb4

NelmioApiDoc bundle wasn't updated because [the upgrade guide](https://github.com/nelmio/NelmioApiDocBundle/blob/master/UPGRADE-3.0.md) doesn't explain how to migrate but provide a command that totally screw up all our annotation.

SchebTwoFactorBundle 3 might need to be updated later (to avoid too much stuff in that PR)


Kdecherf:
MySQL seems broken here

j0k3r:
While it works pretty fine on my local machine :confused:

j0k3r:
Ok problem is here: https://travis-ci.org/wallabag/wallabag/jobs/447797516#L1126 :face_with_head_bandage:

childnode:
⚠️ 
just to throw in.. as long conservative distributions like debian use a lower version in latest stable maintained release or LTS ... perhaps we should not force-update that forces users to backport packages to let some piece of software up to date?!
https://packages.debian.org/search?keywords=mod-php&searchon=names&suite=stable&section=all
https://packages.debian.org/search?keywords=php7&searchon=names&suite=stable&section=all

Ubuntu LTS trusty is stuck on 5.9
https://packages.ubuntu.com/search?suite=trusty&searchon=names&keywords=mod-php

xenial is on 7.0 while only latest bionic lts is on 7.2
referring to https://wiki.ubuntu.com/Releases they are all still valid supported, at least up to 2021 if you skip trusty!

j0k3r:
I know that some people are still running obsolete stuff.

But some PHP versions are going to be dropped in the coming months: PHP 5.6 & 7.0: http://php.net/supported-versions.php
There'll be no more official active support and no more official security support.

Following that Composer PHP stats show that PHP 7.2 is rising and all other versions are declining.
https://blog.packagist.com/php-versions-stats-2018-2-edition/

And we also need to go ahead and drop unsupported PHP versions. A lot of libraries already dropped support for PHP 5 & 7.0. Which means, if we want to use new/update/security stuff from them, we'll also need to drop them.

We don't know when wallabag 2.4.0 will be released (in the coming weeks, months, years?). It depends on how free time we can dedicate to it.

Finally it's not because we are going to release wallabag 2.4.0 that we are going to stop fixing bug in 2.3.*. I guess we'll have a 2.3 branch and we'll push fixes on it from time to time.

Finally (again), there are still alternative that people on debian use a lot to get fresh version of PHP on their server https://deb.sury.org/

nebulade:
Is there anything I could help out here? Since we at Cloudron have updated postgres, our wallabag package had to be suspended for the time being until this is fixed.

To provide some more information, might be totally obvious, but I am not very familiar with wallabag nor Doctrine/DBAL, we do see the failure in a sql query at https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Doctrine/DBAL/Schema/CustomPostgreSqlSchemaManager.php#L26
```
15:07:40 - + /usr/local/bin/gosu www-data:www-data php bin/console doctrine:migrations:migrate --env prod --no-interaction
15:07:40 -                                                               
15:07:40 -                     Application Migrations                    
15:07:40 -                                                               
15:07:40 - 
15:07:40 - Migrating up to 20181029131313 from 20160401000000
15:07:40 - 
15:07:40 -   ++ migrating 20160410190541
15:07:40 - 
15:07:40 - Migration 20160410190541 failed during Execution. Error An exception occurred while executing 'SELECT min_value, increment_by FROM "entry_id_seq"':
15:07:40 - 
15:07:40 - SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist
15:07:40 - LINE 1: SELECT min_value, increment_by FROM "entry_id_seq"
15:07:40 -                ^
15:07:40 - 14:07:40 ERROR     [console] Error thrown while running command "doctrine:migrations:migrate --env prod --no-interaction". Message: "An exception occurred while executing 'SELECT min_value, increment_by FROM "entry_id_seq"':
15:07:40 - 
15:07:40 - SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist
15:07:40 - LINE 1: SELECT min_value, increment_by FROM "entry_id_seq"
15:07:40 -                ^" ["exception" => Doctrine\DBAL\Exception\InvalidFieldNameException { …},"command" => "doctrine:migrations:migrate --env prod --no-interaction","message" => """  An exception occurred while executing 'SELECT min_value, increment_by FROM "entry_id_seq"':\n  \n  SQLSTATE[42703]: Undefined column: 7 ERROR:  column "min_value" does not exist\n  LINE 1: SELECT min_value, increment_by FROM "entry_id_seq"\n                 ^  """]
15:07:40 - 
```

As said I am a bit clueless here, but I cannot find any location where such a column would be created in the first place. Also the file header suggests, that it is indeed there to support postgres 10, so maybe this migration step is simply not finished yet? If so, is there anything I could help?

childnode:
> But some PHP versions are going to be dropped in the coming months: PHP 5.6 & 7.0: http://php.net/supported-versions.php

good hint .. I was not that informed in the present past about the php lifecycle.
ok, then perhaps we should just head them to a migration hint when releasing this change to e.g. dockered environment https://doc.wallabag.org/en/admin/installation/installation.html#installation-with-docker backing up https://doc.wallabag.org/en/admin/backup.html and restore?!

composer.json
```diff
@@ -68,7 +68,7 @@
         "friendsofsymfony/oauth-server-bundle": "^1.5.2",
         "stof/doctrine-extensions-bundle": "^1.2",
         "scheb/two-factor-bundle": "^2.14.0",
-        "grandt/phpepub": "^4.0.7",
+        "grandt/phpepub": "dev-master",
```
Kdecherf: Are you sure you're using @Daniel-KM's PHPePub?

j0k3r: Yup, because of https://github.com/wallabag/wallabag/pull/3758/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R147

Kdecherf: Oh I see, the package name don't change.

tcitworld:
I would have added php7.3 support. Do you want it here or in another PR ?

j0k3r:
@tcitworld this is planned.
I'm currently upgrading all bundle to latest version (and it tooks time to update our code to be compatible with them).

COPYING.md
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2013-2017 Nicolas Lœuillet
+Copyright (c) 2013-current Nicolas Lœuillet
```
tcitworld: Not sure if this is allowed.

app/config/parameters.yml.dist
```diff
@@ -11,8 +11,6 @@ parameters:
     # database_password: %env.database_password%
 
     database_driver: pdo_mysql
-    database_driver_class: ~
```
tcitworld: Don't forget to put in the release notes advice to admins to delete this line.

j0k3r: Yup.
But it’s not really a problem to let that value une the file as it’s not used anymore.
I’ll also update the doc

nebulade:
Thanks for pushing this forward! While trying to test this, I keep on getting the following `composer install` error:

```
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Loading composer repositories with package information
Failed to clone the git@github.com:Daniel-KM/PHPePub.git repository, try running in interactive mode so that you can enter your GitHub credentials

In Git.php line 325:
                                                                               
  Failed to execute git clone --mirror 'git@github.com:Daniel-KM/PHPePub.git'  
   '/root/.composer/cache/vcs/git-github.com-Daniel-KM-PHPePub.git/'           
                                                                               

install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<packages>]...
```

This looks like a new dependency, which needs to be built from git? I am not very familiar with composer, so maybe this is just an issue on my end, in which case, please ignore.

tcitworld:
Do you have `git` installed ?

nebulade:
Yes `git version 2.17.1` and `Composer 1.6.3 2018-01-31 16:28:17`

Kdecherf:
You must have an ssh key linked to your github account when cloning with this address.

If you can't or don't want to link your github account to the system account (`root` in this case which is not a good idea), you could use `https://github.com/Daniel-KM/PHPePub.git` instead.

nebulade:
I do not have patched the wallabag code here, this is just calling `composer install` like before these patches. Maybe I am using composer wrongly with the new dependencies?

gramakri:
@nebulade Maybe we can set `github-protocols` as suggested by https://stackoverflow.com/questions/18638395/how-to-force-composer-to-use-https-instead-of-git . Looks like composer has some built-in support for github and we prefer ssh clone by default.

j0k3r:
What did you get when you manually perform that command `git clone --mirror 'git@github.com:Daniel-KM/PHPePub.git' '/root/.composer/cache/vcs/git-github.com-Daniel-KM-PHPePub.git/'`?
Also, did you try the config global command from the SO first answer? Does it work?

nebulade:
None of those settings have worked, however it is clearly a composer relate issue itself, so not really related to this thread.

PR #3765, opened 2018-11-03
Remove preview picture from single entry view page for #1875

### LINKED ISSUE #1875: Not to show preview-image in full article view
As preview-image usual is taken from the article, in full view it duplicates and takes a lot of space. Better not to show at all or to have an option in config. Look: two pictures takes all the window
image

burkemw3:
Showing the preview picture usually leads to showing a duplicate
image, and frequently leads to showing duplicate images directly
adjacent to each other.

before

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes/no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | yes/no
| Translation   | yes/no
| CHANGELOG.md  | yes/no
| License       | MIT

Before:
screen shot 2018-11-03 at 12 48 24 pm

After:
screen shot 2018-11-03 at 12 57 02 pm

Fixes #1875

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
@@ -96,9 +96,6 @@
                 </div>
             </aside>
         </div>
-        {% if entry.previewPicture is not null %}
-            <div><img class="preview" src="{{ entry.previewPicture }}" alt="{{ entry.title|e|raw }}" /></div>
-        {% endif %}
         <article>
             {{ entry.content | raw }}
         </article>
--- a/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
@@ -275,10 +275,6 @@
                 {{ render(controller( "WallabagCoreBundle:Tag:addTagForm", { 'id': entry.id } )) }}
             </div>
 
-            {% if entry.previewPicture is not null %}
-                <div><img class="preview" src="{{ entry.previewPicture }}" alt="{{ entry.title|striptags|default('entry.default_title'|trans)|raw }}" /></div>
-            {% endif %}
-
         </aside>
         <article>
             {{ entry.content | raw }}


Kdecherf:
Hello @burkemw3,

I thought we already had a debate on how to handle this issue but I can't find it.

Removing the preview image like that may remove actual content as it is not always embedded in the fetched content.

We could consider two different options to handle that:
* compare the preview image against all images embedded in the content (_i don't see it as the easy path :stuck_out_tongue:)
* add a new keyword in the site configs to tell graby that the preview image for a given site is also embedded in the content

@wallabag/core, thoughts?

Side-note: as for now I'm handling it locally with the following css workaround:

```
#article > aside div img.preview {
        display: none;
}
```

burkemw3:
Can you be more specific with how removing the preview picture would remove rendered content? Examples would help.

As I understand it, the preview picture comes from open graph annotations embedded in article markups and those open graph annotations are never rendered by browsers. What am I missing?

burkemw3:
Any examples of removing actual content?

techexo:
> Can you be more specific with how removing the preview picture would remove rendered content? Examples would help.
> 
> As I understand it, the preview picture comes from open graph annotations embedded in article markups and those open graph annotations are never rendered by browsers. What am I missing?

I can't talk for @Kdecherf, but as I understand it you have three cases:

1. The preview picture is also used in the article. The picture is thus shown once as preview image, then again as article content.
One example that I can think of is with [Le Monde website](https://www.lemonde.fr/societe/article/2018/11/21/barrages-filtrants-et-evacuations-le-point-sur-la-cinquieme-journee-des-gilets-jaunes_5386374_3224.html). `og:image` is the same as the image used as header for the article. However, the header picture has a caption. Consequently:
    * if you strip the header picture in the site config, you also strip the caption and thus loose content;
    * if you don't strip the header picture, you have both  preview and header (identical) pictures at the beginning of the articles.

2. The preview picture is not used in the text, so not displaying it could be considered as a loss of content (even if I don't think that it would be of too much importance).

burkemw3:
> The preview picture is not used in the text ... could be considered as a loss of content

I don't understand this. If a user can't see the image when browsing the page in a modern web browser, how is it loss of content if the user can't see the image when viewing the page in wallabag?

techexo:
The preview image is sometime used as a header (without caption), like a banner. One might argue that the choice made by an editor for the banner is part of the editorial content.

Example: the banner here is not fetched in the body, but is the same as the `og:image`.
image


But I totally agree that I am splitting hairs here! ;-) IMHO, preview picture inside the article view is not that important.

Kdecherf:
@techexo you understood pretty well, thanks :)

I think we can give up on this.

burkemw3:
Great, thanks, modified the baggy theme as well.

If there's an image that's shown in a browser, but isn't getting included in the content, that sounds like a parse/site-config issue to me. I think site-config should be used to pull the desired content

Kdecherf:
Fixes #1875

PR #3772, opened 2018-11-16
Update translation when marking as read

techexo:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | yes
| CHANGELOG.md  | yes
| License       | MIT

This very slight modification is to ensure that the option is well understood.
Indeed, it indicated the destination (go to home page or stay on the current page) when one article was mark as read. However, this option is actually also controlling where you go when you mark something as favorite, or when you delete one.

I think this modification will allow one to better understand what this option is doing (I personally couldn't understand why I was taken back to the home page when marking an article as favorite).


j0k3r:
Can you update the english one too?
I haven't checked it but maybe it's almost the same

techexo:
English translation done.

PR #3798, opened 2018-12-02
Enable OTP 2FA

### LINKED ISSUE #2822: 2FA via third party apps

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | needed
| Translation   | missing
| CHANGELOG.md  | no
| License       | MIT

Fixes #2822

Moved stuff from the "reset area" to a dedicated tab. Might be better.

screen 2018-12-02 a 17 37 13

Enable OTP 2FA:
- Update SchebTwoFactorBundle to version 3
- Enable Google 2fa on the bundle
- Disallow ability to use both email and google as 2fa
- use `$this->addFlash` shortcut instead of `$this->get('session')->getFlashBag()->add`
- update admin to be able to enable/reset the 2fa

Here is a gif when enabling OTP app:

2fa

- [x] Backup codes
- [ ] Hash backup codes
- [x] Replace "Google Authenticator"
- [x] Ability to test a generated code

### DIFF
--- a/src/Wallabag/UserBundle/Controller/ManageController.php
+++ b/src/Wallabag/UserBundle/Controller/ManageController.php
@@ -8,6 +8,7 @@ use Pagerfanta\Adapter\DoctrineORMAdapter;
 use Pagerfanta\Exception\OutOfRangeCurrentPageException;
 use Pagerfanta\Pagerfanta;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Annotation\Route;
 use Wallabag\UserBundle\Entity\User;
@@ -31,10 +32,10 @@ class ManageController extends Controller
         // enable created user by default
         $user->setEnabled(true);
 
-        $form = $this->createForm('Wallabag\UserBundle\Form\NewUserType', $user);
-        $form->handleRequest($request);
+        $form = $this->createEditForm('NewUserType', $user, $request);
 
         if ($form->isSubmitted() && $form->isValid()) {
+            $user = $this->handleOtp($form, $user);
             $userManager->updateUser($user);
 
             // dispatch a created event so the associated config will be created
@@ -62,14 +63,14 @@ class ManageController extends Controller
      */
     public function editAction(Request $request, User $user)
     {
-        $deleteForm = $this->createDeleteForm($user);
-        $editForm = $this->createForm('Wallabag\UserBundle\Form\UserType', $user);
-        $editForm->handleRequest($request);
+        $userManager = $this->container->get('fos_user.user_manager');
 
-        if ($editForm->isSubmitted() && $editForm->isValid()) {
-            $em = $this->getDoctrine()->getManager();
-            $em->persist($user);
-            $em->flush();
+        $deleteForm = $this->createDeleteForm($user);
+        $form = $this->createEditForm('UserType', $user, $request);
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            $user = $this->handleOtp($form, $user);
+            $userManager->updateUser($user);
 
             $this->get('session')->getFlashBag()->add(
                 'notice',
@@ -81,7 +82,7 @@ class ManageController extends Controller
 
         return $this->render('WallabagUserBundle:Manage:edit.html.twig', [
             'user' => $user,
-            'edit_form' => $editForm->createView(),
+            'edit_form' => $form->createView(),
             'delete_form' => $deleteForm->createView(),
             'twofactor_auth' => $this->getParameter('twofactor_auth'),
         ]);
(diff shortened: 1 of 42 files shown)


virtadpt:
This is probably a dumb question, but will this interfere with existing app integrations (e.g., Android) if we upgrade to get this functionality and turn it on?

tcitworld:
@virtadpt It will indeed.

j0k3r:
@virtadpt the email 2fa is already available. You can give a try.
We added the OTP option in that PR.

j0k3r:
@Kdecherf done
@tcitworld I moved the QR code upfront. About the code validation I agree, but I don't really know how to better integrate it. I think I'll let it for later and create an issue about that.

I have question regarding the OTP secret & backup codes: should the administrator be able to view them? Like:
image

Kdecherf:
@j0k3r imo administrators must not see secret keys and backup codes. Also, backup codes must not be saved somewhere: they are shown only once.

Here is a side-note on the text of checkbox for OTP: we should avoid "Google" as well here.

j0k3r:
Backup codes are saved in the database, so we can check if the given code is valid (and then remove it from the backup codes).
Ok I agree about secret & codes from the admin.

Kdecherf:
Oh, well, I saw it differently in my mind :D You're right.

tcitworld:
> Backup codes are saved in the database, so we can check if the given code is valid (and then remove it from the backup codes).

Is there any reason why we don't hash them inside database ?

j0k3r:
@tcitworld None. We can hash backup code.
But the secret key generated for the OTP app can't be hashed. Which means, an admin can still configure its OTP app with the secret from a given user using the secret code. 😕

noplanman:
Just popping in to ask about the status on this. Would love to see this merged :blush:

j0k3r:
I need to get back on this to:
- [ ] remove info for the admin POV
- [ ] add a better OTP validation process

j0k3r:
Rebased against the updated 2.4 branch.

src/Wallabag/UserBundle/Controller/ManageController.php
```diff
@@ -62,14 +62,29 @@ public function newAction(Request $request)
...
+        if ($this->getParameter('twofactor_auth') && true === $user->isGoogleAuthenticatorEnabled() && false === $form->isSubmitted()) {
+            $form->get('googleTwoFactor')->setData(true);
+        }
+
+        if ($form->isSubmitted() && $form->isValid()) {
+            // handle creation / reset of the OTP secret if checkbox changed from the previous state
+            if ($this->getParameter('twofactor_auth')) {
+                if (true === $form->get('googleTwoFactor')->getData() && false === $user->isGoogleAuthenticatorEnabled()) {
+                    $user->setGoogleAuthenticatorSecret($this->get('scheb_two_factor.security.google_authenticator')->generateSecret());
+                    $user->setEmailTwoFactor(false);
```
noplanman: Would it be an option to allow setting of both? Or is there a specific reason why OTP takes preference over email?

I've seen other 2FA implementations that allow OTP and U2F for example, to be able to decide which one to use.

j0k3r: Should we allow more than on OTP per account? I don't know. I decided that one take on the other one.
As far as I remember I usually choose only one OTP solution when many are available.

I'll also need to check if both can be enabled.

noplanman: For software-based 2FA it might not be necessary, but I've been super grateful for a second option when using U2F and OTP, as some browsers aren't fully U2F compatible yet. So having only U2F enabled might lock you out in some cases when not using the environment where it was set up.

It was more a question of understanding why only 1.
I think it makes sense to leave it as is and if the need should arise in the future by many users, to then have another look :+1:

j0k3r:
I've updated how it works now.
We can verify a generated code before enabling it.

Just one thing left: hashing backup codes.

j0k3r:
Regarding my [last comment](https://github.com/wallabag/wallabag/pull/3798#issuecomment-445262985) I'm wondering if we really should hash backup code in the database? 🤔

tcitworld:
> Regarding my last comment I'm wondering if we really should hash backup code in the database? 🤔

My opinion would be to check how popular libraries / softs handle this.

j0k3r:
GitHub allow you to view your backup codes. So I guess they didn't hash them.

techexo:
I have at least two services (Ubisoft and I think Microsoft) telling me when activating otp "this is your code and don't lose them". So they probably hash them.

j0k3r:
Rebased because of conflicts.

Kdecherf:
> It seems quite easy to me to have backups codes hashed, so I'd still vote for this before merging it.
> 
> The whole point of having backup codes is not being able to see them again once connected, so that if someone gets access to your session on a computer you just left, he's not able to get access to them.

I didn't think about that vector, you make a point.

Kdecherf:
I'll review and test the workflow tomorrow

PR #3816, opened 2018-12-18
Validate imported entry to avoid error on import

j0k3r:
We got some imports with a missing `url` field generating some errors while trying to retrieve an existing entry with that url (hello @Rurik19 :grimacing:).
Introducing the `validateEntry` allow us to dismiss a message when it doesn't have an url (or other missing stuff in the future)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/src/Wallabag/ImportBundle/Consumer/AbstractConsumer.php
+++ b/src/Wallabag/ImportBundle/Consumer/AbstractConsumer.php
@@ -52,6 +52,13 @@ abstract class AbstractConsumer
 
         $this->import->setUser($user);
 
+        if (false === $this->import->validateEntry($storedEntry)) {
+            $this->logger->warning('Entry is invalid', ['entry' => $storedEntry]);
+
+            // return true to skip message
+            return true;
+        }
+
         $entry = $this->import->parseEntry($storedEntry);
 
         if (null === $entry) {
--- a/src/Wallabag/ImportBundle/Import/AbstractImport.php
+++ b/src/Wallabag/ImportBundle/Import/AbstractImport.php
@@ -119,6 +119,15 @@ abstract class AbstractImport implements ImportInterface
     abstract public function parseEntry(array $importedEntry);
 
     /**
+     * Validate that an entry is valid (like has some required keys, etc.).
+     *
+     * @param array $importedEntry
+     *
+     * @return bool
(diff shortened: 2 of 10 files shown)


src/Wallabag/ImportBundle/Consumer/AbstractConsumer.php
```diff
@@ -52,6 +52,13 @@ protected function handleMessage($body)
 
         $this->import->setUser($user);
 
+        if (false === $this->import->validateEntry($storedEntry)) {
+            $this->logger->warning('Entry is invalid', ['entry' => $storedEntry]);
```
tcitworld: It would be nice to have only the URL(s) here instead of an array.

This would require every import having a `getEntryUrl()` method that returns the appropriate value for each appropriate key (`uri`/`href` and so on).

Other than that, :+1:

j0k3r: The `validateEntry` method only checks for empty url. Which means in case the url is empty I'll put an empty string in the error message :thinking:

Kdecherf: Maybe we could identify unique data for each import implementation for the logging part?
e.g. `guid` for firefox bookmarks, `meta` for pinboard export and title otherwise.

j0k3r: `$storedEntry` isn't the Entry entity, it's an array of the imported json data which means it should contains too much data (like the content of the entry for example)

tcitworld: That's why I was not fond of passing the whole array to the logger. ;)

Kdecherf: (I guess @j0k3r forgot the word "not" :D)

j0k3r: It should **not** contains :)

PR #3819, opened 2018-12-22
Fix read & starred status in Pocket import

stan3:
Would be nice to do a test for this given that a lot of it is already there, just need more assertions but haven't found the time yet so thought it would be good to at least get a fix out.

### DIFF
--- a/src/Wallabag/ImportBundle/Import/PocketImport.php
+++ b/src/Wallabag/ImportBundle/Import/PocketImport.php
@@ -206,10 +206,10 @@ class PocketImport extends AbstractImport
         $this->fetchContent($entry, $url);
 
         // 0, 1, 2 - 1 if the item is archived - 2 if the item should be deleted
-        $entry->setArchived(1 === $importedEntry['status'] || $this->markAsRead);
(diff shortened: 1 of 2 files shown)


src/Wallabag/ImportBundle/Import/PocketImport.php
```diff
@@ -194,10 +194,10 @@ public function parseEntry(array $importedEntry)
         $this->fetchContent($entry, $url);
 
         // 0, 1, 2 - 1 if the item is archived - 2 if the item should be deleted
-        $entry->setArchived(1 === $importedEntry['status'] || $this->markAsRead);
+        $entry->setArchived(1 == $importedEntry['status'] || $this->markAsRead);
```
j0k3r: Maybe the best fix is to cast the value as integer:
```suggestion
        $entry->setArchived(1 === (int) $importedEntry['status'] || $this->markAsRead);
```

j0k3r:
Rebased against the master & added tests.

j0k3r:
Thanks @stan3!

PR #3820, opened 2018-12-25
Fix incorrect reading time calculation for entries with CJK characters

### LINKED ISSUE #3821: Wrong reading time estimation for articles in CJK
### Issue details

CJK languages use no explicit delimiter, e.g. space in English. 
Therefore, current calculation scheme will treat a long sentence in CJK like `一个有十个字符的句子` as a single word, which results in extreme short reading time.

### Environment

* wallabag version (or git revision) that exhibits the issue: docker:latest
* How did you install wallabag? Via docker-compose
* php version: 7.2.13
* OS: Ubuntu 18.04
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): PostgreSQL

Bag the article at `https://sspai.com/post/47150`, and check the calculated reading time. I got `3min` on my instance, which is not quite accurate for such a long post.

lizyn:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT
(Sorry but I don't know how to do a test)

Trying to fix  #3821 . 

Sentences in CJK don't use explicit delimiters, so previous match scheme will treat sentences like `这本来应该是一句很长很长的话...` as a single word, which leads to reading time always being `< 1min`. 

As there is no easy rules to identify words in CJK sentences, it may be a good idea to treat single CJK character as a word for reading time estimation. (And that's how length calculations are usually done for CJK contents.)

In the modified match pattern in file `/src/Wallabag/CoreBundle/Tools/Utils.php`, `\p{xx}` stands for a character in:

| xx             | Language
| ------------- | ---
|  Han    | Chinese
|  Hiragana, Katakana| Japanese
|   Hangul  | Korean

**Note**: This modification should *not* influence reading time calculation for entries without CJK content.


lizyn:
@j0k3r Thanks for your help!

I'm kind of a newbie and currently don't know how to set a test environment, if you're available, could you please help with this:

Try to bag the article at https://sspai.com/post/47150 and check the reading time.

The characters in this articles, if counted in my modified way, should be around `9061` words, so the reading time would be about 45 min (200 words/min).If that's satisfied, then I think there's nothing left to fix.

(For my current instance, which runs wallabag 2.3.5 through docker-compose, the estimated reading time is 3min. And that's way too short.)

FYI, I counted the words in such ways:

pull contents from wallabag instance through API, and count it in PHP, results:
```
php > echo count(preg_split('~[^\p{L}\p{N}\']+~u', strip_tags($str)));                                                                                                                                      780  # the old way
php > echo count(preg_split('~([^\p{L}\p{N}\']+|\p{Han}|\p{Hiragana}|\p{Katakana}|\p{Hangul})~u', strip_tags($str)));
9061 # my modified way
```

lizyn:
Thanks, I think it's ready for review now, nothing wrong AFAIK, @j0k3r @Kdecherf .

j0k3r:
Thanks @lizyn

Issue #3790, opened 2018-11-27
Export all articles in JSON: link not visible


techexo:
### Issue details

The import menu is supposed to allow for the importation of one big JSON file containing all articles.
> This importer will import all your wallabag v2 articles. Go to All articles, then, on the export sidebar, click on "JSON". You will have a "All articles.json" file.

However, the Export sidebar is not visible in _All articles_ on my instance. I checked and JSON export is indeed activated on my instance, and I can export single article just fine.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.3.4
* How did you install wallabag? Via `git clone`
* Last wallabag version that did not exhibit the issue (if applicable): N/A
* php version: 7.2
* OS: Debian
* type of hosting (shared or dedicated): Dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MariaDB

PR #3833, opened 2019-01-10
Fix settings field inverted

### LINKED ISSUE #3832: Settings field inverted
Related to #3790 that I closed because it was not clear.

In wallabag 2.3.5 (and probably before), fields `Enable CSV export` and `Enable JSON export` are inverted in `Internal settings > Export` menu. It is so in French and English and with both themes.

Consequently, activating CSV export allows for JSON exportation and inversely.

techexo:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no

Fixes #3832 #3790 

The conditions in which the JSON or CSV export button were displayed were inverted in the templates.


j0k3r:
Thanks!

PR #3857, opened 2019-01-17
Replace continue; with break; to avoid PHP 7.3 warnings

tcitworld:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Change `continue` to `break` inside loops for PHP 7.3 compatibility.

### DIFF
--- a/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
+++ b/src/Wallabag/CoreBundle/Form/Type/EntryFilterType.php
@@ -108,7 +108,7 @@ class EntryFilterType extends AbstractType
             ->add('httpStatus', TextFilterType::class, [
                 'apply_filter' => function (QueryInterface $filterQuery, $field, $values) {
                     $value = $values['value'];
-                    if (false === array_key_exists($value, Response::$statusTexts)) {
+                    if (false === \array_key_exists($value, Response::$statusTexts)) {
                         return;
                     }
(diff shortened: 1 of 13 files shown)


j0k3r:
Do we really need these changes as the build is passing with proxy-manager update?
https://travis-ci.org/wallabag/wallabag/builds/481398975

tcitworld:
It avoids the following [warnings](https://secure.php.net/manual/fr/migration73.incompatible.php#migration73.incompatible.core.continue-targeting-switch)
```
Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?
```

j0k3r:
Do we have some of these warnings?

j0k3r:
As far as I can see there is no such warning in recent tests on 2.4: https://travis-ci.org/wallabag/wallabag/jobs/490578426
Two options:
- the code you changed isn't covered by tests, so we need to figure out and apply it
- drop that PR

j0k3r:
The warning only applies to `continue` inside a `switch`. We do not have such case.
Your changes aren't related to `switch` at all :)

But anyway I've added some tests to increase coverage.

PR #3859, opened 2019-01-17
Enable PHPStan

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

The idea is to use it more and more and reach the higher level we could.
I fixed most of from the lowest level.

phpstan will run on travis on the same build as cs-fixer

- Fix error for level 0 & 1 (level 7 has 699 errors...)
- Add `updated_at` to site_credential (so the `timestamps()` method applies correctly)


j0k3r:
Still needs to figure out how to properly run it on Travis now :roll_eyes:

PR #3860, opened 2019-01-17
Make dev/install/update script posix compatible

### LINKED ISSUE #3571: update.sh uses BASH_SOURCE to determine directory

baurmatt:
Fixes #3571.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -3,10 +3,14 @@
 # eg: `sh dev.sh`
 
 COMPOSER_COMMAND='composer'
+REQUIRE_FILE='scripts/require.sh'
 
-DIR="${BASH_SOURCE}"
-if [ ! -d "$DIR" ]; then DIR="$PWD/scripts"; fi
-. "$DIR/require.sh"
+if [ ! -f "$REQUIRE_FILE" ]; then
+  echo "Cannot find $REQUIRE_FILE"
+  exit 1
+fi
+
+. "$REQUIRE_FILE"
 
 $COMPOSER_COMMAND install
 php bin/console wallabag:install
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -25,10 +25,14 @@ if [ "$IGNORE_ROOT" -eq 0 ] && [ "$EUID" == "0" ]; then
 fi
 
 COMPOSER_COMMAND='composer'
+REQUIRE_FILE='scripts/require.sh'
 
-DIR="${BASH_SOURCE}"
-if [ ! -d "$DIR" ]; then DIR="$PWD/scripts"; fi
-. "$DIR/require.sh"
+if [ ! -f "$REQUIRE_FILE" ]; then
+  echo "Cannot find $REQUIRE_FILE"
+  exit 1
+fi
+
+. "$REQUIRE_FILE"
 
 TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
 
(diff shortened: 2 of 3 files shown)


kissgyorgy:
$0 in bash is the name of the script itself. It has nothing to do with a directory.
See: https://www.tldp.org/LDP/abs/html/othertypesv.html

techexo:
We could go for something like `dirname $0`, but I don't know if `dirname` is included by default in all distributions.
It seems that `"$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` is the common way to do it in Bash.

techexo:
@baurmatt It seems that you're right! 

But I don't understand why `${BASH_SOURCE}` in update.sh is considered as being the `DIR`, as it returns the name of the script (and if necessary the relative path to go to it). 
The `if [ ! -d "$DIR" ]; then DIR="$PWD/scripts"; fi` instruction should always be true, shouldn't it ? (In the sense that we're basically looking if the script file is a directory...)

techexo:
I mean, do we even _need_ that? If I'm not mistaken, this is just to differentiate between when the scripts are launched from the `scripts` directory and when they are launched from the `GNUMakefile`. Why not say that everything is launched from the makefile (as I reckon it should) and don't bother with this?

baurmatt:
Not sure why this behavior was implemented. Perhaps people starting the update script from outside of the docroot.

j0k3r:
@techexo this seems to be a better solution than try to handle all compatibility.
I think we can remove that check and just have `. "scripts/require.sh"`

Maybe we should add a check if that file exist and if not, exit and ask to use makefile instead.

I guess it was defined like when people launch sh file from the scripts folder directly.

baurmatt:
Thanks everyone! I've updated the PR as suggested by @j0k3r .

j0k3r:
@Kdecherf are you ok with these changes?

Issue #3662, opened 2018-06-02
PHP Fatal error:  Uncaught RuntimeException: Setting "piwik_enabled" couldn't be found. in /var/www/wallabag/vendor/craue/config-bundle/Util/Config.php:182


jinnko:
### Issue details

Using wallabag 2.3.2 via the official docker container I'm unable to get a working installation with MySQL.

### Environment

* wallabag version 2.3.2
* docker image wallabag/wallabag:latest (aef794efb513)
* php version: 7.1.15 and 7.2.6
* OS: Alpine Linux in the docker container
* Dedicated server
* MySQL 5.6

### Steps to reproduce/test case

1. On MySQL create a database and user for the container instance to use

2. Populate config in an env-file

```
cat > env.list <<-EOF
SYMFONY__ENV__DATABASE_DRIVER="pdo_mysql"
SYMFONY__ENV__DATABASE_HOST="172.17.0.1"
SYMFONY__ENV__DATABASE_PORT="3306"
SYMFONY__ENV__DATABASE_NAME="wallabag"
SYMFONY__ENV__DATABASE_USER="wallabag"
SYMFONY__ENV__DATABASE_PASSWORD="dbpass"
SYMFONY__ENV__FOSUSER_REGISTRATION=False
SYMFONY__ENV__DOMAIN_NAME="https://wallabag.local"
SYMFONY__ENV__SECRET="envsecret"
POPULATE_DATABASE=True
EOF
```

3. Run the container to create the database tables:

```
docker run -it --rm --env-file env.list wallabag/wallabag:latest migrate
```

4. Run the container normally

```
docker run -it --rm --env-file env.list wallabag/wallabag:latest
```

5. Make a request in a browser and see the errors logged

```
Starting provisioner...
 [WARNING]: Found both group and host with same name: localhost
Provisioner finished.
[02-Jun-2018 21:19:13] NOTICE: fpm is running, pid 207
[02-Jun-2018 21:19:13] NOTICE: ready to handle connections
2018/06/02 21:19:15 [error] 209#209: *1 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught RuntimeException: Setting "piwik_enabled" couldn't be found. in /var/www/wallabag/vendor/craue/config-bundle/Util/Config.php:182
Stack trace:
#0 /var/www/wallabag/vendor/craue/config-bundle/Util/Config.php(77): Craue\ConfigBundle\Util\Config->createNotFoundException('piwik_enabled')
[...]

PR #3868, opened 2019-01-29
Add missing entries in craue_config_setting.

Nadrieril:
During `wallabag:install`, the `craue_config_setting` table is populated (see https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Command/InstallCommand.php#L280-L285) with some default entries that are needed for wallabag to function. On a shared hosting however, this command cannot be run so wallabag crashes because of the missing entries (with this error: https://github.com/wallabag/wallabag/issues/3662).
Only some of these entries were created by the migration scripts. This adds the missing entries as part of a migration, to allow shared hosting users to have a complete database. I took the entries from https://github.com/wallabag/wallabag/blob/master/app/config/wallabag.yml#L32-L164, except those that were already added by a previous migration.
I ran the tests but they were not testing for this issue. I however also tested on a new VM and this fixed the issue.
It may now be possible to remove the setting `wallabag_core.default_internal_settings` from config/wallabag.yml (and the code that uses it), unless it is useful to be able to override those before install.
Fixes https://github.com/wallabag/wallabag/issues/3662.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | don't know what that means
| Deprecations? | no
| Tests pass?   | not all, but same as on `master`
| Documentation | no
| Translation   | no
| CHANGELOG.md  | don't know what that means
| License       | MIT

### DIFF
--- a/app/DoctrineMigrations/Version20190129120000.php
+++ b/app/DoctrineMigrations/Version20190129120000.php
@@ -8,54 +8,132 @@ use Wallabag\CoreBundle\Doctrine\WallabagMigration;
 /**
  * Add missing entries in craue_config_setting.
  */
-class Version20190129120000 extends WallabagMigration
+final class Version20190129120000 extends WallabagMigration
 {
-    var $settings = array(
-        array("name" => "carrot", "value" => "1", "section" => "entry"),
-        array("name" => "share_diaspora", "value" => "1", "section" => "entry"),
-        array("name" => "diaspora_url", "value" => "http://diasporapod.com", "section" => "entry"),
-        array("name" => "share_shaarli", "value" => "1", "section" => "entry"),
-        array("name" => "shaarli_url", "value" => "http://myshaarli.com", "section" => "entry"),
-        array("name" => "share_mail", "value" => "1", "section" => "entry"),
-        array("name" => "share_twitter", "value" => "1", "section" => "entry"),
-        array("name" => "show_printlink", "value" => "1", "section" => "entry"),
-        array("name" => "export_epub", "value" => "1", "section" => "export"),
-        array("name" => "export_mobi", "value" => "1", "section" => "export"),
-        array("name" => "export_pdf", "value" => "1", "section" => "export"),
-        array("name" => "export_csv", "value" => "1", "section" => "export"),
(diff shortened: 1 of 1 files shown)


app/DoctrineMigrations/Version20190129120000.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $piwikEnabled = $this->container
+            ->get('doctrine.orm.default_entity_manager')
+            ->getConnection()
+            ->fetchArray('SELECT * FROM ' . $this->getTable('craue_config_setting') . " WHERE name = 'piwik_enabled'");
```
j0k3r: Maybe you should do this check inside the `foreach` just in case people not already create missing values.

j0k3r:
As you are adding a new migration, it'll be available in the 2.4

app/DoctrineMigrations/Version20190129120000.php
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        foreach ($this->settings as $setting) {
+            $this->addSql('
+                DELETE FROM ' . $this->getTable('craue_config_setting') . "
```
Kdecherf: I'm quite uncomfortable with this one. A down of this migration may lead to a deletion of legitimate settings. Any thoughts @j0k3r?

j0k3r: Well we always provide a _down_ migration.
But maybe in that case, we shouldn't because these settings are required.

Kdecherf:
Thanks @Nadrieril :+1:

Nadrieril:
Thank you for doing the fixes !

mweinelt:
This fix is unfortunately not part of any 2.3.x release, preventing new setups to work.

j0k3r:
Thanks for the hint @mweinelt, we are close to release the 2.4.0 with the fix

Issue #3648, opened 2018-05-11
[Wishlist] Selecting content to save


jsuarez1:
Most of the articles I'd want to save to mi Wallabag account I have to copy and paste to a Markdown editor and save them to my cloud space as MD files because WB only saves the main article, this is, it removes additional content like the commentaries. This is a pity because meny times, the really interesting stuff is the debate that some blogspot, Youtube video, forum, etc, opens. Thus, when I'm interested in keed an article with its commentaries, I have to manually select the content I want to save, and copy it in a document.
WB would improve a lot if the users were able to select with our mouse the content that we want to save. This is a suggestion also for the mobile app, of course.

Best regards.

shikasta-net:
This is very similar to #2229 and is something I've been thinking about for a while.  I've been thinking of it as a work-around to the problem of capturing the correct part of certain websites that a lot of users experience.  As far as I can see, the API already provides the functionality in the `EntryRestController.php:postEntriesAction` endpoint, but I've not had time to test that yet.

Issue #2321, opened 2016-10-01
Add ability to edit imported contents


amereservant:
### Issue details

This is a feature request.  I'm not sure why this doesn't already exist, but I would like the ability to edit the contents of an imported article.  Sometimes it doesn't import the contents correctly, or an image URL isn't correct, or I'd like to add content to it, etc. and that currently isn't possible.

It would be nice to add an edit button, like the one for the title, so the contents can be edited.  A WYSIWYG editor isn't even necessary.  Editing the raw code would work fine.

Thank you!

rabyte:
I like this idea very much. This happens to me a lot, too. E.g. it would be great beeing able to remove messed up content or unnecessary footers. 

I also was wondering if it is not possible already since editing the title works.
If not I guess it should be straight forward to take over the "title editing function" for the content?

Regards

slavniyteo:
This feature doesn't seem to have a high priority so I'll write here a simple workaround to edit article's content.

There is a form to modify a title of article and it's original url. You can get there by clicking a *pencil* icon next to article's title. We'll modify this form. 

![image](https://user-images.githubusercontent.com/18638390/64022057-07c0d680-cb3e-11e9-89d3-0e649b2b8582.png)

---

Add two pieces of code: to [EditEntryType.php](https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Form/Type/EditEntryType.php)

At the top: 
```php
use Symphony\Component\Form\Extension\Core\Type\TextareaType;
```

At the **buildForm** function, for example between 'title' and 'url' fields:
```php
->add('content', TextareaType::class, [
    'required' => true,
    'label' => 'Content',
    'attr' => [ 'style' => 'height: 20em; padding: 10px' ]
]
```

After you save the file you can refresh the page and get additional field with content of the article.

![image](https://user-images.githubusercontent.com/18638390/64021980-cfb99380-cb3d-11e9-87ab-c2be0a8767d1.png)

Now we can modify the content of article in plain html format. I use it to fix some parsing issues. It is very useful.

lizyn:
> Add two pieces of code: to [EditEntryType.php](https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Form/Type/EditEntryType.php)
> 
> At the top:
> 
> ```
> use Symphony\Component\Form\Extension\Core\Type\TextareaType;
> ```

[...]

PR #3871, opened 2019-02-01
Article content paste/edit

### LINKED ISSUE #2229: Copy Paste content into wallabag
### New feature request

I would love to be able to paste content into wallabag. For example : 
1. Copy a text (word, PDF, web page content)
2. Open a wallabag session in my browser
3. Click on the button "+" (Add a new entry)
4. Paste my content
5. Wallabag recognizes that it's a content, and not a URL
6. Wallabag creates the article with my content

And as additional features:
- The title of the article will be the first sentence of the pasted text, and it could be edited afterwards
- The pasted content can contain formatted text / pictures / table / ... and wallabag is able to grab them correctly
- I can do all of this in my browser extension, without opening the wallabag page

This is related to https://github.com/wallabag/wallabag/issues/1634

shikasta-net:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Resolves #2229 
Relates to #3648

Hi, I'd been wanting the ability to correct miss-import and paste in articles that the scraper was failing, for a while.  Since I came across others asking for this I thought I'd have another look and it was quite easy to implement.

I've added a content editor to the edit page, linked from and Edit button in the article itself.  This has required adding a dependency, so I'm submitting this for feedback on license and implementation before I do any more changes; for example, I haven't provided translations or another theme.

Using the Material theme, the editor can be accessed from an article's side menu.  Content can be highlighted and copied from any web page and pasted into the WYSIWYG editor, which preserves the formatting.  Is there anything I need to specifically have it strip out; urls or linked images for example?

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -25,6 +25,7 @@ class AppKernel extends Kernel
             new Bazinga\Bundle\HateoasBundle\BazingaHateoasBundle(),
             new Lexik\Bundle\FormFilterBundle\LexikFormFilterBundle(),
             new FOS\OAuthServerBundle\FOSOAuthServerBundle(),
+            new FOS\CKEditorBundle\FOSCKEditorBundle(),
             new Stof\DoctrineExtensionsBundle\StofDoctrineExtensionsBundle(),
             new Scheb\TwoFactorBundle\SchebTwoFactorBundle(),
             new KPhoen\RulerZBundle\KPhoenRulerZBundle(),
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -393,3 +393,51 @@ httplug:
                           'X-Accept': 'application/json'
     discovery:
         client: false
+
+fos_ck_editor:
+    default_config: default
+    configs:
+        default:
+            toolbar: basic
+        article_editor:
+            toolbar:
+                - name: 'basicstyles'
+                  items:
+                    - 'Bold'
+                    - 'Italic'
+                    - 'Underline'
+                    - 'Strike'
+                    - 'Subscript'
+                    - 'Superscript'
+                    - '-'
+                    - 'RemoveFormat'
+                - name: 'paragraph'
+                  items:
+                    - 'NumberedList'
+                    - 'BulletedList'
+                    - '-'
+                    - 'Outdent'
+                    - 'Indent'
+                    - '-'
+                    - 'Blockquote'
+                    - 'CreateDiv'
+                    - '-'
+                    - 'JustifyLeft'
+                    - 'JustifyCenter'
+                    - 'JustifyRight'
+                    - 'JustifyBlock'
+                - name: 'links'
+                  items:
+                    - 'Link'
+                    - 'Unlink'
+                    - 'Anchor'
+                - '/'
+                - name: 'styles'
+                  items:
+                    - 'Format'
+                    - 'Font'
+                    - 'FontSize'
+                - '/'
+                - name: 'document'
+                  items:
+                    - 'Source'
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -40,6 +40,7 @@ twig:
     strict_variables: "%kernel.debug%"
     form_themes:
         - "LexikFormFilterBundle:Form:form_div_layout.html.twig"
+        - "@FOSCKEditor/Form/ckeditor_widget.html.twig"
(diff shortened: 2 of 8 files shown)


shikasta-net:
I've gone back through, rebased onto the new master and made sure everything still works. Unfortunately I'm not able to get it to pass CI because an unrelated test is failing:
EntryRestControllerTest::testReloadEntry gets 304 rather than 200 but not if I run it locally.  Some help would be appreciated as I think this is ready unless someone has suggestions.

j0k3r:
Good job on rebasing!
Thanks for your patience.

It's an interesting PR but I'm not convinced in adding the whole CKEditor to just paste content. Do we really need to paste content and then apply some bold/list/whatever on some part of the text?

shikasta-net:
Good questions, @j0k3r.  To be honest, it's been so long since I did this that I can't remember why I chose CKEditor.  I seem to remember that I thought it was a very large download at the time, so there must have been a good reason (maybe to do with ease of connecting to symfony).  It shouldn't be hard to swap to another though.  What particular requirements/objections do you have?  The reason for using a WYSWYG instead of textarea is to preserve the HTML structure when doing a highlight-copy-paste of a page. I've not found a way to configure a textarea that will do this, I just get the raw text without formatting or images.  Pasting the page source is an option but it's an extra step and requires quite a bit more tag trimming (especially annoying for request #3648).  
TinyMCE looks like it will work, so I'll give that a try in the meantime.

j0k3r:
I think it might be too massive to download too. We aren't transforming wallabag into a CMS :)
Maybe don't rush in wasting time about testing other solution. I think we should debate about the possibility before trying sth.

Thought @wallabag/core?

shikasta-net:
Thanks for taking the time to look at this.  My intention is to provide a workaround for when the scraper can't capture a site or if the user wants to capture a different part of a page than the scraper does.  More towards scrap-booking than CMS, but I see your point.

After speaking to someone, I worked out two much lighter options that would be worth considering in your discussions.  There is a way to capture the paste event and grab the structured HTML rather than the raw text, which means it could either
1. paste the raw HTML markup into an editable `textarea`, or
2. paste the rendered HTML into a contenteditbal `div`.

1 is closer to my original intention whereas 2 feels more like pasting the actual page.
Any solution would need to have some sort of pre-save cleanup to strip out scripts, etc.

Anyway, thanks for taking a look.  I've been wanting to do something about this for years so I don't have to keep fixing database entries.

lasc:
That would be excellent feature, I always bump into this. Even just text blob will be better than now, I am using wallabag as bookmark and inability to full text search after drive me nuts.

3bbing:
Oh I did not find this PR before doing this: https://github.com/wallabag/wallabag/issues/2321#issuecomment-612384242
My thoughts went the same direction and also thought about CKE because of its workability with symfony but then just went with tinymce not willing to fiddle. OP here said he'll try but didn't come back to it.
For me it's quite lightweight and working to just import TinyMCE and activate on the content's textarea. (2.4MB unzipped with everything & unstripped) 

Am I right that the js would need to go to /app/Resources/static/themes/_global/ and then needs an import - where would the proper place be to import the js only for the edit view? - and call in the theme for entry edit at src/Wallabag/CoreBundle/Resources/views/themes/ and that's it? 

So far it works fine. But i just pasted the import into the material theme file and simply used tinymce.min.js from a different project of mine - both of these things are ofc just lazy and not publishable. 
Where would it go properly?

shikasta-net:
I'm glad to see someone else thought this was a good solution!  I never got around to trying TinyMCE because it involved integrating a different package.  I was also more interested in the light weight option of making a content-editable DIV, but then got distracted.  I'm very happy to go with TinyMCE if it is acceptable (licence and size) for the project.

The project may have changed a lot since I did this work but my understanding was that you need to use Composer to install a bundle and then include it as a bundle in the AppKernel; basically the same as the first two of my commits.  Beware of which Symfony version you're trying to integrate with: that's a problem I spent ages trying to work out but the project may have upgraded since then.

nicosomb:
I understand your needs with this feature. However, as @j0k3r said, we are not sure that embedding a tool like TinyMCE or CKEditor is a good thing for this single feature.

Issue #3884, opened 2019-02-14
Update script with root produces FileLocatorFileNotFoundException


mart-e:
### Issue details

Try to update the wallabag as root produces an error
```
# ./scripts/update.sh --ignore-root-warning
...
> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache

                                                                               
  [Symfony\Component\Config\Exception\FileLocatorFileNotFoundException]        
  The file "/srv/.../wallabag/app/config/config_--ignore-root-  
  warning.yml" does not exist.                                                 
                                                                               

Script Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache handling the post-cmd event terminated with an exception

                                                                                   
  [RuntimeException]                                                               
  An error occurred when executing the "'cache:clear --no-warmup'" command:        
                                                                                   
    [Symfony\Component\Config\Exception\FileLocatorFileNotFoundException]          
    The file "/srv/../wallabag/app/config/config_--ignore-root-    
    warning.yml" does not exist.                                                   
                                                                                   
                                                                                   

install [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--] [<packages>]...

```
I believe this is a bug as it is considering my argument `--ignore-root-warning` as an environment variable or something.
I am using root as my web server user has no shell access.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.3.6
* How did you install wallabag? Via `git clone`
* php version: 7.0
* OS: debian
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install : MariaDB

j0k3r:
As defined in the `update.sh`, you should use the environment as first argument.
https://github.com/wallabag/wallabag/blob/master/scripts/update.sh#L3

Is it working? `sh install.sh prod --ignore-root-warning`

mart-e:
No because this line expects the ignore-root to be the first argument

https://github.com/wallabag/wallabag/blob/b1992b340e21b6846a1ec2ae6ddd7217f3b24fb5/scripts/update.sh#L8

If I modify the `$1` to `$2` it works then.
Do you want me to submit a PR?

[...]

PR #3885, opened 2019-02-14
Allow optional --ignore-root-warning

mart-e:
Now both
```
./install prod --ignore-root-warning
./install --ignore-root-warning prod
```
will work

$1 was both used for the environment and the root flag

Fixes wallabag/wallabag#3884

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| License       | MIT

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -5,9 +5,17 @@
 IGNORE_ROOT_ARG="--ignore-root-warning"
 IGNORE_ROOT=0
 
-if [ "$1" == "$IGNORE_ROOT_ARG" ]; then
-    IGNORE_ROOT=1
-fi
+while :; do
+    case $1 in
+        $IGNORE_ROOT_ARG) IGNORE_ROOT=1
+        ;;
+        *[a-zA-Z]) ENV=$1
+        ;;
+        *) break
+        ;;
+    esac
+    shift
+done
 
 # Abort running this script if root
 if [ "$IGNORE_ROOT" -eq 0 ] && [ "$EUID" == "0" ]; then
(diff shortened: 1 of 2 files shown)


mart-e:
Not sure if important but `$ENV` will be erased at each step. So
```
./install prod foo bar
```
will keep only `bar` for `$ENV` variable

scripts/install.sh
```diff
@@ -5,9 +5,17 @@
...
-    IGNORE_ROOT=1
-fi
+while :; do
+    case $1 in
+        $IGNORE_ROOT_ARG) IGNORE_ROOT=1
+        ;;
+        *[a-zA-Z]) ENV=$1
+        ;;
+        *) break
+        ;;
```
Deuchnord: I'm not sure to understand what's happening in the `case` block, could you explain, please?

scripts/install.sh
```diff
@@ -22,7 +30,6 @@ DIR="${BASH_SOURCE}"
 if [ ! -d "$DIR" ]; then DIR="$PWD/scripts"; fi
 . "$DIR/require.sh"
 
-ENV=$1
```
Deuchnord: What if `$ENV` is not defined in the loop?

Glandos:
I'm really not familiar with Bash argument parsing (which was not designed to be easy), but it seems to fix the bug.

scripts/install.sh
mart-e: same as if I do `echo $NOTDEFINEDVAR`, just get none/empty string (does bash have the concept of undefined var?). Or same as before if I execute the script with no arguments.

Kdecherf: Bash is permissive by default when dealing with undefined vars. You can set an option to enable a restrictive mode which will fail the whole script if it encounters a single undefined var but it's not the case here.
Anyway, this PR does not change the former behavior regarding `$ENV` so I don't have any objection.
You can add a new commit or open a new PR to handle the specific case of empty `$ENV` :)

PR #3887, opened 2019-02-17
Add support of expect parameter to change return object when deleting entry

### LINKED ISSUE #3711: Remove response returned by API's delete entry call
### Issue details

This issue is a follow up of #3608, failed fix #3516 and https://github.com/wallabag/android-app/issues/646.

When an API client deletes an entry (`DELETE /api/entries/{id}.json`), the server returns the full entry but its id, example:

``` json
{"is_archived":0,"is_starred":0,"user_name":"kdecherf","user_email":"kevin@kdecherf.com","user_id":1,"tags":[],"is_public":false,"id":null,"uid":null,"title":"…
```

The `null` value currently triggers an exception on the Android app, which leads to another side-effect: the client emits two `DELETE` requests for the same entry (see #3608).

It seems that #3516 does not restore the missing id at all.

While checking network packets I saw that the actual content of the entry was also sent. In my case the response to the delete request had a body of 70 KB which may degrade clients' performance on lossy connections.

I don't see any value in sending back the entity when deleting it, so I would like to remove any content sent by this call. The API would then return a `204 No content` for successful deletion, which also fixes https://github.com/wallabag/android-app/issues/646 (_tested here_).

@wallabag/android @wallabag/core do you have any objection?

### Environment

* wallabag version (or git revision) that exhibits the issue: _2.3.3_
* How did you install wallabag? Via `git clone` or by downloading the package? _git_
* Last wallabag version that did not exhibit the issue (if applicable): _N/A_
* php version: _7.0_
* OS: _Debian_
* type of hosting (shared or dedicated): _dedicated_
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): _PostgreSQL_

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes #3711

This PR fixes a test, the clone and add a new `expect` parameter to the `delete entry` call. When specifying `?expect=id` the API will return the id of the removed entry instead of returning the whole object.
By default, `expect` is set to `entry` which is the current behavior.

Fixes https://github.com/wallabag/android-app/issues/646

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -570,18 +570,31 @@ class EntryRestController extends WallabagRestController
      * @ApiDoc(
      *      requirements={
      *          {"name"="entry", "dataType"="integer", "requirement"="\w+", "description"="The entry ID"}
+     *      },
+     *      parameters={
+     *          {"name"="expect", "dataType"="string", "required"=false, "format"="id or entry", "description"="Only returns the id instead of the deleted entry's full entity if 'id' is specified. Default to entry"},
      *      }
      * )
      *
      * @return JsonResponse
      */
-    public function deleteEntriesAction(Entry $entry)
+    public function deleteEntriesAction(Entry $entry, Request $request)
     {
+        $expect = $request->query->get('expect', 'entry');
+        if (!\in_array($expect, ['id', 'entry'], true)) {
+            throw new BadRequestHttpException(sprintf("expect: 'id' or 'entry' expected, %s given", $expect));
+        }
         $this->validateAuthentication();
         $this->validateUserAccess($entry->getUser()->getId());
 
-        // We copy $entry to keep id in returned object
-        $e = $entry;
+        $response = $this->sendResponse([
+            'id' => $entry->getId(),
+        ]);
+        // We clone $entry to keep id in returned object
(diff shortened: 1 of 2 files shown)


src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -570,18 +570,31 @@ public function patchEntriesReloadAction(Entry $entry)
...
+     *          {"name"="expect", "dataType"="string", "required"=false, "format"="id or entry", "description"="Only returns the id instead of the deleted entry's full entity if 'id' is specified. Default to entry"},
      *      }
      * )
      *
      * @return JsonResponse
      */
-    public function deleteEntriesAction(Entry $entry)
+    public function deleteEntriesAction(Entry $entry, Request $request)
     {
+        $expect = $request->query->get('expect') ?: 'entry';
```
j0k3r: This should be enough:
```suggestion
        $expect = $request->query->get('expect', 'entry');
```

Kdecherf: Thanks :+1: 
(I could have used the commit suggestion feature of github but I saw it after my push)

PR #3937, opened 2019-04-23
 Add ability to match many domains for credentials

### LINKED ISSUE #3936: Use stored domain credentials for all its subdomains
### Issue details

Wallabag uses Graby to fetch the links it stores. Graby uses site-specific configuration files, that contain parsing rules and/or paywall configuration. Also, configuration files may be specified for a domain and in turn cover all subdomains on that domain.

When grabbing a link, a SiteConfig is created for the host; in the process a configuration file is loaded from the disk and credentials are loaded from the DB (if available - in both cases). At this point, while we can have site configuration on the domain-level, all the credentials are fetched on the hostname level. _As a result, for some news portals one needs to specify credentials for all the subdomains used by the domain. (As an example take https://ihned.cz, which has articles on various subdomains)_

I suggest to enable storing credentials in the format 

> .domain.com (or .ihned.cz for the given exmaple)

which would make it more aligned with the site config capabilities/specification.

### Environment

* wallabag version (or git revision) that exhibits the issue: **all**
* How did you install wallabag? **Docker**
* php version:
* OS:
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): **PostgreSQL**

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | kind of
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | maybe?
| Translation   | no
| CHANGELOG.md  | yes/no
| License       | MIT

Fixes #3936

Add ability to match many domains for credentials  …
Instead of fetching one domain, we use the same method as in site config (to retrieve the matching file) and handle api.example.org, example.org, .org (yes the last one isn’t useful).
If one of these match, we got it and use it.

### DIFF
--- a/src/Wallabag/CoreBundle/DataFixtures/ORM/LoadSiteCredentialData.php
+++ b/src/Wallabag/CoreBundle/DataFixtures/ORM/LoadSiteCredentialData.php
@@ -5,19 +5,38 @@ namespace Wallabag\CoreBundle\DataFixtures\ORM;
 use Doctrine\Common\DataFixtures\AbstractFixture;
 use Doctrine\Common\DataFixtures\OrderedFixtureInterface;
 use Doctrine\Common\Persistence\ObjectManager;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 use Wallabag\CoreBundle\Entity\SiteCredential;
 
-class LoadSiteCredentialData extends AbstractFixture implements OrderedFixtureInterface
+class LoadSiteCredentialData extends AbstractFixture implements OrderedFixtureInterface, ContainerAwareInterface
 {
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(ContainerInterface $container = null)
+    {
+        $this->container = $container;
+    }
+
     /**
      * {@inheritdoc}
      */
     public function load(ObjectManager $manager)
     {
         $credential = new SiteCredential($this->getReference('admin-user'));
-        $credential->setHost('example.com');
-        $credential->setUsername('foo');
-        $credential->setPassword('bar');
+        $credential->setHost('.super.com');
+        $credential->setUsername($this->container->get('wallabag_core.helper.crypto_proxy')->crypt('.super'));
+        $credential->setPassword($this->container->get('wallabag_core.helper.crypto_proxy')->crypt('bar'));
+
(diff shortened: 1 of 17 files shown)


tcitworld:
Hmm, meaning if there's a subdomain which requires auth, credentials will be sent to all of the other subdomains.
For instance setting credentials for `paywall.wordpress.com` will give them to all of `*.wordpress.com` subdomains, so maybe we should give a warning here.

j0k3r:
@Kdecherf it already works like that :v: 

@tcitworld After more tests that I'll push soon, it doesn't work like that. Let's check some examples.

For `paywall.wordpress.com` defined in credentials:

requested host | looking for credentials | result
------------ | ------------- | -------------
`subdomain.wordpress.com` | `subdomain.wordpress.com` & `.wordpress.com` | :negative_squared_cross_mark: 
`wordpress.com` | `wordpress.com` & `.com` | :negative_squared_cross_mark: 
`paywall.wordpress.com` | `paywall.wordpress.com` & `.wordpress.com` | :heavy_check_mark: 

For `.wordpress.com` defined in credentials:

requested host | looking for credentials | result
------------ | ------------- | -------------
`subdomain.wordpress.com` | `subdomain.wordpress.com` & `.wordpress.com` | :heavy_check_mark: 
`wordpress.com` | `wordpress.com` & `.com` | :negative_squared_cross_mark: 
`paywall.wordpress.com` | `paywall.wordpress.com` & `.wordpress.com` | :heavy_check_mark: 

For `wordpress.com` defined in credentials:

requested host | looking for credentials | result
------------ | ------------- | -------------
`subdomain.wordpress.com` | `subdomain.wordpress.com` & `.wordpress.com` | :negative_squared_cross_mark: 
`wordpress.com` | `wordpress.com` & `.com` | :heavy_check_mark: 
`paywall.wordpress.com` | `paywall.wordpress.com` & `.wordpress.com` | :negative_squared_cross_mark: 

:arrow_right: We need to determine how the last two are handled:

1. keep it like that
1. defining `.wordpress.com` should match host `wordpress.com` too
1. defining `wordpress.com` should match host `*.wordpress.com` too

:warning: The only issue we have is if someone have overlapping credentials like:
- paywall.wordpress.com
- .wordpress.com

The _wildcard_ one should be the last to be created otherwise it'll be the first to match for hosts `*.wordpress.com`.

Kdecherf:
Oh okay 👍

Issue #3919, opened 2019-04-01
Slow query when Wallabager checks a new page on visit.


shtrom:
I have seen this in my slow query log (prepare for meta):
```sql
mariadb root@localhost:wallabag> explain SELECT w0_.id AS id_0, w0_.uid AS uid_1, w0_.title AS title_2, w0_.url AS url_3, w0_.is_archive
                                 d AS is_archived_4, w0_.is_starred AS is_starred_5, w0_.content AS content_6, w0_.created_at AS created
                                 _at_7, w0_.updated_at AS updated_at_8, w0_.published_at AS published_at_9, w0_.published_by AS publishe
                                 d_by_10, w0_.starred_at AS starred_at_11, w0_.mimetype AS mimetype_12, w0_.language AS language_13, w0_
                                 .reading_time AS reading_time_14, w0_.domain_name AS domain_name_15, w0_.preview_picture AS preview_pic
                                 ture_16, w0_.http_status AS http_status_17, w0_.headers AS headers_18, w0_.origin_url AS origin_url_19,
                                  w0_.user_id AS user_id_20 FROM `wallabag_entry` w0_ WHERE w0_.url = 'https://github.com/wallabag/walla
                                 bag/issues?utf8=✓' AND w0_.user_id = 1;
+----+-------------+-------+------+----------------------+----------------------+---------+-------+------+-------------+
| id | select_type | table | type | possible_keys        | key                  | key_len | ref   | rows | Extra       |
+----+-------------+-------+------+----------------------+----------------------+---------+-------+------+-------------+
| 1  | SIMPLE      | w0_   | ref  | IDX_F4D18282A76ED395 | IDX_F4D18282A76ED395 | 5       | const | 5592 | Using where |
+----+-------------+-------+------+----------------------+----------------------+---------+-------+------+-------------+
1 row in set
```
Note the number of rows that needed to be inspected.

It essentially happens everytime I open a tab in a browser running Wallabager. It looks to me like we'd need a new compound index on `user_id` AND something that relates to the `url`.

Now, the problem is that the `url` is a `longtext`, so we can't place an index directly on it. I would suggest adding a column containing a (fast) hash of the URL, so it's a `VARCHAR` of a known size, and include that in an index on `(`user_id`, `url_hash`)`.

This will also require changing all queries that do a `WHERE url_hash=MD5('http://...`)` (we're cool with MD5, there's nothing security-related here, just performance). Note however [0], which further points out.

[...]

PR #3944, opened 2019-05-02
Always hash exists url

shtrom:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  |no
Fixed tickets |#3919
| License       | MIT

Simplify the logic from #3158 by hashing all the urls from the request,
and only doing a search by hash. This allows to get performance benefits
from the new indexed hash column even when using older clients that do
not hash the URL in the request.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -848,8 +848,8 @@ class EntryRestController extends WallabagRestController
     /**
      * Return information about the entry if it exist and depending on the id or not.
      *
-     * @param Entry|null $entry
-     * @param bool       $returnId
+     * @param Entry|bool|null $entry
+     * @param bool            $returnId
      *
      * @return bool|int
      */
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -14,6 +14,7 @@ use Wallabag\CoreBundle\Entity\Entry;
 use Wallabag\CoreBundle\Entity\Tag;
 use Wallabag\CoreBundle\Event\EntryDeletedEvent;
 use Wallabag\CoreBundle\Event\EntrySavedEvent;
+use Wallabag\CoreBundle\Helper\UrlHasher;
 
 class EntryRestController extends WallabagRestController
 {
@@ -56,8 +57,8 @@ class EntryRestController extends WallabagRestController
         }
 
         $urlHashMap = [];
-        foreach($urls as $urlToHash) {
-            $urlHash = hash('sha1', $urlToHash); // XXX: the hash logic would better be in a separate util to avoid duplication with GenerateUrlHashesCommand::generateHashedUrls
+        foreach ($urls as $urlToHash) {
+            $urlHash = UrlHasher::hashUrl($urlToHash);
             $hashedUrls[] = $urlHash;
             $urlHashMap[$urlHash] = $urlToHash;
         }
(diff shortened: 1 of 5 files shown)


shtrom:
@j0k3r this is a small change in your hash_url branch. I still need to test it locally on my instance, but feedback is welcome.

shtrom:
Hum, Travis fails for my branch in the same pattern as it does for 2.4 without my change: https://travis-ci.org/wallabag/wallabag/jobs/524429166

shtrom:
Hum couple more thoughts: maybe we should hide the hash calculation and put it in `findByUrlAndUserId`, so any search actually uses the hash.

I do note, however, that the migration from #3158, while it creates the column, doesn't seem to calculate and populate the hashes for the existing entries. Could this be an issue, or is it handled somewhere else?

shtrom:
Actually, here's another change, making `EntryRepository::findByUrlAndUserId` hash the `url` under the hood, and call `EntryRepository::findByHasedUrlAndUserId` unconditionally.

This probably need some cleanup, as the `hash('sha1', ...)` is starting to be duplicated a bit too often, but what do you think of the idea?

One thing I am not sure about is what to do about the `urldecode` that `findByUrlAndUserId` was doing, but doesn't seem to be used in the `findByHashedUrlAndUserId` path. Or should the URL be `urldecode`d before being hashed?

shtrom:
Ok, tested on my instance, and it works nicely. No more slow queries, and Wallabagger is faster at turning green on already visited pages. This is mostly thanks to my second commit there, which uses the hash for everything. The question remains about the `urldecode` part. It doesn't seem to be used consistently.

Another thing I did notice is that the `wallabag:generate-hashed-urls` command is not ran during the migration, so it would probably be worth adding a note that this needs to happen in the release notes.

src/Wallabag/CoreBundle/Resources/config/services.yml
```diff
@@ -220,6 +220,13 @@ services:
             - "%wallabag_core.site_credentials.encryption_key_path%"
             - "@logger"
 
+    wallabag_core.helper.url_hasher:
```
j0k3r: I'm wondering if this should be a service or a simple class with static method in it :thinking: 
BTW, the file is missing `UrlHasher.php`

shtrom:
There we go. I added a `UrlHasher` helper, with a configurable hash algorithm (defaulting to `sha1` in the constructor). I roped it in in the `services.yml`, though I haven't found a way to make it fetchable by `Entry`s; those instantiate the class directly, and rely on the default hash algorithm to be the same as for the other services (I added a note to that effect in the `services.yml`).

src/Wallabag/CoreBundle/Resources/config/services.yml
shtrom: :man_facepalming:

shtrom: I thought about a static class, then figured that we may want to be able to change the hash algorithm later on at runtime, so made that a configurable instance attribute.

j0k3r: For now I do not think we'll need to update the algorithm at runtime and I prefer to keep that helper dead simple with a static method and no service.

`sha1` can still be the default algo.

```php
<?php

namespace Wallabag\CoreBundle\Helper;

/**
 * Hash URLs for privacy and performance.
 * Internal helper to centralize how the url is hashed.
 */
class UrlHasher
{
    /**
     * @param string $url
     * @param string $algorithm Available algo for the PHP function `hash`
     *
     * @return string hashed $url
     */
    public function hash(string $url, string $algorithm = 'sha1')
    {
        return hash($algorithm, $url);
    }
}
```

shtrom:
How about now? A static class does limit the changes needed quite a bit :+1:

j0k3r:
I rebased, squashed some commits & fix tests.
I think this is ok now.

j0k3r:
Thanks for starting the job @shtrom !

Issue #3505, opened 2017-12-14
cached GIFs are not playable


Strubbl:
### Issue details
When i have image caching enabled on server and i add an article containing a movie in a gif, the gif is not correctly fetched. The animation is gone in wallabag article although there is one in the original article.

### Environment

* wallabag version (or git revision) that exhibits the issue: master commit 494b47f29915cc881cfeba14c16b728d3adcb85a which shows itself in wallabag as 2.3.0
* How did you install wallabag? Via `git clone` or by downloading the package? git
* Last wallabag version that did not exhibit the issue (if applicable): - 
* php version: 7.0
* OS:debian stretch
* type of hosting (shared or dedicated):dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): postgres

### Steps to reproduce/test case

Please provide necessary steps for reproduction of this issue, or better the
reduced test case (without any external dependencies, if possible).

0. have image caching in your wbg instance active
1. add article https://www.wallabag.org/en/news/new-release-wallabag-2-3-is-there-really
2. compare kangaroo animation in wallabag article with original article

j0k3r:
Maybe `imagegif` isn't enough to keep animation.
https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Helper/DownloadImages.php#L131

nicosomb:
https://stackoverflow.com/questions/9806732/moving-gif-animated-file-loses-animation-with-imagegif

Simounet:
Is it a security issue that push you to create every image again? If it's your position, we can try to use `imagemagick` if available and use `imagegif` as a fallback.

PR #3955, opened 2019-05-10
Use Imagick to keep GIF animation

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | kind of
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documentation | not yet
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

If Imagick is available, GIF will be saved using it to keep animation.
Otherwise the previous method will be used and the animation won't be kept.

Fix https://github.com/wallabag/wallabag/issues/3505

- Installing Imagick on Travis to avoid phpstan to fail because it can't find the class
- Fixing CS from php-cs-fixer update

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,10 @@ before_script:
     - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - phpenv config-rm xdebug.ini || echo "xdebug not available"
     - composer self-update --no-progress
+    # install imagick
+    - pear config-set preferred_state beta
+    - pecl channel-update pecl.php.net
+    - yes | pecl install imagick
 
 script:
     - travis_wait bash composer install -o --no-interaction --no-progress --prefer-dist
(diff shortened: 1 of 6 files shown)


src/Wallabag/CoreBundle/Helper/DownloadImages.php
```diff
@@ -135,7 +135,16 @@ public function processSingleImage($entryId, $imagePath, $url, $relativePath = n
 
         switch ($ext) {
             case 'gif':
-                imagegif($im, $localPath);
+                // use Imagick if available to keep GIF animation
+                if (class_exists('\\Imagick')) {
+                    $imagick = new \Imagick();
```
tcitworld: Maybe catch `ImagickException` here ?

.travis.yml
```diff
@@ -54,6 +54,10 @@ before_script:
     - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - phpenv config-rm xdebug.ini || echo "xdebug not available"
     - composer self-update --no-progress
+    # install imagick
```
tcitworld: Ref https://github.com/travis-ci/travis-ci/issues/781#issuecomment-332473065

PR #3959, opened 2019-05-11
mysql: change collation of tag label

### LINKED ISSUE #3302: Wrong tag added if the same slug already exists
Hi there,
On wallabag.it, if I try to add the `cache` tag then validate it, it is switched to the `caché` tag word. I think the person creating a word with a slug that defining other words wins.
It could be great to have some kind of rules for special chars.

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

`utf8mb4_unicode_ci` considers that 'caché' is equal to 'cache' which
can lead to attaching incorrect tags to entries. This issue is due to
some unicode normalization done by MySQL.

`utf8mb4_bin` makes no unicode normalization, letting wallabag to consider
'cache' and 'caché' as two different tags.

Fixes #3302

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/Tag.php
+++ b/src/Wallabag/CoreBundle/Entity/Tag.php
@@ -13,7 +13,10 @@ use JMS\Serializer\Annotation\XmlRoot;
  * Tag.
  *
  * @XmlRoot("tag")
- * @ORM\Table(name="`tag`")
+ * @ORM\Table(
+ *     name="`tag`",
+ *     options={"collate"="utf8mb4_bin", "charset"="utf8mb4"},
+ * )
  * @ORM\Entity(repositoryClass="Wallabag\CoreBundle\Repository\TagRepository")
  * @ExclusionPolicy("all")
  */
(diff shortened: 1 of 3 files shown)


Kdecherf:
**UPDATE**

I change the collation of the whole table as Doctrine does not support
setting a collation on a column for a specific platform (_it tries to
apply `utf8mb4_bin` even for pgsql and sqlite_).

src/Wallabag/CoreBundle/Entity/Tag.php
```diff
@@ -13,7 +13,10 @@
  * Tag.
  *
  * @XmlRoot("tag")
- * @ORM\Table(name="`tag`")
+ * @ORM\Table(
+ *     name="`tag`",
+ *     options={"collate"="utf8mb4_bin", "charset"="utf8mb4"},
```
j0k3r: Don't you need to update the migration accordingly?
How is it handled by PG & SQLite as they shouldn't know what utf8mb4 is?

j0k3r:
Ok I've tried updating the migration to reflect the collate on the whole table but your test then fail.
As we build the database using migration, it'll be ok. As far as I remember the collate change to the whole table was only to fix tests.

PR #3960, opened 2019-05-11
api/entries: add parameter detail to exclude or include content in response

### LINKED ISSUE #2817: performance issue in entries API
### Issue details

It seems to me there is a big performance issue in the entries listing API. From what I can tell, the entries are listed in the `_embedded` JSON field, and in there have more or less the [following structure](https://github.com/Strubbl/wallabago/blob/master/entries.go#L28) (sorry for the golang, I couldn't find that documentation in the API directly):

```
type Item struct {
	Links       Links         `json:"_links"`
	Annotations []interface{} `json:"annotations"`
	CreatedAt   WallabagTime  `json:"created_at"`
	DomainName  string        `json:"domain_name"`
	ID          int           `json:"id"`
	IsArchived  int           `json:"is_archived"`
	IsStarred   int           `json:"is_starred"`
	Mimetype    string        `json:"mimetype"`
	ReadingTime int           `json:"reading_time"`
	Tags        []interface{} `json:"tags"`
	UpdatedAt   WallabagTime  `json:"updated_at"`
	UserEmail   string        `json:"user_email"`
	UserID      int           `json:"user_id"`
	UserName    string        `json:"user_name"`
}
```

Most of those fields are fine and small, but wise API developers will notice one field is missing: `content`. It is missing because it was optimized out of there: it's just too frigging huge to carry around the memory all the time, in my program.

Unfortunately, even though I don't parse it, it is still sent by the API. This means that, in my use case, even if I just *list* the unread entries with the API, i receive a 2MB JSON blob. And that's for 66 unread entries. I can't imagine the monstrosity the API would return if I would ask it to list all 5000 archived entries.

I know there's paging and everything, but it seems to me that listing entries and *fetching* individual entries should be separate operations.

It is unclear to me *why* the full content is sent in API operations, given that there's a distinct `/api/entries/{entry}.{format}` endpoint. I understand it may be useful for certain endpoints to avoid making multiple roundtrips, but I suspect it makes a lot of API calls needlessly slow...

Could there be a flag to keep the API from sending the actual content? That would retain backwards compatibility and allow for faster requests for those who just want a list of article titles (for example).

This is a key performance issue I have found when developing the [wallabako](https://gitlab.com/anarcat/wallabako/) downloader, which fetches the list of articles and then downloads EPUB versions of given entries. It doesn't need the actual content of the items when it does the listing, because it wants the EPUB version.. To give you an idea, it takes at least one full second to list the 66 unread entries in the API here. Downloading them after takes of course much longer, but once that's done, there's nothing else to do and most of the time the program is waiting on the API:

```
2017/01/31 10:14:21 logging in to https://lib3.net/wallabag
2017/01/31 10:14:21 CSRF token found: 200 OK
2017/01/31 10:14:21 logged in successful: 302 Found
2017/01/31 10:14:22 found 66 unread entries
2017/01/31 10:14:22 completed in 1.07s
2017/01/31 10:14:22 URL https://lib3.net/wallabag/export/23128.epub older than local file /home/anarcat/tmp/epubs2/23128.epub, skipped
2017/01/31 10:14:22 URL https://lib3.net/wallabag/export/19964.epub older than local file /home/anarcat/tmp/epubs2/19964.epub, skipped
2017/01/31 10:14:22 URL https://lib3.net/wallabag/export/23152.epub older than local file /home/anarcat/tmp/epubs2/23152.epub, skipped
2017/01/31 10:14:22 URL https://lib3.net/wallabag/export/23160.epub older than local file /home/anarcat/tmp/epubs2/23160.epub, skipped
2017/01/31 10:14:22 URL https://lib3.net/wallabag/export/23170.epub older than local file /home/anarcat/tmp/epubs2/23170.epub, skipped
2017/01/31 10:14:22 processed: 5, downloaded: 0
2017/01/31 10:14:22 completed in 1.53s
```

1.07 seconds for the API call, about 60% of the time here...

can this be improved?

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | _sort of_
| BC breaks?    | no
| Deprecations? | _maybe?_
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

`detail=metadata` will nullify the `content` field of entries in order to
make smaller responses.

`detail=full` keeps the former behavior, it sends the content of entries.
It's the default, for backward compatibility.

Fixes #2817

With this change, calls to `/api/entries?detail=metadata` will return `Entry` objects with `content: null`. Are you ok with that @wallabag/core?

### DIFF
--- a/src/Wallabag/CoreBundle/Repository/EntryRepository.php
+++ b/src/Wallabag/CoreBundle/Repository/EntryRepository.php
@@ -139,14 +139,29 @@ class EntryRepository extends EntityRepository
      * @param string $order
      * @param int    $since
      * @param string $tags
+     * @param string $detail     'metadata' or 'full'. Include content field if 'full'
+     *
+     * @todo Breaking change: replace default detail=full by detail=metadata in a future version
      *
      * @return Pagerfanta
      */
-    public function findEntries($userId, $isArchived = null, $isStarred = null, $isPublic = null, $sort = 'created', $order = 'asc', $since = 0, $tags = '')
+    public function findEntries($userId, $isArchived = null, $isStarred = null, $isPublic = null, $sort = 'created', $order = 'asc', $since = 0, $tags = '', $detail = 'full')
     {
+        if (!\in_array(strtolower($detail), ['full', 'metadata'], true)) {
+            throw new \Exception('Detail "' . $detail . '" parameter is wrong, allowed: full or metadata');
+        }
+
         $qb = $this->createQueryBuilder('e')
             ->leftJoin('e.tags', 't')
             ->where('e.user = :userId')->setParameter('userId', $userId);
+
+        if ('metadata' === $detail) {
+            $fieldNames = $this->getClassMetadata()->getFieldNames();
+            $fields = array_filter($fieldNames, function ($k) {
+                return 'content' !== $k;
+            });
+            $qb->select(sprintf('partial e.{%s}', implode(',', $fields)));
+        }
 
         if (null !== $isArchived) {
             $qb->andWhere('e.isArchived = :isArchived')->setParameter('isArchived', (bool) $isArchived);
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -103,6 +103,7 @@ class EntryRestController extends WallabagRestController
      *          {"name"="tags", "dataType"="string", "required"=false, "format"="api,rest", "description"="a list of tags url encoded. Will returns entries that matches ALL tags."},
      *          {"name"="since", "dataType"="integer", "required"=false, "format"="default '0'", "description"="The timestamp since when you want entries updated."},
      *          {"name"="public", "dataType"="integer", "required"=false, "format"="1 or 0, all entries by default", "description"="filter by entries with a public link"},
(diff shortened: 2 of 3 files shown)


src/Wallabag/CoreBundle/Repository/EntryRepository.php
```diff
@@ -139,15 +139,30 @@ public function getRawBuilderForUntaggedByUser($userId)
      * @param string $order
      * @param int    $since
      * @param string $tags
+     * @param string $detail
```
j0k3r: Explain what it does and which values are allowed

tcitworld:
Also let's ping app maintainers.

Kdecherf:
> I'm wondering if `simple` might no be clearer than `metadata` thinking

Hm well, with `metadata` you clearly don't expect to see any content whereas `simple` can be misleading imho.

j0k3r:
@Simounet @bourvill @di72nn @jlnostr what do you think?

bourvill:
useful

I think that metadata is not bad, all the entry except the content?

j0k3r:
> all the entry except the content?

Yup.

jlnostr:
LGTM, although I'm not maintaining the app anymore.

PR #3986, opened 2019-05-29
Fix some Scrutinizer issues

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| License       | MIT

Cleanup some issues from Scrutinizer.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -564,7 +564,7 @@ class EntryRestController extends WallabagRestController
         }
 
         // if refreshing entry failed, don't save it
-        if ($this->getParameter('wallabag_core.fetching_error_message') === $entry->getContent()) {
+        if ($this->container->getParameter('wallabag_core.fetching_error_message') === $entry->getContent()) {
             return new JsonResponse([], 304);
         }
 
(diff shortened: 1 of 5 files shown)


src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -564,7 +564,7 @@ public function patchEntriesReloadAction(Entry $entry)
         }
 
         // if refreshing entry failed, don't save it
-        if ($this->getParameter('wallabag_core.fetching_error_message') === $entry->getContent()) {
+        if ($this->container->getParameter('wallabag_core.fetching_error_message') === $entry->getContent()) {
```
tcitworld: If we start make cleaner we might as well inject parameters properly inside `services.yml`.

tcitworld: But that's for another PR…

j0k3r: We should inject service into controller too, etc. :)

j0k3r: But thah change was mostly because the parent is now extending `AbstractFOSRestController` instead of `FOSRestController` (which force us to use `$this->container` for parameter because the controller wasn't "container aware")

PR #3993, opened 2019-06-05
Show untagged entries count on tag list

### LINKED ISSUE #3235: Show article counter in untagged articles
Show article counter in untagged articles
2017-06-16 23_32_17-etiquetas wallabag

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| License       | MIT

Fix #3235

image

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/TagController.php
+++ b/src/Wallabag/CoreBundle/Controller/TagController.php
@@ -88,7 +88,7 @@ class TagController extends Controller
         $tags = $this->get('wallabag_core.tag_repository')
             ->findAllFlatTagsWithNbEntries($this->getUser()->getId());
         $untagged = $this->get('wallabag_core.entry_repository')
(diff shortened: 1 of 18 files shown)


tcitworld:
We might as well hide this line if there's no untagged articles ?

j0k3r:
> We might as well hide this line if there's no untagged articles ?

It might be troubling and I think with that addition (the counter), we don't need to remove the link, it's more obvious that we won't have entry if you click on it:
image

tcitworld:
> > We might as well hide this line if there's no untagged articles ?
> 
> It might be troubling and I think with that addition (the counter), we don't need to remove the link, it's more obvious that we won't have entry if you click on it:
> image

I would still like to rephrase to something like: « No untagged entries » and remove an useless link.

j0k3r:
Updated:
image

PR #3995, opened 2019-06-06
Display login in config / user

### LINKED ISSUE #3645: Username and login
### Issue details

Login username is not **Name** from `Config > User information` and if you change this one, and you don't remeber your login username… And you have no way to recover it (in web interface at least). So if a user want to log in with his username, 
Login is partially a problem since [we can login with email](https://github.com/wallabag/wallabag/pull/3615) but sill.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.2.2

* change your **Name** from `Config > User information`
* try to log in with it: you can't and if you don't remember your login username…

Maybe add an input "Login username" in `Config`?

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| License       | MIT

Just so the user can see its login

Fix #3645

image

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/translations/messages.da.yml
+++ b/src/Wallabag/CoreBundle/Resources/translations/messages.da.yml
@@ -102,6 +102,7 @@ config:
         # feed_limit: 'Number of items in the feed'
     form_user:
         # two_factor_description: "Enabling two factor authentication means you'll receive an email with a code OR need to use an OTP app (like Google Authenticator, Authy or FreeOTP) to get a one time code on every new untrusted connection. You can't choose both option."
+        # login_label: 'Login'
         name_label: 'Navn'
         email_label: 'Emailadresse'
         two_factor:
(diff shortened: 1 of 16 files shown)


tcitworld:
Maybe add something like `(Can't be changed)`?

j0k3r:
Well it's not an input field so there is no way to change it. Shouldn't this be enough?

tcitworld:
> Well it's not an input field so there is no way to change it.

Yeah, I suggested this to make it more explicit that it's normal that you can't edit this field in particular (by the way, we totally could allow it, right ?).

j0k3r:
> (by the way, we totally could allow it, right ?).

Yes of course and admin can do that from the user management menu.

src/Wallabag/CoreBundle/Resources/translations/messages.da.yml
```diff
@@ -102,6 +102,7 @@ config:
         # feed_limit: 'Number of items in the feed'
     form_user:
         # two_factor_description: "Enabling two factor authentication means you'll receive an email with a code OR need to use an OTP app (like Google Authenticator, Authy or FreeOTP) to get a one time code on every new untrusted connection. You can't choose both option."
+        # login_label: 'Login (can not be changed) (can not be changed)'
```
tcitworld: Typo

j0k3r:
Failing tests aren't related.

Issue #4007, opened 2019-06-12
update issue - composer not installed message (but it's there)


jkrobin:
I have a working installation of Wallabag 2.3.1 on a VPS hosted by Dreamhost that I wish to update to 2.3.8. Following the Upgrading from 2.3.x to 2.3.y instructions at https://doc.wallabag.org/en/admin/upgrade.html, I go to my directory and run 'make update' but get an error returned:

**"composer.phar not found, we'll see if composer is installed globally.
wallabag requires composer but it's not installed (see http://doc.wallabag.org/en/master/user/installation.html). Aborting.
make: *** [update] Error 1"**

Obviously, composer had to be there when I installed 2.3.1 and previous versions, which have been working great for a long time.

Suggestions/help needed and welcome! Thanks.

j0k3r:
You have to install composer on your machine.
https://getcomposer.org/download/

Launch command from the root of wallabag. It'll install the `composer.phar` file in that directory.
Then you can re-launch the update.

We'll have to update the doc about that point.

jkrobin:
Thank you. I got composer.phar installed and then ran make update. 

The update process generated multiple error and warning messages, but appeared to continue through to the end, with the two 'red' error messages both containing references to SQLite (which I am using). There are also several references to items being "already loaded in Unknown on line 0."

See the attached text output of the whole process, as well as a screenshot showing the colored output in the terminal.

After the process, nothing appears to have changed on my live site, and clicking on 'about' at the bottom of the live site still shows the installed version to be 2.3.1.
<img width="1222" alt="Screen Shot 2019-06-12 at 10 59 01 AM" src="https://user-images.githubusercontent.com/33295430/59362409-65910f80-8d01-11e9-8ee0-4a2bd827b971.png">
[full make update output.txt](https://github.com/wallabag/wallabag/files/3281908/full.make.update.output.txt)

j0k3r:
How did you install wallabag in the first place?
Which database are you using? Looks like it's SQLite

jkrobin:
I installed it manually back in the 2.2x series and it has been running great for a long time. Update to 2.3.x series was done following the shared hosting instructions (though, technically, I'm on a VPS at Dreamhost, not normal shared hosting). 

And, yes -- I am using SQLite.

j0k3r:
[...]

PR #4009, opened 2019-06-12
Check if .git directory exist before making the update

j0k3r:
If that folder doesn’t exist, it means the udpate script won’t be able to run because it uses git to retrieve the update.

Tests will fail because of it needs https://github.com/wallabag/wallabag/pull/4008 to be merged.
Related https://github.com/wallabag/wallabag/issues/4007

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -40,7 +40,7 @@ fi
 # Check for wallabag .git folder
 if [ ! -d .git ]; then
     echo "Can not update because wallabag wasn't installed using git (see https://doc.wallabag.org/en/admin/upgrade.html#upgrade-on-a-shared-hosting). Aborting.";
-    exit 1;
+    exit 2;
 fi
 
 rm -rf var/cache/*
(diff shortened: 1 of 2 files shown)


scripts/update.sh
```diff
@@ -37,6 +37,12 @@ fi
 
 . "$REQUIRE_FILE"
 
+# Check for wallabag .git folder
+if [ ! -d .git ]; then
+    echo "Can not update because wallabag wasn't installed using git (see https://doc.wallabag.org/en/admin/upgrade.html#upgrade-on-a-shared-hosting). Aborting.";
+    exit 1;
```
tcitworld: Special exit code maybe ?

j0k3r: Like 2 instead of 1?

tcitworld: For instance, just so we know where the issue comes when using an automatic system.

j0k3r: Okay

PR #4018, opened 2019-06-18
Fix migration with prefix table name

j0k3r:
Looks like we missed one `wallabag_user` in migration.
It can work for most people but if someone use a different prefix, it'll break.


j0k3r:
@Kdecherf it was related to my changes

Issue #4037, opened 2019-06-30
Difficulties to install wallabag


ggdagg:
### Environment

* wallabag version (or git revision) that exhibits the issue: tried with 2.3.7 and 2.3.8
* How did you install wallabag? Via `git clone` or by downloading the package? tried with git clone and package
* Last wallabag version that did not exhibit the issue (if applicable): -
* php version: PHP 7.3.6
* OS: CentOS Linux release 7.6.1810 (Core)
* type of hosting (shared or dedicated): dedicated
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MariaDB

### Error
At the end of the installation after creating the admin user the following is displayed:

[Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException]                                                                                                                                                             
  An exception occurred while executing 'INSERT INTO `wallabag_config` (theme, items_per_page, language, rss_token, rss_limit, reading_speed, pocket_consumer_key, action_mark_as_read, list_mode, user_id) VALUES (?, ?, ?,   
  ?, ?, ?, ?, ?, ?, ?)' with params ["material", 12, "en", null, 50, 1, null, 1, 0, 1]:                                                                                                                                        
                                                                                                                                                                                                                               
  SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`wallabag`.`wallabag_config`, CONSTRAINT `FK_87E64C53A76ED395` FOREIGN KEY (`user_id`) REFERENCES `  
  wallabag_user` (`id`))                                                                                                                                                                                                       
                                                                                                                                                                                                                          
  [Doctrine\DBAL\Driver\PDOException]                                                                                                                                                                                          
  SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`wallabag`.`wallabag_config`, CONSTRAINT `FK_87E64C53A76ED395` FOREIGN KEY (`user_id`) REFERENCES `  
  wallabag_user` (`id`))                                                                                                                                                                                                       
                                                                                                                                                                                                                 
  [PDOException]                                                                                                                                                                                                               
  SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`wallabag`.`wallabag_config`, CONSTRAINT `FK_87E64C53A76ED395` FOREIGN KEY (`user_id`) REFERENCES `  
[...]

PR #4039, opened 2019-07-01
Fix hardcoded `wallabag_` prefix in migrations

j0k3r:
Again.
Looks like there are still some which were missing to be replaced.

Found after checking migration after seeing https://github.com/wallabag/wallabag/issues/4037


tcitworld:
This is old... 😓

j0k3r:
@tcitworld you created the first one I did the last two :)

PR #4053, opened 2019-07-08
Add ability to manually define the reading speed

### LINKED ISSUE #4051: Custom reading speed
This is an enhancement request. Currently, the reading speed settings under config only allows values of 100, 200, 300 or 400 WPM. It would be better if the user could just input their own reading speed there, giving more accurate estimates. The jump between 100 and 200 WPM is especially big. The same article will be estimated to take either 15 or 30 minutes to read with no in-between.

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| License       | MIT

Fix #4051

Instead of using a select, let the user decide its own speed.
Improve the help message to explain what is 1.00 means in "WPM" (words per minute)

image

@Najrim what do you think?


tcitworld:
Why can't we just put the value in number words per minute ?

j0k3r:
@tcitworld because this is how it was defined 🤷‍♂️
```twig
{% set readingTime = entry.readingTime / app.user.config.readingSpeed %}
```

Should we add a new migration to migrate all current reading speed to WPM (I guess readingspeed * 200) and then update the way we calculate?

Najrim:
I think it would be way more user-friendly to do as @tcitworld suggests and let the user input the WPM and then just calculate the multiplication factor in the code.

tcitworld:
> @tcitworld because this is how it was defined

Yeah but for a user PoV all you had to do was to pick an option among WPM values, not enter a plain value with no clues.

No need for a migration, just do the math before saving the value.

j0k3r:
> No need for a migration, just do the math before saving the value.

We still to update all existing value in the database.

j0k3r:
The migration is still better to have a better consistency between what's in the database and what the form is displaying. It required a dirt hack to convert the value from the database into a WPM in the form.
As this will be in the upcoming 2.4 we can use a migration.

j0k3r:
Updated:

image

app/DoctrineMigrations/Version20190708122957.php
```diff
@@ -0,0 +1,22 @@
...
+use Wallabag\CoreBundle\Doctrine\WallabagMigration;
+
+/**
+ * Change reading speed value.
+ */
+final class Version20190708122957 extends WallabagMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('UPDATE ' . $this->getTable('config', true) . ' SET reading_time = reading_time*200;');
```
tcitworld: I'm not sure if it's an issue or not, but `;` isn't required at the end of SQL statements called like this.

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entries.html.twig
```diff
@@ -46,7 +46,7 @@
         <div id="entry-{{ entry.id|e }}" class="{% if listMode == 0 %}entry{% else %}listmode entry{% endif %}">
             <h2><a href="{{ path('view', { 'id': entry.id }) }}" title="{{ entry.title|e|raw }}">{{ entry.title | striptags | truncate(80, true, '…') | default('entry.default_title'|trans) | raw }}</a></h2>
 
-            {% set readingTime = entry.readingTime / app.user.config.readingSpeed %}
+            {% set readingTime = entry.readingTime / app.user.config.readingSpeed * 200 %}
```
tcitworld: No need for this anymore, right ?

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entry.html.twig
```diff
@@ -62,7 +62,7 @@
             {% endif %}
 
             <i class="tool icon icon-time">
-                {% set readingTime = entry.readingTime / app.user.config.readingSpeed %}
+                {% set readingTime = entry.readingTime / app.user.config.readingSpeed * 200 %}
```
tcitworld: Same.

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/_reading_time.html.twig
```diff
@@ -1,4 +1,4 @@
-{% set readingTime = entry.readingTime / app.user.config.readingSpeed %}
+{% set readingTime = entry.readingTime / app.user.config.readingSpeed * 200 %}
```
tcitworld: Same.

src/Wallabag/CoreBundle/Resources/views/themes/baggy/Entry/entries.html.twig
j0k3r: Yes we need it.
- readingTime = 5 (minutes)
- readingSpeed = 300 (WPM)
- readingTime/readingSpeed = 0,016666667 (minutes)
- readingTime/readingSpeed*200 = 3,333333333 (minutes)

j0k3r:
Squashed all commits into one

PR #4058, opened 2019-07-09
Copy client info to clipboard

### LINKED ISSUE #3969: Add copy icon on API client
### Issue details

Suggestion, on the page managing the client API, add in front of the clientSecret and clientId an icon in order to be able to copy the value.

This could save time during registration on various applications when the selection is tedious

Capture d’écran 2019-05-14 à 14 27 24

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| License       | MIT

From the listing page and the create summary page, you can now copy client info to the clipboard using dedicated buttons.

Fix #3969

Using https://clipboardjs.com/

Listing:
image

Created page:
image

Drawbacks:
1. I wasn't able to trigger a material tooltip to notice when the text is copied :disappointed: 
2. I wasn't able to use a link instead of a button to be able to use an icon (`content_copy`) instead of a text.


bourvill:
it's a small step for man but a giant leap for mankind' 🚀

app/Resources/static/themes/_global/index.js
```diff
@@ -107,4 +109,12 @@ $(document).ready(() => {
   $('#user_emailTwoFactor').on('change', () => {
     $('#user_googleTwoFactor').prop('checked', false);
   });
+
+  // handle copy to clipboard for developer stuff
+  const clipboard = new ClipboardJS('.btn');
+  clipboard.on('success', (e) => {
+    console.info(e.text);
```
bourvill: can be remove for prod?

j0k3r: I put that because I can't only use:

```js
new ClipboardJS('.btn');
```

Because ESLint don't allow it, see https://eslint.org/docs/rules/no-new
So I've to store it in a variable, but then I need to use that variable (otherwise ESLint isn't happy again). So I decided to use the `.on('success'` to use the variable

:man_shrugging:

tcitworld:
Maybe replace the text with 📋 emoji ?

bourvill:
btw as a reminder, the goal is to help diffilcuted to copy paste on mobile.
The screenshots are on desktop, what is happening on mobile?
Maybe the button should be placed at the beginning to avoid having to scroller

app/Resources/static/themes/_global/index.js
bourvill: console.info is required ?

j0k3r:
@bourvill here are some screenshots:
image
image
image

j0k3r:
> Maybe replace the text with  emoji ?
> 
> EDIT : Or just « Copy »

"Copy" might be better.
Using the emoji means we'll have an emoji as text for the button :thinking:

bourvill:
> @bourvill here are some screenshots:
> image
> image
> image

Great! I that it will help a lot of our users. Thank @j0k3r

nicosomb:
👍

PR #4063, opened 2019-07-24
Use a custom cookiejar to avoid error when the cookie is badly saved

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| License       | MIT

It happens sometimes on wallabag.it, the json inside the cookie is badly saved and the json isn't valid. It generates an exception and avoid people to use the api and import contents.

To fix that, we use a dedicated `FileCookieJar`, which extends the default one from Guzzle to fix these issues:
- line 31: add the `LOCK_EX`
- line 56/61: add a try catch when parsing the json from the cookie

Also updated deps

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1470,16 +1470,16 @@
         },
         {
             "name": "egulias/email-validator",
-            "version": "2.1.9",
+            "version": "2.1.10",
             "source": {
                 "type": "git",
                 "url": "https://github.com/egulias/EmailValidator.git",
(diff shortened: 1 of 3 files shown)


j0k3r:
@tcitworld I've added the logger

PR #4072, opened 2019-07-26
Test on PHP 7.4 instead of PHP 8.0 on Travis

j0k3r:
PHP 7.4 will be available before the 8.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ php:
     - 7.2
     - 7.3
     - 7.4
+    - nightly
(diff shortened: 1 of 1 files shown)


tcitworld:
Let's keep both ?

tcitworld:
phpenv doesn't seem to have 7.4 support though. Try `7.4snapshot` ?

j0k3r:
I prefer to keep 7.4 and as soon as it'll be available, it'll works :)

PR #4128, opened 2019-09-27
Added MathJax support

### LINKED ISSUE #1408: Support of mathjax
MathJax is a Javascript library which allows to view LaTeX (mathematical formulas) code inside a browser. The result is a combination of CSS, HTML and SVG.
More information here: https://en.wikipedia.org/wiki/MathJax
And here: https://www.mathjax.org/

It would be really nice to enable MathJax in order to be able to read mathematica articles in Wallabag.
With the use of the CDN of MathJax, it is pretty simple: http://docs.mathjax.org/en/latest/start.html

The following points could be a problem:
- Decide when to enable MathJax
- I don't really know how it works on Android, which could be a problem for an offline mode
- If one wants to use another CDN, it could be more complicated

If the Wallabag team want to implement that, I'm ready to help.

Best.

Rafael

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixed #1408 

Example with http://math.rgug.ch/2013/03/08/matrices-complementaires-et-produits/ 

Capture d’écran 2019-09-27 à 16 02 56


nicosomb:
@tcitworld done

PR #4151, opened 2019-10-09
✨ Allow custom styles system wide

### LINKED ISSUE #4060: Allow Custom Syles (site wide)
### Issue details
Related to #1521 and #1278 it would be great to be able to add a custom stylesheet to override the default styles. For example, store `custom.css` outside of the core folder then have it loaded within wallabag. In my case I like the Material theme but want to make a few customizations such as changing the logo, a few colors, etc.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.3.8
* How did you install wallabag? Via `git clone` or by downloading the package? `git clone`
* php version: 7.3.7
* OS: Debian 10
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): MariaDB 10.14.6

ldidry:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | yes
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Should fix #4060

- add a setting called `custom_css` in `app/config/parameters.yml`, available for twig
- when set to true, Wallabag calls http://your-wallabag/custom.css on every page
- put a file named `custom.css` in `web/` directory: it will be used by Wallabag.

Screenshot_2019-10-09 Articles non lus – wallabag(2)

The green sidebar of the screenshot is created by `custom.css`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ app/Resources/build/
 admin-export.json
 specialexport.json
 /data/site-credentials-secret-key.txt
+
+# Custom CSS file
+web/custom.css
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -10,14 +10,14 @@ help: ## Display this help menu
 clean: ## Clear the application cache
 	rm -rf var/cache/*
 
-install: ## Install wallabag with the latest version
+install: customcss ## Install wallabag with the latest version
 	@./scripts/install.sh $(ENV)
 
 update: ## Update the wallabag installation to the latest version
 	@./scripts/update.sh $(ENV)
 
 dev: ENV=dev
-dev: build ## Install the latest dev version
+dev: build customcss ## Install the latest dev version
 	@./scripts/dev.sh
 
 run: ## Run the wallabag built-in server
@@ -26,6 +26,9 @@ run: ## Run the wallabag built-in server
 build: ## Run webpack
 	@npm install
 	@npm run build:$(ENV)
+
+customcss:
+	@touch web/custom.css
 
 prepare: clean ## Prepare database for testsuite
 ifdef DB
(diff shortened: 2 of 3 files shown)


ldidry:
@j0k3r If you think it’s better, I’m okay with that. I’ll change that tonight.

ldidry:
Unfortunately, once you have commit a file, `.gitignore` does not work on it anymore. There is some ways to ignore modifications to such a file, but it only works locally.

I can try to create a (twig extension to check if the file exists](https://stackoverflow.com/questions/16898628/symfony2-and-twig-check-if-an-asset-exists) though, but it will take me some time.

j0k3r:
You just have to do git add -f web/custom.css

ldidry:
But then, when people add rules in it, git will say that the file has changed, it won’t be ignored.

j0k3r:
Try it it’ll be ignored

j0k3r:
Wow I was sure it works. 😞 
Anyway, we should then add a `web/custom.css.dist` (with comment inside about what the purpose of that file and how to use it), ignore `web/custom.css` and explain in the doc that we should `cp web/custom.css.dist web/custom.css` to enable custom css.

But then is it acceptable to always load the `web/custom.css` even when it does not exist?
poke @wallabag/core

ldidry:
> But then is it acceptable to always load the web/custom.css even when it does not exist?

I think not: power user will ask them-self why they have a 404 call on each page.

But I thought about another way: what about `touch web/custom.css` on `make install` ?

j0k3r:
That might do the trick

ldidry:
Should I put that in `GNUmakefile` or in `scripts/install.sh`?

ldidry:
I’ll put it in GNUmakefile as a new target: we need that for `make dev too`.

ldidry:
Done

PR #4254, opened 2020-01-19
Sidebar sub entries not conditioned by hover anymore

Simounet:
When you opened a sub entry from the sidebar (Share for example), it was closed when loosing focus. On mobile, it was annoying because we should open it again after getting back to the wallabag screen.
I think this is more predictable. If the user did an action, we should remove it until he wants to. Now, you have to click a second time to close it.
Have a nice day.


j0k3r:
Looks ok but can you share a gif before/after of the behavior?

Simounet:
Before:

before

After:

after

PR #4257, opened 2020-01-21
Use lang attribute

Simounet:
Another quick one using the app language for the main html tag and the grabbed site one for our entry content.

### DIFF
--- a/src/Wallabag/CoreBundle/Entity/Entry.php
+++ b/src/Wallabag/CoreBundle/Entity/Entry.php
@@ -788,6 +788,19 @@ class Entry
     }
 
     /**
+     * Format the entry language to a valid html lang attribute.
+     */
+    public function getHTMLLanguage()
+    {
+        $parsedLocale = \Locale::parseLocale($this->getLanguage());
+        $lang = '';
+        $lang .= $parsedLocale['language'] ?? '';
+        $lang .= isset($parsedLocale['region']) ? '-' . $parsedLocale['region'] : '';
+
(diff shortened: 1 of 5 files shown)


nicosomb:
You can maybe add a new test 😉

src/Wallabag/CoreBundle/Controller/EntryController.php
```diff
@@ -270,6 +270,7 @@ public function redirectRandomEntryAction($type = 'all')
     public function viewAction(Entry $entry)
     {
         $this->checkUserAction($entry);
+        $entry->setLanguage($entry->getHTMLLanguage());
```
j0k3r: I'm not a big fan of overriding the already saved data with something else only when rendering the entry.

Instead, you should keep your `getHTMLLanguage` and use it directly in the twig using `entry.html_language` (I'm not sure about the `_`)

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -787,6 +787,16 @@ public function getLanguage()
         return $this->language;
     }
 
+    public function getHTMLLanguage()
```
j0k3r: Provide a docblock to explain the purpose of that method

Simounet:
Thanks for not letting me be a lazy boy. :D

j0k3r:
You need to run the fixer locally: `php bin/php-cs-fixer fix`

Simounet:
Done.

PR #4272, opened 2020-01-29
Load custom.css only if exists

Simounet:
Hello there,
I get the point of this custom.css file but I think we should avoid requesting files when they are empty. I would prefer writing documentation about this rather then doing an extra HTTP request.
If this PR is okay for you, let me know the best place to write about it in the doc repo.

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -16,6 +16,7 @@ services:
             - "@security.token_storage"
             - "%wallabag_core.cache_lifetime%"
             - "@translator"
+            - "%kernel.root_dir%"
         tags:
             - { name: twig.extension }
 
--- a/src/Wallabag/CoreBundle/Twig/WallabagExtension.php
+++ b/src/Wallabag/CoreBundle/Twig/WallabagExtension.php
@@ -18,14 +18,16 @@ class WallabagExtension extends AbstractExtension implements GlobalsInterface
     private $tagRepository;
     private $lifeTime;
     private $translator;
+    private $rootDir;
 
-    public function __construct(EntryRepository $entryRepository, TagRepository $tagRepository, TokenStorageInterface $tokenStorage, $lifeTime, TranslatorInterface $translator)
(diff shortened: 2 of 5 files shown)


app/config/services.yml
```diff
@@ -16,6 +16,7 @@ services:
             - "@security.token_storage"
             - "%wallabag_core.cache_lifetime%"
             - "@translator"
+            - "@kernel"
```
j0k3r: Use directly the variable instead of injecting the whole class.

```suggestion
            - "%kernel.root_dir%"
```

src/Wallabag/CoreBundle/Twig/WallabagExtension.php
```diff
@@ -165,6 +169,12 @@ public function displayStats()
         ]);
     }
 
+    public function assetFileExists($name)
+    {
+        $fileExists = \file_exists(realpath($this->kernel->getRootDir() . '/../web/' . $name));
```
j0k3r:
```suggestion
        return \file_exists(realpath($this->kernel->getRootDir() . '/../web/' . $name));
```

src/Wallabag/CoreBundle/Twig/WallabagExtension.php
```diff
@@ -165,6 +169,12 @@ public function displayStats()
...
+        return $fileExists;
```
j0k3r:
```suggestion
```

Simounet:
Hmmm, I was expecting this PR to be merged. Can you tell me what to do? Do I really have to test a `file_exists()` method from a template?

Kdecherf:
@Simounet no, sorry, I was using the `Missing: tests` label to indicate that tests are failing.

Could you fix them?

Simounet:
Oh, of course! Sorry, I didn't get the label purpose right. :p

Simounet:
Fixed.

Kdecherf:
> Oh, of course! Sorry, I didn't get the label purpose right. :p

Well, consider that \*I\* didn't get the label purpose right :D

PR #4302, opened 2020-03-28
Replace grandt/phpepub with wallabag version

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | -
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -73,7 +73,7 @@
         "friendsofsymfony/oauth-server-bundle": "^1.5",
         "stof/doctrine-extensions-bundle": "^1.2",
         "scheb/two-factor-bundle": "^4.4",
-        "grandt/phpepub": "dev-master",
+        "wallabag/phpepub": "^4.0.7.2",
         "wallabag/php-mobi": "~1.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -73,7 +73,7 @@
         "friendsofsymfony/oauth-server-bundle": "^1.5",
         "stof/doctrine-extensions-bundle": "^1.2",
         "scheb/two-factor-bundle": "^4.4",
-        "grandt/phpepub": "dev-master",
+        "wallabag/phpepub": "^4.0.7.2",
```
j0k3r:
```suggestion
        "wallabag/phpepub": "^4.0",
```

Kdecherf: I don't agree with the version constraint change: there is a certain amount of changes since 4.0 (_and since the last 4.0.7.1_), maybe we should soon make a major bump.

Kdecherf:
> Why is it still WIP?

I still need to commit the resulting change on the lock file.

PR #4310, opened 2020-04-04
TagController: fix duplicated tags when renaming them

### LINKED ISSUE #4216: Renaming a tag removes the relation between tag and entry
### Issue details

I tried the renaming tags feature. The tag is renamed but the relation between tag and entry is removed. 

### Environment

* wallabag version (or git revision) that exhibits the issue: master
* php version: 7.3
* OS: debian 
* type of hosting (shared or dedicated): wallabag.it

You can't reproduce it on wallabag.it because I disabled this feature to avoid bad experience for users.

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | 
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

The fix relies on a workaround available on TagsAssigner, see the
AssignTagsToEntry() signature for detail.

I replaced the findOneByLabel in the corresponding test to assert that
there is no duplicate.

Fixes #4216

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/TagController.php
+++ b/src/Wallabag/CoreBundle/Controller/TagController.php
@@ -151,7 +151,10 @@ class TagController extends Controller
         $form = $this->createForm(RenameTagType::class, new Tag());
         $form->handleRequest($request);
 
-        if ($form->isSubmitted() && $form->isValid()) {
+        if ($form->isSubmitted()
+            && $form->isValid()
+            && $form->get('label')->getData() !== $tag->getLabel()
+           ) {
             $newTagLabel = $form->get('label')->getData();
             $newTag = new Tag();
             $newTag->setLabel($newTagLabel);
@@ -171,12 +174,12 @@ class TagController extends Controller
 
             $em = $this->getDoctrine()->getManager();
             $em->flush();
-        }
 
-        $this->get('session')->getFlashBag()->add(
-            'notice',
-            'flashes.tag.notice.tag_renamed'
-        );
+            $this->get('session')->getFlashBag()->add(
+                'notice',
+                'flashes.tag.notice.tag_renamed'
+            );
+        }
 
         $redirectUrl = $this->get('wallabag_core.helper.redirect')->to($request->headers->get('referer'), '', true);
 
--- a/src/Wallabag/CoreBundle/Controller/TagController.php
+++ b/src/Wallabag/CoreBundle/Controller/TagController.php
@@ -151,13 +151,21 @@ class TagController extends Controller
         $form = $this->createForm(RenameTagType::class, new Tag());
         $form->handleRequest($request);
 
-        if ($form->isSubmitted()
-            && $form->isValid()
-            && $form->get('label')->getData() !== $tag->getLabel()
-           ) {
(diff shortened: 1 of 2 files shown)


Kdecherf:
Question @wallabag/core:

If we rename a tag with the same label than another existing tag, should we merge them together or should we keep both?

j0k3r:
We should merge I think

Kdecherf:
@j0k3r changes were made to merge tags when renaming a tag with the same label of an existing tag.

src/Wallabag/CoreBundle/Controller/TagController.php
```diff
@@ -152,27 +152,40 @@ public function renameTagAction(Tag $tag, Request $request)
...
-            );
-            foreach ($entries as $entry) {
-                $this->get('wallabag_core.tags_assigner')->assignTagsToEntry(
-                    $entry,
-                    $form->get('label')->getData()
+            $formLabel = $form->get('label')->getData();
+            $newTag = new Tag();
+            $newTag->setLabel($formLabel);
+
+            if ($newTag->getLabel() !== $tag->getLabel()) {
```
j0k3r, line 159: I tend to avoid using condition which generate too much code indented below.

Maybe you can move the redirect url definition above that condition, flip the check to have a small part in the `if` and return early:

```php
$redirectUrl = $this->get('wallabag_core.helper.redirect')->to($request->headers->get('referer'), '', true);

if ($newTag->getLabel() === $tag->getLabel()) {
    return $this->redirect($redirectUrl);
}

// whole code about renaming

return $this->redirect($redirectUrl);
```

src/Wallabag/CoreBundle/Controller/TagController.php
```diff
@@ -152,27 +152,40 @@ public function renameTagAction(Tag $tag, Request $request)
...
-                $this->getUser()->getId(),
-                $tag->getId()
-            );
-            foreach ($entries as $entry) {
-                $this->get('wallabag_core.tags_assigner')->assignTagsToEntry(
-                    $entry,
-                    $form->get('label')->getData()
+            $formLabel = $form->get('label')->getData();
+            $newTag = new Tag();
+            $newTag->setLabel($formLabel);
```
j0k3r, line 157: This will avoid having to use a temporary variable

```suggestion
            $newTag->setLabel($form->get('label')->getData());
```

src/Wallabag/CoreBundle/Controller/TagController.php
```diff
@@ -152,27 +152,40 @@ public function renameTagAction(Tag $tag, Request $request)
...
+                    $this->get('wallabag_core.tags_assigner')->assignTagsToEntry(
+                        $entry,
+                        $newTag->getLabel(),
+                        [$newTag]
+                    );
+                    $entry->removeTag($tag);
+                }
+
+                $em = $this->getDoctrine()->getManager();
+                $em->flush();
```
j0k3r, line 180: Can this be enough?

```suggestion
                $this->getDoctrine()->getManager()->flush();
```

PR #4316, opened 2020-04-06
Fixed bug when we delete entry via API

### LINKED ISSUE #4291: Downloaded images are not removed after deleting an article using the API
### Issue details

When deleting an article, an event is fired to trigger things elsewhere like the removal of locally downloaded images. It appears that the removal does not work when the article is removed using the API (_e.g. from the Android app_) which leads to an accumulation of orphan files.

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.3.8
* Last wallabag version that did not exhibit the issue (if applicable): -
* php version: 7.3
* OS: debian
* which storage system you choose at install (SQLite, MySQL/MariaDB or PostgreSQL): PostgreSQL 9.6

1. Enable local download of images
2. Add an article with images
3. Remove the article using the API
4. Look at the folder of the downloaded images for this article

nicosomb:
Fixes #4291

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

The event was triggered after the entry deletion, so wallabag had no way to find entry data.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -239,12 +239,12 @@ class EntryRestController extends WallabagRestController
             $results[$key]['url'] = $url;
 
             if (false !== $entry) {
+                // entry deleted, dispatch event about it!
+                $this->get('event_dispatcher')->dispatch(EntryDeletedEvent::NAME, new EntryDeletedEvent($entry));
+
                 $em = $this->getDoctrine()->getManager();
                 $em->remove($entry);
                 $em->flush();
-
-                // entry deleted, dispatch event about it!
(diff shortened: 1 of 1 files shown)


Kdecherf:
I'm wondering if we should keep the event dispatching *before* actually removing the entry. What would happen if the removal fails but the event has been successfully triggered?

Also we should add tests in the CoreBundle and ApiBundle to ensure that DeletedEvent is correctly dispatched.

nicosomb:
You're right. This PR is just a little patch, to have the same behavior in the API than in the front office. 

Maybe we can have a new issue for this improvement?

PR #4318, opened 2020-04-07
Changed authentication order in GrabySiteConfigBuilder

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes/no
| CHANGELOG.md  | no
| License       | MIT

Today, when we are in a Symfony Command (for example, `ImportCommand`), if we want to save an URL for paywalled websites, as we are not authenticated, the whole content is not saved. 

So we need to improve the Command, to create a Token Storage. 

But, when we create it, Graby losed it because the current user in `GrabySiteConfigBuilder` is initialized in the constructor. 

My PR is here to change this behavior. 

@wallabag/core need your review

### DIFF
--- a/src/Wallabag/CoreBundle/GuzzleSiteAuthenticator/GrabySiteConfigBuilder.php
+++ b/src/Wallabag/CoreBundle/GuzzleSiteAuthenticator/GrabySiteConfigBuilder.php
@@ -8,7 +8,6 @@ use Graby\SiteConfig\ConfigBuilder;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
 use Wallabag\CoreBundle\Repository\SiteCredentialRepository;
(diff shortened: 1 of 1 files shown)


src/Wallabag/CoreBundle/GuzzleSiteAuthenticator/GrabySiteConfigBuilder.php
```diff
@@ -131,4 +135,11 @@ protected function processExtraFields($extraFieldsStrings)
 
         return $extraFields;
     }
+
+    private function initUser()
+    {
+        if ($this->token->getToken()) {
```
j0k3r, line 141:
```suggestion
        if ($this->token->getToken() && null !== $this->token->getToken()->getUser()) {
```

nicosomb:
Done.

PR #4309, opened 2020-04-04
Replace wallabag's fork of tcpdf with the original one, PHP 7.4 support

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | ~
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This PR replaces our fork of TCPDF with the original version and fixes some issues seen with PHP 7.4.

Note: I currently use the `master` version of Graby to run the tests, we may need to release a new version to include the change for the TCPDF dependency. (cc @j0k3r)

PR #4320, opened 2020-04-08
PHP 7.4 should not be allowed to fail from CI

j0k3r:
Also, ensure we are not using default value from Travis to avoid any future migration from them.


Kdecherf:
This change is also proposed in #4309

Issue #3358, opened 2017-09-21
Basic edition capabilities


eemantsal:
Would be nice to be able to do some basic edition in our saved articles, in the articles' body, not just in the title, as is already possible.
For those articles we want to keep, most of the time we aren't interested in all the saved text or images, and would be nice to have the possibility of deleting or writing some text and delete/move embedded elements.

nicosomb:
Hello @eemantsal. 

Did you see that you can annotate an article? 
Mayve this feature could be interesting for you.

eemantsal:
Hi, @nicosomb 

Yes, I saw it, and is an awesome feature. I'm one of those readers who have their books' margins full of annotations made with pencil, so I sincerely love it. But I meant text edition, not in the sense a word processor would do, with tons of features, but very basic edition possibilities, like deleting or copying and pasting. Many times we just want to keep a part of the article, not the entire one: how many times a blog post begins with a long introduction we aren't insterested in at all, or the author begins to divagate in the middle of the article, or ar the end...? I'm sure you have experienced what I mean.
Would be nice if we could select what we want to save or if we could log in our WB account and edit those articles to delete every section we aren't insterested in.

gantheaume:
I was about to open a new issue, when I found that one, sorry for the resurrection.
I think it would be a very useful feature for an extra reason, I just bumped an example today:

**Some sites have additional popups**, like anti-add-blocking messages **that prevent from saving the article**, and I can guess how hard it can be to adjust to every site that does this (is there some kind of adjustment work done by wallabag actually? I'm not sure).
**But I can easily access the text myself** from the browser and **if I could edit the bagged article**, I could prevent this kind of situation from stopping me to save any article if the regular way failed, thus having a maybe temporary but immediate **quick fix of the problem.**
So I know that ideally, I should have a subscription, and connect with it to save the article, but I'm not a big online reader and quite often, the articles I want to save are from sites I don't visit over periods of months.

My today's example:
I wanted to save this article from the Washington post: https://www.washingtonpost.com/entertainment/trump-speaks-of-images-hes-never-seen-before-but-theyve-existed--we-just-refused-to-look/2020/04/09/89128fec-79c8-11ea-a130-df573469f094_story.html
But in the wallabag save, it just shows the text of the popup asking to disable add-blockers.

[...]

PR #4334, opened 2020-04-15
Editing Entry function using TinyMCE in Material

3bbing:
Adds support for changing the HTML content of an Entry with TinyMCE in Material Theme.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT & GNU LGPL from TinyMCE

Adresses #3871 , #3358 , #2321 

It is an alternative to commit #3871 using TinyMCE rather than CKEditor. Initially inspired by @slavniyteo comment at #2321 . 
I understand maintainers might not want to include it as discussed in #3871 and that's fine. Also, I am not sure about licensing issues integrating GNU LGPL into MIT. However, for those who want to add this functionality on their own, I just wanted to share this commit which I deem to be a quick addition of the requested feature. Should be easy to add for those who really want it.
Implementation also works post composing & thus on managed hosts too. As always, clear cache.

Changes:
Added TinyMCE 5.2.1 into web resources folder as described in Symfony 3.3. Best Practices https://symfony.com/doc/3.3/best_practices/web-assets.html
(GNU LGPL license, again, hope there is no conflict with MIT)
If you replicate this, maybe get the latest version from TinyMCE repo again and paste it into `web/js` (or use existing instance)

Added content edit field in Form,
`src/Wallabag/CoreBundle/Form/Type/EditEntryType.php`
changed View of material theme to include it:
To have it loading at then end of page, so that the page won't get lazy,
added javascripts block at the end of base layout twig.
`src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/edit.html.twig`
Activated loading TinyMCE in the edit View twig.
`src/Wallabag/CoreBundle/Resources/views/base.html.twig`

A switch to turn the edit mode on or off might come in handy if people are expected to not like or want the feature. This could be added by putting an {% if setting %} into the twigs. Assuming this is only for those who really want it though, I did not bother with it for obvious reasons.


src/Wallabag/CoreBundle/Form/Type/EditEntryType.php
```diff
@@ -30,6 +30,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'label' => 'entry.edit.origin_url_label',
                 'default_protocol' => null,
             ])
+            ->add('content', TextareaType::class, [
```
j0k3r, line 33: You should add `use Symfony\Component\Form\Extension\Core\Type\TextareaType` at the top of that file otherwise the page will display an error

src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/edit.html.twig
```diff
@@ -44,3 +49,13 @@
     </div>
 
 {% endblock %}
+
+{% block javascripts %}
+    <script src="{{ asset('js/tinymce/tinymce.min.js') }}"></script>
+    <script type="text/javascript">
+        tinymce.init({
+        selector: '#entry_content',
+        menubar: 'edit view format'
```
j0k3r, line 58:
```suggestion
            selector: '#entry_content',
            menubar: 'edit view format'
```

src/Wallabag/CoreBundle/Form/Type/EditEntryType.php
3bbing: Oh shit, forgot! Thanks! Added.

3bbing:
> Can't TinyMCE be added in the `package.json` file instead?
> Just to avoid having all those `*.min.js` file in the repo

I agree, should be nicer. 
I honestly got confused with all the symfony ways to include web assets (Webpack Encore, just into the Web folder, Assetic, custom build packages like https://github.com/stfalcon/TinymceBundle .. there might be more and I did not know which way you usually go). So I went with the 3.3 Best Practice as it seemed easiest and on the fly to me. But indeed it looks ugly. Will look into it in the evening, thanks.

src/Wallabag/CoreBundle/Form/Type/EditEntryType.php
j0k3r: And don't forget to run `php bin/php-cs-fixer fix` to resolve coding standard issue

nicosomb:
I'm not sure that we should have this feature in wallabag ...

3bbing:
> Hello there,
> 
> I think this feature can reduce the security of an instance as the edited content could embed javascript code that could be rendered publicly [with a public link]. Then I would require an instance-wide switch to enable/disable this feature (_and I would see it disabled by default_).

In theory [TinyMCE claims to strip scripts](https://www.tiny.cloud/docs/advanced/security/#scriptsandxssvulnerabilities). But sure it adds the possibility add a vulnerability to bypass that. can have code in  changed title though too. Would it make sense to rerun it through graby cleanup?

It gets more complicated than I intended it to become. With yours and the doubt of @nicosomb and considering it's easy to manually add if really wanted but impossible to remove if unwanted I'm fine to just have it in mind as manual change for those who want it rather than as permanently added feature. 
Still, it helped for my workflow with many pdfs and library-vpn protected material that I can now add instead of just having stubs in the bag. 

Thanks for your help @j0k3r !

michi-zuri:
Thank you so much for sharing this! I wish I had found it earlier. I definitely consider this a missing feature of wallabag. I hope it will make into the the main release one day, but a manual patch worked as you described.

lasc:
That would be a nice feature, just disabled by default?

PR #4363, opened 2020-04-28
Upgrade Piwik Twig Extension to Matomo

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | yes
| CHANGELOG.md  | no
| License       | MIT


src/Wallabag/CoreBundle/Resources/views/base.html.twig
```diff
@@ -76,7 +76,7 @@
         {% block footer %}{% endblock %}
 
         {% if craue_setting('piwik_enabled') %}
-            {{ piwik(craue_setting('piwik_host'), craue_setting('piwik_site_id')) }}
+            {{ matomo(craue_setting('piwik_host'), craue_setting('piwik_site_id')) }}
```
j0k3r, line 79: I was wondering if we should change them too: `piwik_host`, `piwik_site_id` & `piwik_enabled ` ?

nicosomb: It's only because we need a migration ... that's why I didn't write it.

app/config/wallabag.yml
```diff
@@ -125,15 +125,15 @@ wallabag_core:
...
         -
-            name: piwik_enabled
+            name: matomo_enabled
             value: 0
             section: analytics
         -
-            name: piwik_host
-            value: v2.wallabag.org
+            name: matomo_host
+            value: matomo.wallabag.org
```
nicosomb, line 133: I changed this line because in fact, we need to configure the URL of the matomo instance there.

app/DoctrineMigrations/Version20200428072628.php
```diff
@@ -0,0 +1,27 @@
...
+use Wallabag\CoreBundle\Doctrine\WallabagMigration;
+
+/**
+ * Renamed Piwik to Matomo in configuration 
+ */
+final class Version20200428072628 extends WallabagMigration
+{
+    public function up(Schema $schema) : void
+    {
+        $this->addSql('UPDATE ' . $this->getTable('internal_setting', true) . ' SET name = "matomo_enabled" where name = "piwik_enabled";');
```
Kdecherf, line 15: On PostgreSQL you can't use `"` to enclose values as it's restricted to column and table names. You must change `"` to `'`, see failing tests.

Kdecherf:
(also cs-fixer is failing)

j0k3r:

PR #4368, opened 2020-04-29
Ajax

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

I will help for controller part (@j0k3r @K-Phoen the EntryController need to be improved) and for the JS part (@tcitworld @Simounet). Maybe we need a single js file for this feature. 
About the JS part, I know that there are improvements to do. But it's just a POC, I need you to improve it. 

When you'll be ok for each part, I'll work on the other features (fav, delete).


src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
```diff
@@ -67,9 +67,9 @@
         {% endif %}
 
         <li class="bold hide-on-med-and-down">
-            <a class="waves-effect collapsible-header markasread" title="{{ markAsReadLabel|trans }}" href="{{ path('archive_entry', { 'id': entry.id }) }}" id="markAsRead">
-                <i class="material-icons small">{% if entry.isArchived == 0 %}done{% else %}unarchive{% endif %}</i>
-                <span>{{ markAsReadLabel|trans }}</span>
+            <a class="waves-effect collapsible-header markasread" title="{{ markAsReadLabel|trans }}" href="#" data-icon-read="done" data-icon-unread="unarchive" data-label-read="{{ 'entry.view.left_menu.set_as_read'|trans }}" data-label-unread="{{ 'entry.view.left_menu.set_as_unread'|trans }}" id="link-archive">
```
Kdecherf, line 70: Replacing the value of `href` will definitely break browsers disabling javascript. You should prefer something based on `preventDefault()` to offer a "backward" compatibility.

app/Resources/static/themes/material/index.js
```diff
@@ -120,4 +120,30 @@ $(document).ready(() => {
...
   }
+
+  $('.markasread').on('click', () => {
+    const article = document.getElementById('article');
+    const link = document.getElementById('link-archive');
+    const articleId = article.dataset.id;
+    const xhr = new XMLHttpRequest();
+
+    xhr.onload = function toggleArchive() {
+      if (xhr.status === 200) {
```
Kdecherf, line 131: Is there an icon indicating that a request is in progress? What if the request fails?

PR #4421, opened 2020-05-28
Upgrade scheb/two-factor-bundle to fix vulnerability

j0k3r:
The bundle was locked to a version which contains a security bug.
Jumping to the most recent fix the issue.

But we'll have the `ListenerInterface` issue, again.
So, downgrading sentry-symfony instead.

See:
- https://github.com/wallabag/wallabag/pull/4416/commits/0cf743f6fbfbc449c599ee0271e4e62ece0e47ac
- #4416
- #4419


j0k3r:
Rebased & force pushed to fix conflicts

PR #4523, opened 2020-08-25
Upgrade to NodeJS 12

j0k3r:
Update javascript deps to latest version.
Rebuild assets

- Keep `stylehint` in a lower version because starting at v9.0.0 it generates too much errors. We might fix them later.
- Don't touch `materialize-css` to avoid too much side effect. We might update it later (to 1.0.0).

This might also fix a lot of security issues.


tcitworld (approved):
Tested and works.

PR #4617, opened 2020-10-03
Check user credentials using imap authentication

exeba:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This commit allows me to authenticate users using an imap server, the same way it is done in Owncloud with the External User Backends app.

This is the updated login workflow:
  1) Submit credentials
  2) Local user exists -> 3), else 4)
  3) Local password match -> Success, else -> 4)
  4) Imap server refuses credentials -> Fail, else 5)
  5) Create user if needed
  6) Success

In principle it could be extended with other authentication methods.
If you think this is interesting, I'd be happy to work on it.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -36,6 +36,7 @@ class AppKernel extends Kernel
             new OldSound\RabbitMqBundle\OldSoundRabbitMqBundle(),
             new Http\HttplugBundle\HttplugBundle(),
             new Sentry\SentryBundle\SentryBundle(),
+            new Seb\AuthenticatorBundle\SebAuthenticatorBundle(),
 
             // wallabag bundles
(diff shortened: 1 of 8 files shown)


j0k3r:
If it works for you, I'm really happy.
But I'm not sure this will be in the main wallabag app. The main reason is that not everyone will need that kind of connexion (like the ldap stuff see https://github.com/wallabag/wallabag/pull/3263).
We should provide this kind of stuff as a bundle, or an extension but it's not possible right now in wallabag.

PR #4726, opened 2020-11-11
Fix duplicated icons

Simounet:
This reverts commit 07dab0dbc9b86171da78572cc2903f3fb90c721d.
Hi don't get the initial commit. Maybe something changed on materializecss or I could be wrong but right now, we have to choose between the class and the tag content.

Before
wallabag-before

After
wallabag-after


j0k3r (approved):
Thanks!!

PR #4747, opened 2020-11-23
Ditch Travis to use GitHub Actions

j0k3r:
Jump to GitHub Actions mostly because I found them most reliable these days (weeks?) than Travis.

Main inspiration from https://github.com/api-platform/core/blob/master/.github/workflows/ci.yml

About the whole stuff:
- mysql & pgsql aren't retrieve from Docker image because they are already defined in `ubuntu-16.04`. We can't force the version but we don't really care for now
- `PGPASSWORD` env var is defined globally so when we use command line to create pgsql database, we don't have to provide the password
- rabbitmq & redis port are forced to avoid GitHub to assign a random port


j0k3r:
@wallabag/core any suggestions / remarks about how I split the Travis build?

tcitworld (dismissed):
Only thing I've noticed are the errors like
```
SQLSTATE[HY000] [1045] Access denied for user 'root'@'localhost' (using pas  
  sword: NO)
```
in the install deps with Composer step since we don't have the `app/config/parameters_test.yml` file setup yet but shouldn't matter much.

j0k3r:
@tcitworld weird, because we didn't have that error in Travis and the `composer install` was also done before the `make prepare` 🤔

Kdecherf (commented):
Looks good.
Two notes though:
* It could be interesting to move to ubuntu-18.04 as the 16.04 edition will soon enter in extended security maintenance
* It could be interesting to mark the Assets workflow as failed if the produced assets are different from those currently committed.

j0k3r:
> It could be interesting to move to ubuntu-18.04 as the 16.04 edition will soon enter in extended security maintenance

Good idea, I stayed at 16.04 because 20.04 doesn't provide tidy as default extension for PHP 7.2 & 7.3. But it seems 18.04 does! https://github.com/shivammathur/setup-php/wiki/Php-extensions-on-ubuntu-18.04

> It could be interesting to mark the Assets workflow as failed if the produced assets are different from those currently committed.

I'm not sure about that.
What if a javascript deps change something and assets needs to be rebuild? We'll have to manually rebuilt them to integrate the change? (I'm just asking myself too)

Kdecherf:
> > It could be interesting to mark the Assets workflow as failed if the produced assets are different from those currently committed.
> 
> I'm not sure about that.
> What if a javascript deps change something and assets needs to be rebuild? We'll have to manually rebuilt them to integrate the change? (I'm just asking myself too)

I would say yes

nicosomb:
Good idea.

j0k3r:
@Kdecherf what about adding that action in the assets build in case change are generated? https://github.com/marketplace/actions/git-auto-commit

Kdecherf:
> @Kdecherf what about adding that action in the assets build in case change are generated? https://github.com/marketplace/actions/git-auto-commit

I would avoid any auto commit action, I think we should investigate in case a change occurs to prevent any malicious act nor missing stuff.

j0k3r:
We need actions now because Travis doesn't work anymore.
image

j0k3r:
I've disabled the asset check for now until we fix it properly.

tcitworld (approved):

Kdecherf (approved):

PR #4754, opened 2020-11-25
public assets files removed from webpack

Simounet:
We don't need these files from `_global` to be directly accessible. They are imported into themes files and never used by the browser.


j0k3r:
I'm wondering if it's not used when user share an entry?

j0k3r:
https://github.com/wallabag/wallabag/blob/master/src/Wallabag/CoreBundle/Resources/views/themes/common/Entry/share.html.twig#L20-L24

Simounet:
You got me! I didn't think about this part. I will change that.

0098912200770 (approved):

0098912200770:
My wallet account send tanks

در تاریخ پنجشنبه ۲۶ نوامبر ۲۰۲۰،‏ ۰۸:۲۹ Jérémy Benoist <
notifications@github.com> نوشت:

j0k3r (approved):

PR #4844, opened 2020-12-14
Add a command to clean downloaded images

### LINKED ISSUE #4827: Delete unused images
### Issue details
My wallabag instance is now a few years old. The complete wallabag directory has a size of 2 GB (I already cleared var/logs/prod.log, which had over 250 MB).
The directory web/assets/images uses 1,7 GB. There are 18630 files in it. As today I have 225 articles (incl. archived) in wallabag. I often delete saved articles, so I think there are still images from deleted articles there.
How can I do house keeping for the images directory?

### Environment

* wallabag version (or git revision) that exhibits the issue: 2.4.0
* How did you install wallabag? Via `git clone` or by downloading the package? Think I used package, but "git status" prints "HEAD detached at 2.4.0". Upgrade from 2.3 to 2.4 done with "make update"
* php version: 7.4.3
* OS: Ubuntu 20.04.1 LTS

j0k3r:
| Q             | A
| ------------- | ---
| Bug fix?      | kind of
| New feature?  | kind of
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| License       | MIT

Fixes #4827

There were a bug in versions prior to 2.4.0 where images weren't properly removed (mostly when coming from the API).
With that command, we'll be able to remove images which aren't associated to any entries.

Like other command you can pass a username to only clean one user.

### DIFF
--- a/src/Wallabag/CoreBundle/Command/CleanDownloadedImagesCommand.php
+++ b/src/Wallabag/CoreBundle/Command/CleanDownloadedImagesCommand.php
@@ -2,10 +2,9 @@
 
 namespace Wallabag\CoreBundle\Command;
 
-use Doctrine\ORM\NoResultException;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
-use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Finder\Finder;
@@ -23,10 +22,11 @@ class CleanDownloadedImagesCommand extends ContainerAwareCommand
         $this
             ->setName('wallabag:clean-downloaded-images')
             ->setDescription('Cleans downloaded images which are no more associated to an entry')
-            ->addArgument(
-                'username',
-                InputArgument::OPTIONAL,
-                'User to clean'
+            ->addOption(
+               'dry-run',
+               null,
+               InputOption::VALUE_NONE,
+               'Do not remove images, just dump counters'
             );
     }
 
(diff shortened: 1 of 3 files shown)


src/Wallabag/CoreBundle/Command/CleanDownloadedImagesCommand.php
```diff
@@ -0,0 +1,123 @@
...
+            ->in($baseFolder);
+
+        $existingPaths = [];
+        foreach ($finder as $file) {
+            $existingPaths[] = $file->getFilename();
+        }
+
+        // check if existing path are valid, if not, remove all images and the folder
+        foreach ($existingPaths as $existingPath) {
+            if (!\in_array($existingPath, $validPaths, true)) {
```
Kdecherf, line 107: In my current understanding of this file, it will not only remove unused files but also all files owned by other users

Kdecherf (changes requested):
It could be interesting to have a `dry-run` option to know what would do the command before actually deleting files.

src/Wallabag/CoreBundle/Command/CleanDownloadedImagesCommand.php
j0k3r: Oh right, hard to predict then.
Or I just remove the option to clean only one given user?

Kdecherf: Yeah we may just remove the user selector. I don't see value in cleaning assets for a single user here.

Kdecherf (changes requested):
Just saw your changes.
The dry-run mode looks good.

However the `clean()` logic still removes files of other users: we should build the list of files outside of `clean()` as we need to merge file lists for all users before checking against existing files.

j0k3r:
image

Kdecherf (approved):

PR #4869, opened 2020-12-19
Fix extra padding on the homepage main content

### LINKED ISSUE #4862: Display on entries view has changed
Capture d’écran_2020-12-18_14-41-44

A picture is better than words. You can see that the entries list is not stuck any more to the left sidebar. The margin is more important than before. 

@Simounet any idea?

Simounet:
Back to normal spacing for the homepage main content.

Before:
wallabag-before

After:
wallabag-after

Fixes #4862.


nicosomb (approved):

Kdecherf (approved):
Looks good

PR #4908, opened 2021-01-05
Fix nav lang filter

Simounet:
I think this was a wrong JS calculation because of this `min-width`.


j0k3r:
Screenshot before/after?

Simounet:
Before:
wallabag-before

After:
wallabag-after

j0k3r (approved):

PR #4983, opened 2021-01-27
Dark theme updated

Simounet:
Hi there,
I'm trying to keep this dark theme variation "simple". So I switched the colors to hexa values instead of rgba (we don't need transparency here). I removed collection-item custom style. I don't think this is a good idea to make layout changes here, both light and dark mode should share the same layout.


j0k3r (commented):
Can you share some screenshots before / after?

Simounet:
Home before
wallabag-home-before

Home after
wallabag-home-after

Entry before
wallabag-entry-before

Entry after
wallabag-entry-after

j0k3r (approved):

nicosomb (approved):

Kdecherf (approved):

PR #5249, opened 2021-04-29
Update validators.cs.yml

marapavelka:
Updated Czech translation


j0k3r:
Thanks for that but you should use Weblate instead, which is the system we are using to centralize all translations: https://hosted.weblate.org/projects/wallabag/validators/cs/

PR #5378, opened 2021-08-03
Show when articles is read

### LINKED ISSUE #869: View if articles are read/favorited when searching
When doing a search, read articles and favorited articles should display differently than regular articles.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes #869

The feature was already implemented, but only in tag view. Now, we have it for all the routes where it's needed. 

Capture d’écran_2021-08-03_11-10-41
Capture d’écran_2021-08-03_11-10-56

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
@@ -20,6 +20,7 @@
 
 {% block content %}
     {% set listMode = app.user.config.listMode %}
+    {% set entriesWithArchivedClassRoutes = ['tag_entries', 'search', 'all'] %}
     {% set currentRoute = app.request.attributes.get('_route') %}
     {% if currentRoute == 'homepage' %}
         {% set currentRoute = 'unread' %}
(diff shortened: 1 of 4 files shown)


src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entries.html.twig
```diff
@@ -20,6 +20,7 @@
 
 {% block content %}
     {% set listMode = app.user.config.listMode %}
+    {% set displayArchivedEntriesRoutes = ['tag_entries', 'search', 'all'] %}
```
Kdecherf, line 23: The variable should have a better name like `entriesWithArchivedClassRoutes` or something similar.
Also, I'm not a huge fan of putting `search` here, we should limit this to searches on all route (`currentRoute=all`)

nicosomb: I changed the name. 

About `search`, I added it because I think it's better to have consistency.

j0k3r (commented):

j0k3r (approved):

Kdecherf (approved):

PR #5379, opened 2021-08-03
Fixed search on homepage

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

When we are on `https://mywallabag/`, we are on the `homepage` current page, which is the unread view. This page needs to react as the unread page (`https://mywallabag/unread/list`).

This issue fixes a bug when we research articles. The search was done on all articles: it needs to be done on unread articles only.


Kdecherf (approved):
I'm fine with this change.
However I think we should rework the goal of the homepage in a future version (_e.g. giving the ability to the user to set the default display for the homepage_)

PR #5380, opened 2021-08-03
Added random button on article view

### LINKED ISSUE #5112: Random article button missing from sidebar when reading article
### Environment

* **Version**: 2.4.1
* **Installation**:
* **PHP version**:
* **OS**:
* **Database**:
* **Parameters**:

Read article.

On the sidebar, they should be a "random article" button, making it easy to continue reading random articles when you're finished reading the first one. Coming back to the "main" page is possible, but it kind of breaks the focus you have.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Should fix #5112. 

The random research will be done on all the articles. Because we don't know the current context: if we are on "starred" articles, we click on the random buttom, we have a random article inside the favorite ones. But, when we are on the article view, we don't know where we were before.

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
@@ -216,6 +216,13 @@
         </li>
         {% endif %}
 
+        <li class="bold border-top border-bottom">
+            <a class="waves-effect collapsible-header" title="{{ 'menu.top.random_entry'|trans }}" href="{{ path('random_entry', { 'type': 'all' }) }}">
+                <i class="material-icons small">casino</i>
+                <span>{{ 'menu.top.random_entry'|trans }}</span>
+            </a>
+        </li>
+
         <li class="bold">
             <a class="waves-effect collapsible-header">
                 <i class="material-icons small">file_download</i>


src/Wallabag/CoreBundle/Resources/views/themes/material/Entry/entry.html.twig
```diff
@@ -216,6 +216,14 @@
         </li>
         {% endif %}
 
+        <li class="bold border-top border-bottom hide-on-med-and-down">
+            <a class="waves-effect collapsible-header" title="{{ 'menu.top.random_entry'|trans }}" href="{{ path('random_entry', { 'type': 'all' }) }}">
+                <i class="material-icons small">casino</i>
+                <span>{{ 'menu.top.random_entry'|trans }}</span>
+            </a>
+            <div class="collapsible-body"></div>
```
tcitworld, line 224: Note : All other items have this `<div>` but I'm not even sure it's needed in this case.

tcitworld (approved):

Kdecherf (commented):
Like you said in the PR, this button does not take into account if we're looking at starred or unread or archived articles. Thus I'm unsure if this change is valuable.

Also, please note that it does not appear on mobile view as you used `hide-on-med-and-down` class

j0k3r (commented):

j0k3r (approved):

Kdecherf (approved):

Issue #4987, opened 2021-01-28
Update breaking on PHP 8


opsydev:
### Environment

* **Version**: 2.4.1 or master
* **Installation**: Git Clone on a VPS
* **PHP version**: PHP 8.0
* **OS**: Ubuntu 20.04
* **Database**: Postgresql
* **Parameters**:

<details>
  <summary>My <code>app/config/parameters.yml</code> is:</summary>

  ```
  parameters:
    database_driver: pdo_pgsql
    database_host: 127.0.0.1
    database_port: null
    database_name: xxxxxxxx
    database_user: xxxxxxxx
    database_password: xxxxxxxx
    database_path: null
    database_table_prefix: wallabag_
    database_socket: null
    database_charset: utf8
    domain_name: 'https://xxxxxxxx'
    mailer_transport: smtp
    mailer_host: xxxxxxxx
    mailer_user: xxxxxxxx
    mailer_password: xxxxxxxx
    locale: en
    secret: xxxxxxxx
    twofactor_auth: true
    twofactor_sender: xxxxxxxx
    fosuser_registration: true
    fosuser_confirmation: true
    from_email: xxxxxxxx
    rss_limit: 50
    rabbitmq_host: localhost
    rabbitmq_port: 5672
    rabbitmq_user: xxxxxxxx
    rabbitmq_password: xxxxxxxx
    rabbitmq_prefetch_count: 10
    redis_scheme: tcp
    redis_host: localhost
    redis_port: 6379
    redis_path: null
    redis_password: null
    mailer_port: false
    mailer_encryption: null
    mailer_auth_mode: null
    fos_oauth_server_access_token_lifetime: 3600
    fos_oauth_server_refresh_token_lifetime: 1209600
    sentry_dsn: null
    server_name: 'xxxxxxxx xxxxxxxx'
  ```
</details>

### What steps will reproduce the bug?
Upgrade Ubuntu to use system PHP 8.0. Run the wallabag `make update`

It breaks because of Doctrine issues, the following error shows up:

```
  [RuntimeException]
  An error occurred when executing the "'cache:clear --no-warmup'" command:



  PHP Fatal error:  Declaration of Doctrine\DBAL\Driver\PDOConnection::query() must be compatible with PDO::query(string $query, ?int $fetchMode = null, mixed ...$fetchModeArgs) in /path/to/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOConnection.php on line 81

[...]

PR #5591, opened 2022-01-31
Add build test on PHP 8.0 & 8.1

j0k3r:
Add `isTransactional` to `WallabagMigration` because PHP 8 behave differently with PDO transaction.
This is a workaround because we can't upgrade Doctrine Migration for now (upper versions have the fix, see https://github.com/doctrine/DoctrineMigrationsBundle/issues/393).

Fix https://github.com/wallabag/wallabag/issues/4987

- Build is now using Composer v2 (instead of v1)
- All actions have been updated to latest version
- Fix bug in PHP 8 were `$entry->getTags()` can't be properly used as a _traversable_ by `assertContains` during tests. Added a custom method `Entry::getTagsLabel()` which return a flatted tag array with only label
- Replace `assertNotRegExp` by `assertDoesNotMatchRegularExpression` because it was deprecated

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -20,15 +20,15 @@ jobs:
         uses: "shivammathur/setup-php@v2"
         with:
           coverage: "none"
-          php-version: "7.3"
-          tools: cs2pr, pecl, composer:v1
+          php-version: "7.4"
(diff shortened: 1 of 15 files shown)


Kdecherf (approved):
LGTM

Issue #5596, opened 2022-02-03
generated epub with wrong metadata


Monirzadeh:
### Environment

* **Version**: 2.4.2
* **Installation**: docker image
* **PHP version**:
* **OS**: Debian
* **Database**: MariaDB

</details>

### What steps will reproduce the bug?
create a epub version of some persian webpage [example](https://www.1pezeshk.com/archives/2022/01/the-machinist-2004.html) than check epub metadata.
if you see source code of saved article in wallabag webapp you see `lang="fa-IR"` that mean wallabag detect language correctly but when it make epub set language metadata uncorrected to english.

### why is it important?
some app reader automatically read metadata and base on that implement some style for specific language.

j0k3r:
Thanks for the report @Monirzadeh!
We defined the language from the entry without checking it. `fa-IR` isn't accepted by epub, we must use `fa` instead. I've fixed it in a PR #5647

Monirzadeh:
i try to test that (have some wired problem with setup development environment) if i can successfully build that report any problem here.
about [this](https://github.com/wallabag/wallabag/pull/5647#issuecomment-1057365878). my first target for this feature is [koreader](https://github.com/koreader/koreader). i try to test that on this for now.
thanks for your followup
P.S. is it possible for you to add more details to how setup development environment in wallabag wiki it can be helpful.

Monirzadeh:
hi @j0k3r sorry for late response i check this commit and it better than before but some wired behavior observe.
i test that with multiple link but some problem still exist.
we get different tag for same language.
for example
[this article](https://www.1pezeshk.com/archives/2022/02/why-your-brain-is-a-time-machine-living-15-seconds-in-the-past.html)  languge tag is `fa-IR`
but [this one](https://fa.wikipedia.org/wiki/%D9%BE%D8%B3%D8%AA%DA%AF%D8%B1%D8%B3%E2%80%8C%DA%A9%DB%8C%D9%88%D8%A7%D9%84) languge tag is `fa`

another that maybe not related. [this](https://blog.alihoseiny.ir/%D8%A8%D8%B1%D8%A7%DB%8C-%D8%A7%DB%8C%D9%86%D8%AA%D8%B1%D9%86%D8%AA%D8%8C-%D8%A8%D8%B1%D8%A7%DB%8C-%D8%A2%D8%B2%D8%A7%D8%AF%DB%8C/) page language tag detect as `en` instead of `fa-IR`

PR #5647, opened 2022-03-02
Fix epub language

j0k3r:
Ensure language has `_` instead of `-`
And update wallabag/phpepub

Fix https://github.com/wallabag/wallabag/issues/5596

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/EntriesExport.php
+++ b/src/Wallabag/CoreBundle/Helper/EntriesExport.php
@@ -161,8 +161,8 @@ class EntriesExport
          */
 
         $book->setTitle($this->title);
-        // Not needed, but included for the example, Language is mandatory, but EPub defaults to "en". Use RFC3066 Language codes, such as "en", "da", "fr" etc.
-        $book->setLanguage($this->language);
+        // EPub specification requires BCP47-compliant languages, thus we replace _ with -
+        $book->setLanguage(str_replace('_', '-', $this->language));
         $book->setDescription('Some articles saved on my wallabag');
 
         $book->setAuthor($this->author, $this->author);
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,7 @@
         "tecnickcom/tcpdf": "^6.3.0",
         "twig/extensions": "^1.5",
         "wallabag/php-mobi": "~1.0",
-        "wallabag/phpepub": "^4.0.7.2",
+        "wallabag/phpepub": "^4.0.10",
         "willdurand/hateoas-bundle": "~2.1"
     },
     "require-dev": {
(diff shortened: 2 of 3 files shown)


tcitworld:
> As the comment in the doc says, EPub defined language MUST be on 2 chars.

Are you sure about this? That's not what RFC3066 says. From what I can see ePub references BCP 47 language tags anyway. :thinking:

j0k3r:
At least I've tested locally with Calibre and the generated epub with 2 chars language is properly detected while the one defined with 5 chars language is displayed as English (the default value then).

Kdecherf (commented):
EPUB 3.2 Specification states that the language value must be compliant with BCP 47[^1].
BCP 47 redirects to RFC 5646[^2] which specifies a syntax that technically accepts values like "fa-IR".

This sounds more like a bug in the PHPePub library, see https://github.com/wallabag/PHPePub/blob/master/src/PHPePub/Core/EPub.php#L1460

I think we should close this PR and discuss about it on the library side instead.

Edit:
[^1]: https://www.rfc-editor.org/info/bcp47
[^2]: https://www.rfc-editor.org/rfc/rfc5646.txt

j0k3r:
Awesome review @Kdecherf, I'll debug using the lib.

src/Wallabag/CoreBundle/Helper/EntriesExport.php
```diff
@@ -162,7 +162,7 @@ private function produceEpub()
 
         $book->setTitle($this->title);
         // Not needed, but included for the example, Language is mandatory, but EPub defaults to "en". Use RFC3066 Language codes, such as "en", "da", "fr" etc.
-        $book->setLanguage($this->language);
+        $book->setLanguage(substr($this->language, 0, 2));
```
Kdecherf, line 165: Even though https://github.com/wallabag/PHPePub/pull/3#discussion_r825345271 will fix an issue on PHPePub side, we still need to fix something in wallabag, here is a suggestion @j0k3r as we store language with `_`:

```suggestion
        // EPub specification requires BCP47-compliant languages, thus we replace _ with -
        $book->setLanguage(str_replace('_', '-', $this->language));
```

src/Wallabag/CoreBundle/Helper/EntriesExport.php
```diff
@@ -162,7 +162,8 @@ private function produceEpub()
 
         $book->setTitle($this->title);
         // Not needed, but included for the example, Language is mandatory, but EPub defaults to "en". Use RFC3066 Language codes, such as "en", "da", "fr" etc.
```
Kdecherf, line 164: Could you remove the first comment?

Kdecherf (commented):
Note on a comment to remove, could you also squash your commits?

Kdecherf (dismissed):

Kdecherf (approved):

PR #5664, opened 2022-03-13
Feat/home entries updated

Simounet:
Hi,

I updated the style for the home entries. An overflow issue was bothering me for a long time.
Before:
overflow before

After:
overflow after

Because of that, I needed to extract the mass action list from the list (it wasn't semantical anyway).

mass-action-extracted

And because I was there, I replaced the entries' unordered list by an ordered one.

Have a nice day.


j0k3r:
Looks like updating the HTML structure broken tests

Simounet:
Ok, I get it and I will fix it.

Simounet:
Fixed it. I'm proposing to use the `data-test` attribute for testing purpose to separate concerns.

j0k3r:
Great! Good idea.
Can you squash?

Simounet:
I can but each of these commits mean something.

j0k3r (approved):
I'm ok with that!

nicosomb (approved):

PR #5673, opened 2022-03-15
Add new endpoint for API: config

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | yes/no
| License       | MIT

Added a new API endpoint, to get user configuration (ping @bourvill).  

Maybe one day, we can improve this endpoint (to update configuration, etc). 

~~One issue with this new feature: I don't want to get the user entries in the json response. I need to fix this.~~

### DIFF
--- a/src/Wallabag/CoreBundle/DataFixtures/ConfigFixtures.php
+++ b/src/Wallabag/CoreBundle/DataFixtures/ConfigFixtures.php
@@ -24,7 +24,6 @@ class ConfigFixtures extends Fixture implements DependentFixtureInterface
         $adminConfig->setPocketConsumerKey('xxxxx');
         $adminConfig->setActionMarkAsRead(0);
         $adminConfig->setListMode(0);
-        $adminConfig->setListMode(0);
 
         $manager->persist($adminConfig);
 
--- a/src/Wallabag/CoreBundle/DataFixtures/ConfigFixtures.php
+++ b/src/Wallabag/CoreBundle/DataFixtures/ConfigFixtures.php
@@ -24,6 +24,7 @@ class ConfigFixtures extends Fixture implements DependentFixtureInterface
         $adminConfig->setPocketConsumerKey('xxxxx');
         $adminConfig->setActionMarkAsRead(0);
         $adminConfig->setListMode(0);
+        $adminConfig->setListMode(0);
 
         $manager->persist($adminConfig);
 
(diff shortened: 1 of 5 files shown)


src/Wallabag/CoreBundle/DataFixtures/ConfigFixtures.php
```diff
@@ -24,6 +24,7 @@ public function load(ObjectManager $manager): void
         $adminConfig->setPocketConsumerKey('xxxxx');
         $adminConfig->setActionMarkAsRead(0);
         $adminConfig->setListMode(0);
+        $adminConfig->setListMode(0);
```
j0k3r, line 27: To revert

src/Wallabag/CoreBundle/Entity/Config.php
```diff
@@ -77,34 +88,44 @@ class Config
...
+     * @Groups({"config_api"})
      */
     private $readingSpeed;
 
     /**
      * @var string
      *
      * @ORM\Column(name="pocket_consumer_key", type="string", nullable=true)
+     *
+     * @Groups({"config_api"})
```
j0k3r, line 110: I don't think we need to expose it, it won't be used on client side only on server side

src/Wallabag/CoreBundle/Entity/Config.php
```diff
@@ -37,6 +40,8 @@ class Config
      *
      * @Assert\NotBlank()
      * @ORM\Column(name="theme", type="string", nullable=false)
+     *
+     * @Groups({"config_api"})
```
j0k3r, line 44: Is it really useful? As there will be only one them starting from 2.6.0

tests/Wallabag/ApiBundle/Controller/ConfigRestControllerTest.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $this->assertArrayHasKey('id', $content);
+        $this->assertArrayHasKey('items_per_page', $content);
+        $this->assertArrayHasKey('language', $content);
+        $this->assertArrayHasKey('reading_speed', $content);
+        $this->assertArrayHasKey('action_mark_as_read', $content);
+        $this->assertArrayHasKey('list_mode', $content);
+
+        $this->assertSame(200.0, $content['reading_speed']);
+        $this->assertSame('en', $content['language']);
```
Kdecherf, line 24: I think it could be useful to check that there is no other properties in the resulting array. In this case we could check that it only contains 6 keys.

nicosomb:
Thank you @j0k3r. 

@Kdecherf could you review this PR please?

Kdecherf (approved):
LGTM

PR #5676, opened 2022-03-15
Fix row and card style regressions

Simounet:
Fix regression on common `.card` and `.row` class use. Visible on the login and settings pages.


j0k3r (approved):

nicosomb:
Thank you!

PR #5708, opened 2022-04-01
Composer 2.3 isn't compatible with wallabag

j0k3r:
SensioDistributionBundle isn't compatible anymore with Composer 2.3 which break the build when bootstrapping the application.
To install wallabag you must use a version lower than Composer 2.3.


Kdecherf (approved):
Looks good

Issue #3460, opened 2017-12-05
Baggy theme deprecation


Simounet:
Maybe it is time to mark our long time loved baggy buddy as deprecated and invite users to switch to material. I don't know if you have any information about this theme being still used.
What do you think?

nicosomb:
FYI, on app.wallabag.it, 123 users (on 3172 users) have the baggy theme.

Simounet:
It's a great info. It could be great to be in touch with them (mailing with a poll) to know why they are still using it instead of material.

yelfathi:
The default of Material theme is the preview picture, I prefer to have a minimalist list of articles instead. So maybe add an option to hide this pictures.

tcitworld:
There's a list view in material theme

yelfathi:
I don't see the option at wallabag.it.

tcitworld:
It's the button next to the RSS feed button.
![](https://screenshotscdn.firefoxusercontent.com/images/46b1183f-696f-45c2-96f3-b6d553bed7fb.png)

yelfathi:
Ah this button yes I have it but on Material theme you still see the preview pictures (indeed in lower format) but it's not minimalist as I said earlier. Thanks.

j0k3r:
I'm all in for that.

PR #5742, opened 2022-04-20
Deprecated Baggy

j0k3r:
- a big message will be displayed to user using the Baggy theme

image

- switching from Material to Baggy is no more allowed in config (it'll be forced to material), here is the message when you selected the Baggy theme:
image

- the theme label in the config for Baggy is now _Baggy (DEPRECATED)_

image

See https://github.com/wallabag/wallabag/issues/3460

Do we need to bother translating the warning message about the forced theme? That message will be removed in 2.6.0 anyway.


nicosomb (approved):
:tada:

PR #5775, opened 2022-05-03
Deprecate mobi export

### LINKED ISSUE #5773: Deprecate/remove MOBI export
It looks like Amazon is planning[1] to support conversion of EPUB files for their Kindle devices, which is a quite good news.

They also plan to phase out their support of the legacy MOBI file format; that we support in wallabag mostly for Kindle devices afaik.

@wallabag/core I think we could simplify our dependencies and maintenance by planning the deprecation and removal of this export format. What do you think?

[1]: https://gizmodo.com/amazon-kindle-e-readers-will-now-convert-epub-to-kindle-1848867278

Kdecherf:
- Disable mobi export on new installations
- Put a note in the setting label

@wallabag/core do you see other things I should put in this deprecation part?

Closes #5773

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | ~
| Documentation | no
| Translation   | ~
| CHANGELOG.md  | no
| License       | MIT


app/Resources/CraueConfigBundle/translations/CraueConfigBundle.en.yml
```diff
@@ -3,7 +3,7 @@ download_pictures: Download pictures onto your server
 carrot: Enable share to Carrot
 diaspora_url: diaspora* URL, if the service is enabled
 export_epub: Enable ePub export
-export_mobi: Enable .mobi export
+export_mobi: Enable .mobi export (deprecated, will be removed in 2.6.0)
```
j0k3r, line 6: Can't this be directly in the template to avoid having all traductions to be up to date?
I put the deprecated message for Baggy directly in the template, without translation.

Kdecherf: It would be only in english then?

j0k3r: Yes I think it's ok because the message will be removed anyway in 2.6.0

Kdecherf: Fine

j0k3r:
image

j0k3r (approved):

stephen-hill:
Hi. I'm curious why you have decided to start removing support for the mobi format?

Without it, I will not be able to add items to my Kindle.

PR #5822, opened 2022-05-23
Jump to Node 16 to build assets

j0k3r:
We were used the old node 12 and some deps are dropping that version.


Kdecherf (approved):
Looks good

PR #5861, opened 2022-06-14
Add tag deletion from tags list

### LINKED ISSUE #2952: Remove tag from tag/list
Hi,
It would be great to be able to remove tags from the tag/list view. I've got 87 tags generated automatically and I don't want to go to each article to remove it. Let me know if I missed something.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | yes/no
| License       | MIT

Fixed #2952

Need help for design (@Simounet maybe?)


Kdecherf (changes requested):
Could you add a screenshot and a prompt for asking the user to confirm the deletion?

Simounet:
Here is the screen. I used a JS `confirm()` but this is not a good practice so if I'm motivated enough, I'll do another PR to replace the `confirm()`'s by a better handler like [a11ydialog](https://a11y-dialog.netlify.app/) but I don
tag-delete-action.

nicosomb:
Thank you @Simounet

yguedidi:
@nicosomb according to the milestone set, shouldn't this PR target the `2.6.0` branch instead of `master`? or change the milestone of course! :)

tests/Wallabag/CoreBundle/Controller/TagControllerTest.php
```diff
@@ -141,6 +141,53 @@ public function testRemoveTagFromEntry()
...
+        $this->getEntityManager()->persist($entry);
+
+        $entry2 = new Entry($this->getLoggedInUser());
+        $entry2->setUrl('http://0.0.0.0/bar');
+        $entry2->addTag($tag);
+        $this->getEntityManager()->persist($entry2);
+        $this->getEntityManager()->flush();
+        $this->getEntityManager()->clear();
+
+        $client->request('GET', '/tag/delete/' . $tag->getSlug());
```
yguedidi, line 164: @nicosomb wouldn't it be better to use a POST request for an action that alters the database?

nicosomb: You're right ... for the API side. 

But here, it's not the API part. Our tests are web test cases. We just simulate a click on the link.

yguedidi: I think even in interfaces it should be a POST action :)
using a form and a submit button.
IMO it's a good practice the have all actions that alter the database to be POST actions

j0k3r: Why don't you use the client and then the crawler to retrieve the button and simulate a click on it?

nicosomb: @j0k3r done

Kdecherf (approved):

PR #5925, opened 2022-08-09
Use localhost alias instead of hard coded IP

goodnewz:
On some systems, localhost could be in other private IP ranges not just 127.0.0.1 (e.g., 10.X.X.X or 172.X.X.X). This suggestion minimizes errors in setting up Wallabag by new users who might have MYSQL/MariaDB configured on other private localhost IP ranges or on multiple bind IPs. By using 'localhost' the user doesn't need to know where MYSQL/MariaDB was set up which prevents database connection errors when installing the application because of incorrectly specified server IPs.

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -11,7 +11,7 @@ parameters:
     # database_password: %env.database_password%
 
     database_driver: pdo_mysql
-    database_host: 127.0.0.1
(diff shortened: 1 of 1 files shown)


j0k3r:
As it's defined in the `.dist` you are free to use a different value when copying to the final destination `app/config/parameters.yml`.
We don't need to update the default value.

Issue #5328, opened 2021-06-28
DownloadImage Locally with webp,svg,icon


iSwiftLiu:
Hello, is there any plan to support webp, svg these types of pictures stored locally? I tried to modify it according to the code, but because I don’t know PHP, I managed to store the static webp file locally through the imagewebp function, hoping to support svg and webp

Kdecherf:
Hello,

Do you have example links with webp and svg files?

iSwiftLiu:
Hello,
svg images:
https://fuckcloudnative.io/posts/use-wireguard-as-kubernetes-cni/
https://fuckcloudnative.io/posts/egress-tcp/

webp images:
https://www.upyun.com/webp?utm_source=zhihu&utm_medium=referral&utm_campaign=26474171&utm_term=webp
static:
https://p.upyun.com/demo/webp/webp/jpg-0.webp
https://p.upyun.com/demo/webp/webp/png-0.webp
https://p.upyun.com/demo/webp/webp/gif-0.webp
dynamic:
https://p.upyun.com/demo/webp/webp/animated-gif-0.webp

my code about static webp:
```
            case 'webp':
                imagealphablending($im, true);
                imagesavealpha($im, true);
                imagewebp($im, $localPath, 100);
                $this->logger->debug('DownloadImages: Re-creating webp');
                break;
```
500 internal error appears when storing dynamic webp.

iSwiftLiu:
It seems that although this error will be reported, the dynamic webp image saved separately has been saved.
```
500: Internal Server Error
Error: gd-webp cannot allocate temporary buffer
```

dxarwc3:
I had the same issue which also inhibited the mobile app from syncing items. My hack (adapted from a web search result) into Docker image 2.5.1:

#### wallabag/src/Wallabag/CoreBundle/Helper/DownloadImages.php:~150
``` php
// WEBP animations cause E500 with gd-webp and mobile app upload failure.
// Exit here and do not try imagecreate when WEBP animation detected.
if ( $ext == 'webp' && (
   stripos($res->getBody(), 'WEBPVP8X') !== false ||
   stripos($res->getBody(), 'ANIM') !== false )
) {
   return false;
}
```
####
 (sorry, not enough experience for a PR)

j0k3r:
Can you share an url so we can reproduce the bug and fix it?

dxarwc3:
[...]

PR #5992, opened 2022-09-27
Add support to download SVG locally

j0k3r:
We can't process a SVG using PHP (with GD) so instead we sanitize the SVG (to ensure it's a real SVG and not a bad input file) and then just copy the response from the server locally.

Using that method my only concern is about security. Is https://github.com/darylldoyle/svg-sanitizer safe enough to trust it?
What kind of other security check we can add?

Related https://github.com/wallabag/wallabag/issues/5328

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/DownloadImages.php
+++ b/src/Wallabag/CoreBundle/Helper/DownloadImages.php
@@ -2,6 +2,7 @@
 
 namespace Wallabag\CoreBundle\Helper;
 
+use enshrined\svgSanitize\Sanitizer;
 use GuzzleHttp\Psr7\Uri;
 use GuzzleHttp\Psr7\UriResolver;
 use Http\Client\Common\HttpMethodsClient;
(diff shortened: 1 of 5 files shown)


tcitworld:
`svg-sanitizer` excludes `<script>` and `<foreignObject>` and entity loader is disabled so the basics are here.

tcitworld (dismissed):

src/Wallabag/CoreBundle/Helper/DownloadImages.php
```diff
@@ -146,6 +147,26 @@ public function processSingleImage($entryId, $imagePath, $url, $relativePath = n
...
+            $sanitizer->minify(true);
+            $sanitizer->removeRemoteReferences(true);
+            $cleanSVG = $sanitizer->sanitize((string) $res->getBody());
+
+            // add an extra validation by checking about `<svg `
+            if (false === $cleanSVG || false === strpos($cleanSVG, '<svg ')) {
+                $this->logger->error('DownloadImages: Bad SVG given', ['path' => $imagePath]);
+
+                return false;
+            }
```
Kdecherf, line 164: Shouldn't we put a try-catch here, like for lines 171-177?

j0k3r: What other alternative we have, like in the catch from lines 171-177?
Or maybe just a try/catch with a `return false` in the catch?

Kdecherf: Yes, we could put `return false` in the catch

Kdecherf (approved):

PR #6085, opened 2022-11-14
Remove some deprecation

j0k3r:
Just some stuff deprecated and replaced by other Interface.
Might ease the Symfony upgrade (poke @yguedidi)

Also remove router from `src/Wallabag/CoreBundle/Helper/PreparePagerForEntries.php` because it's not used


Kdecherf (approved):

yguedidi:
Yes I had such a commit in a local branch making services private. I'll open it soon

PR #6099, opened 2022-11-23
Upgrade to Symfony 4.4 

j0k3r:
- disable autowiring for Event (because the Entry entity was injected)
- rename `getClient()` for test to `getTestClient()` to avoid error while overriding (from `BrowserKitAssertionsTrait`)
- accessing the logger from a controller is a bit more complex now, so I just commented the lines (it requires to use `AbstractController` and service injecting into the controller, etc.)


tests/Wallabag/CoreBundle/Command/InstallCommandTest.php
```diff
@@ -201,6 +202,16 @@ public function testRunInstallCommandChooseResetSchema()
 
     public function testRunInstallCommandChooseNothing()
     {
+        /*
+         *  [PHPUnit\Framework\Error\Warning (2)]
+         *  filemtime(): stat failed for /home/runner/work/wallabag/wallabag/var/cache/tes_/ContainerNVNxA24/appAppKernelTestDebugContainer.php
+         *
+         * I don't know from where the "/tes_/" come from, it should be "/test/" instead ...
```
nicosomb, line 209: I often have this kind of folder with `de_` instead of `dev`.

j0k3r: On which Symfony version?

yguedidi:
@j0k3r sorry for my absence since a moment 😓 
I'll push my preparation PR making services private today

j0k3r:
@yguedidi Don't worry you already did a great job and that PR was much more easier to create than before your work 👍🏼

tests/Wallabag/CoreBundle/Command/InstallCommandTest.php
nicosomb: on the current one. For example, today on wallabag.it (it's with a `~` but same issue I think) 

Capture d’écran 2022-11-24 à 11 47 59

Kdecherf:
> accessing the logger from a controller is a bit more complex now, so I just commented the lines (it requires to use AbstractController and service injecting into the controller, etc.)

@j0k3r should we wait a fix before reviewing and merging this PR?

j0k3r:
> @j0k3r should we wait for a fix before reviewing and merging this PR?

Nop, it'll be fixed when controllers will be defined as a service, injecting the logger will be much much easier.

nicosomb (approved):

PR #6119, opened 2022-12-04
Fix: Use Session instead of Referrer for Redirection

Spoons:
### Introduction

Thank you for taking the time to create and maintain Wallabag! I appreciate your efforts.

### Problem

Wallabag uses the `referrer` HTTP header field to return the browser to the originating location after an action, such as star, archive, etc., is completed. As a push towards better privacy, in [Firefox 87](https://blog.mozilla.org/security/2021/03/22/firefox-87-trims-http-referrers-by-default-to-protect-user-privacy/), Mozilla trims the referrer field when crossing domains. Furthermore, they have recently introduced the [network.http.referer.spoofSource](https://wiki.mozilla.org/Security/Referrer) flag. This flag causes Firefox always to send the target URL as the referrer. With this behavior, Wallabag no longer functions properly. Performing any action causes a recursive 302 redirect loop. 

### Solution

This pull request stores the previous URL in the session and retrieves it during routing. While this issue likely does not affect many users, as privacy concerns grow, spoofing and disabling referrers are becoming increasingly common. Almost all privacy-conscious Firefox users will have this flag enabled. Many user prefs.js setups recommend it.

Furthermore, multiple plugins for Firefox and Chrome replicate this behavior. There are only a few sites known to break with spoofed referrers; unfortunately, Wallabag is one of them! 

Please let me know what I can do for you to merge this. Doing so will allow me to use Wallabag in a way that is convenient and useful for me.

This issue is mentioned [here](https://github.com/pyllyukko/user.js/issues/94#issuecomment-600672820).

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | yes
| License       | MIT


nicosomb:
Hello @Spoons. Thank you for your first pull request :-) 
It seems that the testsuite is broken with your changes. Could you please have a look on it?

Spoons:
Absolutely. I'll take a look at it this weekend.

Spoons:
The issue was the test was testing for absolute URL's and my change makes the application generate 'root relative' URLs. My understanding is that relative URLs are preferable for navigating within the same domain / application. So, I updated the test to compare against root relative URLs.

However, if you would prefer that we continue to generate absolute URLs, just let me know and I'll make it do that.

src/Wallabag/CoreBundle/Controller/TagController.php
```diff
@@ -253,6 +253,6 @@ public function removeTagAction(Tag $tag, Request $request)
 
         $redirectUrl = $this->get(Redirect::class)->to($request->headers->get('referer'), '', true);
```
j0k3r, line 255:
```suggestion
```

Kdecherf (commented):
Didn't check how it runs yet but what is the expected behavior if you open different wallabag pages in several tabs regarding the session value?

Spoons:
> Didn't check how it runs yet but what is the expected behavior if you open different wallabag pages in several tabs regarding the session value?

My understanding is that session storage is unique to each tab. They survive page refreshes and navigation but should stay unique per tab. 

[sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage)

I didn't notice any incorrect behavior with tabs when I was testing it. I did test with multiple tabs. However, I certainly wasn't looking for it. I take another look to ensure it's working properly.

Spoons:
> @Spoons your pull request has conflicts. Could you please fix them?

Sure. I'll take a look.

nicosomb:
@Spoons any news about your PR?

Spoons:
Merged with main and updated. Passing tests on my end.

nicosomb:
I tested your PR and it's ok for me 🚀

PR #6136, opened 2022-12-13
Update to FOSUserBundle 3.1

j0k3r:
Also remove some deprecation from Symfony.
Use `LegacyEventDispatcherProxy` to handle Symfony 4 dispatch from FOSUser

PR #6143, opened 2022-12-14
Update all Doctrine deps

j0k3r:
~~⚠️ It needs https://github.com/wallabag/wallabag/pull/6136 to be merged first.~~

Also update these deps to be compatible with latest Doctrine version:
- `friendsofsymfony/oauth-server-bundle`
- `lexik/form-filter-bundle`
- `dama/doctrine-test-bundle` (but there are issue about rollbacking in some cases https://github.com/dmaicher/doctrine-test-bundle/issues/58)

The PR is WIP because I can't generate a diff from the database:

```
In Exception.php line 112:

  Unknown column type "json_array" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If this error occurs during database introspection then you might have forgotten to register all database types for a Doctrine Type
  . Use AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem with the cache or forgot some mapping information.
```

I guess it might be because of other deps not yet updated. We'll see later.


j0k3r:
I've created a separate issue (https://github.com/wallabag/wallabag/issues/6149) about the `json_array` field type to avoid blocking the update (because other ones depends on it).

nicosomb (approved):
seems ok and testsuite is green.

composer.json
```diff
@@ -56,18 +56,17 @@
...
         "doctrine/event-manager": "^1.1",
-        "doctrine/migrations": "^1.8",
+        "doctrine/migrations": "^3.2",
         "doctrine/orm": "^2.6",
-        "doctrine/persistence": "^1.3",
+        "doctrine/persistence": "^3.0",
         "enshrined/svg-sanitize": "^0.15.4",
         "friendsofsymfony/jsrouting-bundle": "^2.2",
-        "friendsofsymfony/oauth-server-bundle": "^1.5",
+        "friendsofsymfony/oauth-server-bundle": "dev-master",
```
Kdecherf, line 69: Not a big fan of putting this kind of version constraint, could we maybe fix a commit for safety?

app/DoctrineMigrations/Version20160401000000.php
```diff
@@ -70,7 +70,7 @@ public function up(Schema $schema)
...
-CREATE TABLE {$this->getTable('user')} (id INT AUTO_INCREMENT NOT NULL, username VARCHAR(180) NOT NULL, username_canonical VARCHAR(180) NOT NULL, email VARCHAR(180) NOT NULL, email_canonical VARCHAR(180) NOT NULL, enabled TINYINT(1) NOT NULL, salt VARCHAR(255) NOT NULL, password VARCHAR(255) NOT NULL, last_login DATETIME DEFAULT NULL, locked TINYINT(1) NOT NULL, expired TINYINT(1) NOT NULL, expires_at DATETIME DEFAULT NULL, confirmation_token VARCHAR(255) DEFAULT NULL, password_requested_at DATETIME DEFAULT NULL, roles LONGTEXT NOT NULL COMMENT '(DC2Type:array)', credentials_expired TINYINT(1) NOT NULL, credentials_expire_at DATETIME DEFAULT NULL, name LONGTEXT DEFAULT NULL, created_at DATETIME NOT NULL, updated_at DATETIME NOT NULL, authCode INT DEFAULT NULL, twoFactorAuthentication TINYINT(1) NOT NULL, trusted LONGTEXT DEFAULT NULL COMMENT '(DC2Type:json_array)', UNIQUE INDEX UNIQ_1D63E7E592FC23A8 (username_canonical), UNIQUE INDEX UNIQ_1D63E7E5A0D96FBF (email_canonical), UNIQUE INDEX UNIQ_1D63E7E5C05FB297 (confirmation_token), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB;
```
Kdecherf, line 73: It's a bit touchy to make changes in these queries, we may need to check that we don't break installations (_e.g. migration issue with some yunohost packages_)

composer.json
j0k3r: Good idea, I'll lock to the latest current commit

app/DoctrineMigrations/Version20160401000000.php
j0k3r: Well, I don't have the choice, I'm forced to do so because in the way we test, we run all migrations when we setup the project.
In that case the `json_array` field type was removed, so the migration can't run.

Kdecherf: I was a reminder, understood

Kdecherf (approved):

PR #6150, opened 2022-12-15
Replace SwiftMailer by Symfony Mailer

j0k3r:
I've added a new `UserMailer` because FOSUser doesn't support Symfony Mailer, so we need to create a custom class.
We'll have to mention (and update the doc) about the updated parameter for the DSN mailer.

Resources:
- https://symfony.com/doc/current/mailer.html
- https://symfonycasts.com/screencast/mailer/unit-test
- https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/Resources/doc/emails.rst#using-a-custom-mailer


Kdecherf (approved):
Looks good

nicosomb (dismissed):

nicosomb (approved):

PR #6151, opened 2022-12-15
Upgrade to Twig 3

j0k3r:
Use https://github.com/twigphp/twig-extra-bundle as a replacement (mostly for the `truncate` filter).


nicosomb (approved):
LGTM

PR #6154, opened 2022-12-15
Fix EventDispatcher & events

j0k3r:
Looks like parameter for the `->dispatch(` have been flipped (event first then event name). Define events should now extends `Symfony\Contracts\EventDispatcher\Event`


Kdecherf (approved):
Looks good

nicosomb (approved):

nicosomb:
Thank you @j0k3r !

PR #6159, opened 2022-12-16
Move to controller as a service

j0k3r:
Mostly using autowiring to inject deps.
The only tricky part was for import because all producer use the same class and have a different alias. So we must write them down in the service definition, autowiring doesn't work in that case.

Usually:
- if a controller has a constructor, it means injected services are at least re-used once in actions
- otherwise, service are injected per action

~~Also remove custom controller to handle error (just adding template is enough).~~
☝️ reverted because of format error problem (requesting json format for a 404 from the api render a html page ...)

### DIFF
--- a/app/config/routing_dev.yml
+++ b/app/config/routing_dev.yml
@@ -7,7 +7,7 @@ _profiler:
     prefix: /_profiler
 
 _errors:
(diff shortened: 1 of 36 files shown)


nicosomb:
All is red 🔴  😉

nicosomb (approved):
All is green 🟢 🎉

PR #6171, opened 2022-12-21
Properly handle `json_array` type removal

### LINKED ISSUE #6149: Fix the problem about creating a new migrations
It seems there is a problem after upgrading the whole doctrine deps (in https://github.com/wallabag/wallabag/pull/6143) about a field type `json_array`.

```
In Exception.php line 112:

  Unknown column type "json_array" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If this error occurs during database introspection then you might have forgotten to register all database types for a Doctrine Type
  . Use AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem with the cache or forgot some mapping information.
```

This must be fixed before 2.6.0.

j0k3r:
The `json_array` type was removed from DBAL v3, we should handle it using a migration to avoid error. I've also added the remove type because we need it during migration.

Following https://github.com/wallabag/wallabag/pull/6143
Fix #6149


nicosomb (approved):
seems ok for me. 

I just want to remove sqlite support and do not review anymore these kind of queries.

PR #6226, opened 2023-01-09
Add prefix for tag slugs

### LINKED ISSUE #6048: Can't add a tag with a single character like >
### Environment

* **Version**: 2.5.2
* **Installation**: locally
* **PHP version**: 8.0
* **OS**: debian
* **Database**: mariadb
* **Parameters**:

1. Save an article
2. Add a new tag to this article with `>` as name
3. On unread page, click on the tag to see all the articles linked to this tag
4. See the 404 page 

In database, no slug are stored for tags like this one, that's why we have a 404. 

To fix this bug, we need to store a slug. 
We also need to write a database migration to fetch all tags with empty slug and generate a new one.

### Expected bahaviour 

I want to see the articles with the `>` tag.

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | 
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This should be considered as a temporary fix, we may deprecate tag slugs in the future.

Fixes #6048


j0k3r:
> 2.6 or 3.0 ?

We do not remove the slug, so it can be 2.6

nicosomb (approved):
ok for me

PR #6288, opened 2023-02-07
Fix XSS on username on share page

j0k3r:
Before:

image

After:

image

Thanks to @gabriel-vernilo

### DIFF
--- a/src/Wallabag/CoreBundle/Resources/views/themes/common/Entry/share.html.twig
+++ b/src/Wallabag/CoreBundle/Resources/views/themes/common/Entry/share.html.twig
@@ -28,7 +28,7 @@
         <header class="block">
             <h1>{{ entry.title|e|raw }}</h1>
             <a href="{{ entry.url|e }}" target="_blank" rel="noopener" title="{{ 'entry.view.original_article'|trans }} : {{ entry.title|e|raw }}" class="tool">{{ entry.domainName|removeWww }}</a>
-            <p class="shared-by">{{ "entry.public.shared_by_wallabag"|trans({'%wallabag_instance%': url('homepage'), '%username%': entry.user.username})|raw }}.</p>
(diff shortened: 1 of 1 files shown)


Kdecherf (approved):

src/Wallabag/CoreBundle/Resources/views/themes/common/Entry/share.html.twig
```diff
@@ -28,7 +28,7 @@
         <header class="block">
             <h1>{{ entry.title|e|raw }}</h1>
             <a href="{{ entry.url|e }}" target="_blank" rel="noopener" title="{{ 'entry.view.original_article'|trans }} : {{ entry.title|e|raw }}" class="tool">{{ entry.domainName|removeWww }}</a>
-            <p class="shared-by">{{ "entry.public.shared_by_wallabag"|trans({'%wallabag_instance%': url('homepage'), '%username%': entry.user.username})|raw }}.</p>
+            <p class="shared-by">{{ "entry.public.shared_by_wallabag"|trans({'%wallabag_instance%': url('homepage'), '%username%': entry.user.username|escape})|raw }}.</p>
```
yguedidi, line 31: @j0k3r a more advanced solution to the issue would be to rework every place using the `|raw` filter to avoid it, as it makes useless the benefit of Twig autoescaping :)

yguedidi: what do you think?

j0k3r: Yep, a good idea. We did a `|raw` here because the translation as some html which might no be a good idea after all.

yguedidi: @j0k3r just created #6302 to cover this

PR #6289, opened 2023-02-07
Fix CSRF on user deletion

j0k3r:
The route `/account/delete` was accessible using a `GET` request, allowing any user click on the link while being connected to have his account deleted.

Thanks @leorac & @bauh0lz

### DIFF
--- a/tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
+++ b/tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
@@ -794,7 +794,7 @@ class ConfigControllerTest extends WallabagCoreTestCase
         $this->assertGreaterThan(1, $body = $crawler->filter('body')->extract(['_text']));
         $this->assertStringNotContainsString('config.form_user.delete.button', $body[0]);
 
-        $client->request('GET', '/account/delete');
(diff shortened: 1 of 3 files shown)


tests/Wallabag/CoreBundle/Controller/ConfigControllerTest.php
```diff
@@ -794,9 +794,6 @@ public function testDeleteUserButtonVisibility()
         $this->assertGreaterThan(1, $body = $crawler->filter('body')->extract(['_text']));
         $this->assertStringNotContainsString('config.form_user.delete.button', $body[0]);
 
-        $client->request('GET', '/account/delete');
```
Kdecherf, line 797: I guess we should still test that an invalid csrf token leads to an exception

Kdecherf (approved):

yguedidi:
@j0k3r a more future proof solution to this would be to rework all actions to have GET request never alter app state (session, database, files, etc). And use POST for others. It's a good practice that GET requests must be idempotent, can be called multiple time safely, and also be cached.
what do you think?

j0k3r:
I totally agree and I've no idea why we did a GET request to remove something!

PR #6290, opened 2023-02-07
Fix adding tag to entries from other people

j0k3r:
One user can add a new tag to an entry from an other user.
I've also limited tag length to 20 chars (and limit adding more than 5 tags at once)

Thanks @juylang

### DIFF
--- a/src/Wallabag/CoreBundle/Controller/TagController.php
+++ b/src/Wallabag/CoreBundle/Controller/TagController.php
@@ -17,7 +17,7 @@ use Wallabag\CoreBundle\Form\Type\RenameTagType;
 class TagController extends Controller
 {
     /**
-     * @Route("/new-tag/{entry}", requirements={"entry" = "\d+"}, name="new_tag")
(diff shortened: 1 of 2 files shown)


Kdecherf (approved):

src/Wallabag/CoreBundle/Controller/TagController.php
```diff
@@ -222,4 +234,14 @@ public function tagThisSearchAction($filter, Request $request)
 
         return $this->redirect($this->get('wallabag_core.helper.redirect')->to($request->headers->get('referer'), '', true));
     }
+
+    /**
+     * Check if the logged user can manage the given entry.
+     */
+    private function checkUserAction(Entry $entry)
```
yguedidi, line 241: @j0k3r looks like to be a hotfix and it's understandable :)
but wouldn't a cleaner solution be the use of Symfony voters? what do you think?

j0k3r: That's a good idea!

yguedidi: just created #6300 to cover this

PR #6394, opened 2023-03-27
#6242  back button on the web app goes "back" instead of to unread entries

### LINKED ISSUE #6242: have back button on the web app go "back" instead of to unread entries

javadsaeid:
| Q             | A
| ------------- | ---
| Bug fix?      | yes

Fixes #6242


Kdecherf:
Hello @javadsaeid,

The client-side `Referer` header can't be used reliably, especially for back buttons. There's also an open issue regarding a recent fix to replace the use of this header elsewhere on the application: https://github.com/wallabag/wallabag/issues/6944

Thus I'll close this PR as it requires a different approach.

PR #6525, opened 2023-05-17
Fix dark mode URL add input color

Simounet:
Fix dark mode add entry form submitting.

Before
wallabag-before

After
wallabag-after


Simounet:
And I also updated the background color for side panels.
Before
wallabag-before

After
wallabag-after

ping @hydrargyrum

Kdecherf (approved):

PR #6545, opened 2023-05-24
[FIX] round reading time in export

mart-e:
Before this commit, the exported entry (pdf, epub,...) could look like:

> Estimated reading time:
> 2.6666666666667 min

Now it will be:
> Estimated reading time
> 3 min

Screenshot from 2023-05-24 17-08-32

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | ?
| Deprecations? | no
| Tests pass?   | :crossed_fingers: 
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/EntriesExport.php
+++ b/src/Wallabag/CoreBundle/Helper/EntriesExport.php
@@ -210,7 +210,7 @@ class EntriesExport
                 $publishedDate = $entry->getPublishedAt()->format('Y-m-d');
             }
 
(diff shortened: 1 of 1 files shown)


j0k3r (approved):
Good catch, same behavior as in the [Twig template](https://github.com/wallabag/wallabag/blob/18d46f0f68174b4379c94106cefa74717ba80d16/src/Wallabag/CoreBundle/Resources/views/Entry/_reading_time.html.twig#L4) 👍

PR #6613, opened 2023-06-14
Feat/entry info UI

Simounet:
Some love for the tags UI on the entry view. Plus, I removed the annotation text when empty.

Before:
wallabag-desktop-before
wallabag-mobile-before

After:
wallabag-desktop-after
wallabag-mobile-after


nicosomb:
I like the color difference between icon and text. But on the 2nd screenshot, I have the impression that the tag is too large.

Simounet:
@nicosomb For the mobile one or both views?

Simounet:
For touchable actions, I can assure you it's not that big. You can be tricked by the cropped page I shared. Height of 32px is kind of "standard" and could be small for big fingers. 😅
Did you try it on your phone?

nicosomb (approved):

Issue #6659, opened 2023-06-22
Issue while installing Wallabag 2.6.1


prbt2016:
Hello ,

I was in the process of manual installation of wallabag 2.6.1 using Github release package on Centos 7 with PHP 7.4, MYSQL 5.5, Apache 2.2  using the following steps i.e : 

https://doc.wallabag.org/en/admin/installation/installation.html#on-shared-hosting

However after filling in parameters.yml and executing the following command i.e : 

`php bin/console wallabag:install --env=prod`

The following error is thrown ON MYSQL 5.5 and on MYSQL 5.6 environment as well i.e : 

```
 [WARNING] Some commands could not be registered:


In ExceptionConverter.php line 49:

  An exception occurred while executing a query: SQLSTATE[42S02]: Base table
  or view not found: 1146 Table '{{DBNAME}}.{{DBPREFIX}}internal_setting' doesn't
  exist


In Exception.php line 28:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wallabag261.prew
  b_internal_setting' doesn't exist


In Statement.php line 121:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wallabag261.prew
  b_internal_setting' doesn't exist


In ExceptionConverter.php line 49:

  An exception occurred while executing a query: SQLSTATE[42S02]: Base table
  or view not found: 1146 Table '{{DBNAME}}.{{DBPREFIX}}internal_setting' doesn't
  exist


In Exception.php line 28:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wallabag261.prew
  b_internal_setting' doesn't exist


In Statement.php line 121:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wallabag261.prew
  b_internal_setting' doesn't exist


In ExceptionConverter.php line 49:

  An exception occurred while executing a query: SQLSTATE[42S02]: Base table
  or view not found: 1146 Table '{{DBNAME}}.{{DBPREFIX}}internal_setting' doesn't
  exist


In Exception.php line 28:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wallabag261.prew
  b_internal_setting' doesn't exist


In Statement.php line 121:

  SQLSTATE[42S02]: Base table or view not found: 1146 Table 'wallabag261.prew
  b_internal_setting' doesn't exist




 [WARNING] Some commands could not be registered:


In ExceptionConverter.php line 49:

[...]

PR #6739, opened 2023-07-19
Fix installer when database not exists

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Documentation |no
| Translation   |no
| CHANGELOG.md  |no
| License       | MIT

I'd prefer have a regex to check the whole exception message. If someone wants to have a look on it. 

My PR is a pointer to the bug, we can improve it. 

I want to kill an other bug quickly https://github.com/wallabag/wallabag/issues/6659

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -153,7 +153,7 @@ jobs:
         run: "php bin/simple-phpunit -v"
 
   phpunit_no_database:
-    name: "PHP ${{ matrix.php }} using ${{ matrix.database }} who don't exist"
+    name: "PHP ${{ matrix.php }} using ${{ matrix.database }} without DB created"
     runs-on: "ubuntu-20.04"
     services:
       rabbitmq:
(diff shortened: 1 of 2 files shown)


app/config/parameters.clean
j0k3r, line 1: What is that file?

j0k3r (commented):
Maybe you can add an extra build without creating the MySQL or PG database first to see if it fix the problem

nicosomb:
All is green but the CI is broken 😉

.github/workflows/continuous-integration.yml
```diff
@@ -151,3 +151,65 @@ jobs:
 
       - name: "Run PHPUnit"
         run: "php bin/simple-phpunit -v"
+
+  phpunit_no_database:
```
Kdecherf, line 155: I'm not a big fan of running these new tests on all PR as it unnecessarily increases load on the CI. @nicosomb could we disable them by default (or drop the commit before merging)?

Kdecherf (changes requested):

.github/workflows/continuous-integration.yml
j0k3r: Maybe we can just keep one? For MySQL or PG

nicosomb:
@j0k3r can you help me about the new build to create please?

PR #6797, opened 2023-08-05
Use Twig instead of templating

yguedidi:
> Removed the `framework.templating` option, configure the Twig bundle instead.

from https://github.com/symfony/symfony/blob/4.4/UPGRADE-5.0.md#frameworkbundle

remove ~1180 direct depreciation, see from [2119](https://github.com/wallabag/wallabag/actions/runs/5743870218/job/15569056118#step:10:130) to [940](https://github.com/wallabag/wallabag/actions/runs/5771520789/job/15645559336?pr=6797#step:10:130) for example


src/Wallabag/CoreBundle/Controller/ExceptionController.php
```diff
@@ -1,40 +0,0 @@
...
-    protected function findTemplate(Request $request, $format, $code, $showException)
-    {
-        $name = $showException ? 'exception' : 'error';
-        if ($showException && 'html' === $format) {
-            $name = 'exception_full';
-        }
-
-        // For error pages, try to find a template for the specific HTTP status code and format
-        if (!$showException) {
-            $template = sprintf('@WallabagCore/Exception/%s.%s.twig', $name, $format);
```
yguedidi, line 23: this was the only difference from the base one, so now the base error template is moved to use the official way of overriding a template from a bundle

app/config/config.yml
```diff
@@ -87,8 +84,13 @@ doctrine_migrations:
             executed_at_column_name: 'executed_at'
 
 fos_rest:
+    zone:
+        - { path: ^/api }
+        - { path: ^/annotations }
     param_fetcher_listener: true
     body_listener: true
+    exception:
+        serializer_error_renderer: true
```
yguedidi, line 93: this is needed so that exceptions thrown manually in rest controllers like in https://github.com/wallabag/wallabag/blob/master/src/Wallabag/ApiBundle/Controller/EntryRestController.php#L313 respond with JSON instead of HTML

app/config/config.yml
```diff
@@ -87,8 +84,13 @@ doctrine_migrations:
             executed_at_column_name: 'executed_at'
 
 fos_rest:
+    zone:
+        - { path: ^/api }
+        - { path: ^/annotations }
```
yguedidi, line 89: added because of the addition of `serializer_error_renderer`

app/config/config.yml
```diff
@@ -117,9 +119,6 @@ fos_rest:
             - { path: "^/api/entries/([0-9]+)/export.(.*)", priorities: ['epub', 'mobi', 'pdf', 'txt', 'csv'], fallback_format: json, prefer_extension: false }
             - { path: "^/api", priorities: ['json', 'xml'], fallback_format: json, prefer_extension: false }
             - { path: "^/annotations", priorities: ['json', 'xml'], fallback_format: json, prefer_extension: false }
-            # for an unknown reason, EACH REQUEST goes to FOS\RestBundle\EventListener\FormatListener
-            # so we need to add custom rule for custom api export but also for all other routes of the application...
```
yguedidi, line 121: not sure about the "for custom api export" part here

Kdecherf (approved):
LGTM, made some calls on the API as well to check

PR #6809, opened 2023-08-07
Replace kernel.root_dir by kernel.project_dir

yguedidi:
> The `Kernel::getRootDir()` and the `kernel.root_dir` parameter have been removed

from https://github.com/symfony/symfony/blob/4.4/UPGRADE-5.0.md#httpkernel

### DIFF
--- a/app/config/webpack/common.js
+++ b/app/config/webpack/common.js
@@ -2,16 +2,16 @@ const path = require('path');
 const webpack = require('webpack');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
 
-const rootDir = path.resolve(__dirname, '../../../');
+const projectDir = path.resolve(__dirname, '../../../');
 
 module.exports = {
   entry: {
-    material: path.join(rootDir, './app/Resources/static/themes/material/index.js'),
-    public: path.join(rootDir, './app/Resources/static/themes/_global/share.js'),
+    material: path.join(projectDir, './app/Resources/static/themes/material/index.js'),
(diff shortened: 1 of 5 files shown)


app/config/webpack/common.js
```diff
@@ -2,16 +2,16 @@ const path = require('path');
 const webpack = require('webpack');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
 
-const rootDir = path.resolve(__dirname, '../../../');
+const projectDir = path.resolve(__dirname, '../../../');
```
yguedidi, line 5: this rename is not needed as it's in the JS, but I did it for consistency with back-end naming

docker/php/config/parameters.yml
```diff
@@ -6,7 +6,7 @@ parameters:
     database_name: ${DATABASE_NAME:-symfony}
     database_user: ${DATABASE_USER:-root}
     database_password: ${DATABASE_PASSWORD:-~}
-    database_path: '${DATABASE_PATH:-"%kernel.root_dir%/data/db/wallabag.sqlite"}'
+    database_path: '${DATABASE_PATH:-"%kernel.project_dir%/data/db/wallabag.sqlite"}'
```
yguedidi, line 9: looks like the previous default value was broken as root_dir is the directory where the kernel is, so `app/`, while `data/` is at the root of the project, not inside `app/`

nicosomb (approved):

yguedidi:
Here the equivalent on the docker repo https://github.com/wallabag/docker/pull/361

Kdecherf (approved):

Issue #2089, opened 2016-05-16
API: POST on existing entry clears starred and archived flags


Rurik19:
### Issue details

API: POST on existing entry clears starred and archived flags

### Environment

* wallabag version 2.0.4

PR #6822, opened 2023-08-09
Resolve self depreciation

yguedidi:
All self depreciation notices were about accepting null in some functions is deprecated.
And it looks like there was a bug in entry PATCH action tests that made them pass why not working as expected

### DIFF
--- a/src/Wallabag/CoreBundle/Command/UpdatePicturesPathCommand.php
+++ b/src/Wallabag/CoreBundle/Command/UpdatePicturesPathCommand.php
@@ -46,11 +46,15 @@ class UpdatePicturesPathCommand extends Command
         $io->text('Retrieve existing entries');
         $i = 1;
         foreach ($query->toIterable() as $entry) {
-            $content = str_replace($oldUrl, $this->wallabagUrl, $entry->getContent());
-            $entry->setContent($content);
+            $content = $entry->getContent();
+            if (null !== $content) {
+                $entry->setContent(str_replace($oldUrl, $this->wallabagUrl, $content));
+            }
 
-            $previewPicture = str_replace($oldUrl, $this->wallabagUrl, $entry->getPreviewPicture());
-            $entry->setPreviewPicture($previewPicture);
+            $previewPicture = $entry->getPreviewPicture();
+            if (null !== $previewPicture) {
+                $entry->setPreviewPicture(str_replace($oldUrl, $this->wallabagUrl, $previewPicture));
+            }
 
             if (0 === ($i % 20)) {
                 $this->entityManager->flush();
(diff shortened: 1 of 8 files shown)


nicosomb (dismissed):

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -948,10 +948,14 @@ public function patchEntriesAction(Entry $entry, Request $request, ContentProxy
             $entry->updateArchived((bool) $data['isArchived']);
         }
 
+        $entry->updateArchived($entry->isArchived());
```
nicosomb, line 951: So here, the updateArchived() can be called twice?

yguedidi: this addition is needed to make the test [`testSaveIsArchivedAfterPatch`](https://github.com/wallabag/wallabag/pull/6822/files#diff-b384b50ea4d69b72b2deb3d900a6da3a1f8d542c2df9abe0a442a08a669b0c5dR968-R993) pass.
if I understand it correctly, it's expected to have the archivedAt property filled/updated when a PATCH is made on an entry patching just another property when that entry has just the isArchived property to true but no archivedAt yet.
same for starred.
maybe it's the test expectation that is wrong?

j0k3r: In fact the test ensure that when you patch an entries, the `is_archived` value doesn't change.
It was related to that issue: https://github.com/wallabag/wallabag/issues/2089

So I think you can remove the double call to `updateArchived` and revert changes on the test too.

yguedidi: @j0k3r There must be a change in the test `testSaveIsStarredAfterPatch` at least, as on master it triggers a deprecation because `(new \DateTime($content['starred_at']))` is called with `$content['starred_at']` being null.
I could remove the commit "Fix entry PATCH API action" but I'd need to know what is the best way to fix that deprecation while still being aligned with what is expected :)

Hypothesis: maybe the issue is that the fixture entry that is fetched with `'isStarred' => true` condition has no starred at date?

j0k3r: > Hypothesis: maybe the issue is that the fixture entry that is fetched with `'isStarred' => true` condition has no starred at date?

That's the problem: https://github.com/wallabag/wallabag/blob/c12239c4f8799259acd4beb34f0859d002877696/src/Wallabag/CoreBundle/DataFixtures/EntryFixtures.php#L77-L88

yguedidi: @j0k3r fixed and rebased

j0k3r (approved):

PR #6823, opened 2023-08-10
Use Webpack Encore

yguedidi:
<details>
<summary>Old description</summary>
This PR is doing multiple things that could be extracted in separated PR later if you want.
I'm opening it as draft so that you can have a look at it, and so that we can start taking decisions and drafting solutions to the issues I faced.

What this PR is about more precisely:
- some assets clean up
- merge of global and material as no more theming
- follow standard convetions (root assets/ folder, build/ instead of wallassets)
- simplify Webpack config by using Webpack Encore
- better builds following Webpack Encore sane defaults (but some tweaks to fit with Wallabag needs)

edit: found a solution, see https://github.com/wallabag/wallabag/pull/6823#issuecomment-1675575464

It's really almost done and it actually works!
But there is one thing that is a blocker: the fact that production generated assets are tracked in the repository.
Why is that an issue: Webpack Encore relies mainly on 2 files, manifest.json and entrypoints.json
It's possible to have different manifests for prod and dev (what I did already), but for entrypoints.json it's not possible.
And this leads to the fact that the file is changed when running dev build or watch. Not a super DX, see commit `[WIP] issue after dev build` for what happen after to the tracked entrypoints.json

So what would be the right direction in your option?
To me, production generated assets shouldn't be tracked in the dev repository.
But I guess the docker image and ZIP packages rely on that so not that simple to manage the production assets building step outside the repository. Am I right here?

Feel free to tell me what you accept that I extract in a separated PR, because the folder structure changes are not impacted by the issue mentionned above, just the use of Webpack Encore
</details>

This adds Webpack Encore with its sane defaults, that simplifies a lot the Webpack configuration.
No more dev vs prod build as the app will use properly what have been built thanks to manifest.json file.
To build it this requires to add node to the CI.


yguedidi:
Managed to find a solution! 🎉 
It may look a bit "hacky" (and I personally think it is), but it works the way Wallabag works today, by having generated production assets be part of the source code.
I used the [possibility to have multiple webpack builds](https://symfony.com/doc/4.4/frontend/encore/advanced-config.html#defining-multiple-webpack-configurations) to solve the issue, one build for prod in a public folder that is tracked by git, and another build for dev generated in another folder that is not tracked by git.
Not sure this feature was made for that usage but it's OK for now I would say!
The cleaner solution not having production assets in the source code is another topic to be solved later.

.babelrc
```diff
@@ -3,7 +3,8 @@
     [
       "@babel/preset-env",
       {
-        "modules": false
+        "useBuiltIns": "usage",
+        "corejs": "3.23"
```
yguedidi, line 7: to match the Webpack Encore Bundle recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/webpack.config.js#L55-L56

.gitignore
```diff
@@ -59,3 +59,5 @@ specialexport.json
 
 # Custom CSS file
 web/custom.css
+
+yarn-error.log
```
yguedidi, line 63: to match the Webpack Encore Bundle recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/manifest.json#L16

package.json
```diff
@@ -58,6 +60,7 @@
     "postcss": "^8.4.27",
     "postcss-loader": "^7.3.3",
     "postcss-scss": "^4.0.6",
+    "regenerator-runtime": "^0.14.0",
```
yguedidi, line 63: to match the Webpack Encore Bundle recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/package.json#L9

package.json
```diff
@@ -70,9 +73,9 @@
     "url-loader": "^4.1.1",
     "webpack": "^5.88.2",
     "webpack-cli": "^5.1.4",
-    "webpack-dev-server": "^4.15.1",
     "webpack-manifest-plugin": "^5.0.0",
-    "webpack-merge": "^5.9.0"
+    "webpack-merge": "^5.9.0",
+    "webpack-notifier": "^1.15.0"
```
yguedidi, line 78: to match the Webpack Encore Bundle recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/package.json#L12

package.json
```diff
@@ -44,8 +44,10 @@
     "@babel/core": "^7.22.9",
     "@babel/eslint-parser": "^7.22.9",
     "@babel/preset-env": "^7.22.9",
+    "@symfony/webpack-encore": "^4.4.0",
     "autoprefixer": "^10.4.14",
     "babel-loader": "^9.1.3",
+    "core-js": "^3.32.0",
```
yguedidi, line 50: to match the Webpack Encore Bundle recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/package.json#L8

package.json
```diff
@@ -70,9 +73,9 @@
     "url-loader": "^4.1.1",
     "webpack": "^5.88.2",
     "webpack-cli": "^5.1.4",
-    "webpack-dev-server": "^4.15.1",
```
yguedidi, line 73: will still be installed as a transitive dependency

postcss.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  plugins: {
+    autoprefixer: {},
+  },
+};
```
yguedidi, line 5: from the doc: https://symfony.com/doc/4.4/frontend/encore/postcss.html

src/Wallabag/CoreBundle/Resources/views/Entry/share.html.twig
```diff
@@ -17,12 +17,8 @@
         <meta name="twitter:site" content="@wallabagapp" />
         <meta name="twitter:title" content="{{ entry.title|e|raw }}" />
         <meta name="twitter:description" content="{{ entry.content|striptags|slice(0, 300)|raw }}&hellip;" />
-        {% if app.debug %}
-            <script src="{{ asset('wallassets/public.dev.js') }}"></script>
-        {% else %}
-            <link rel="stylesheet" href="{{ asset('wallassets/public.css') }}">
-        {% endif %}
 
+        {{ encore_entry_link_tags('public', null, app.environment == 'dev' ? 'dev' : 'prod') }}
```
yguedidi, line 21: now test environment is using prod assets that are generated.
did this change because on CI dev assets are not generated, and it can happen that on dev we don't have them neither, so using prod assets ensure we don't have assets missing errors when running tests

webpack.config.js
```diff
@@ -1,8 +1,58 @@
...
+  if (!enc.isRuntimeEnvironmentConfigured()) {
+    enc.configureRuntimeEnvironment(process.env.NODE_ENV || 'dev');
+  }
 
-module.exports = buildConfig;
+  enc
+    .addEntry('material', './assets/material.js')
+    .addEntry('public', './assets/public.js')
+    .splitEntryChunks()
+    .enableSingleRuntimeChunk()
```
yguedidi, line 13: to match the Webpack Encore Bundle recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/webpack.config.js#L28-L33

assets/js/shortcuts/main.js
```diff
@@ -90,3 +90,17 @@ $(document).ready(() => {
     }
   });
 });
+
+/* Shortcuts */
+
+/* Go to */
```
j0k3r, line 96: Is that coming from a deleted file?

web/build/118.809c3054.css
j0k3r, line 1: Won't the name change each time the file is built?

j0k3r (commented):
We are _tracking_ prod asset into Git because we don't want end user to install node, yarn, etc.. to build asset on their side.
If assets are ready from the repo, it's way much easier for us. We do already have a lot of request because of server installation problem and we don't want to add more with the installation of node, etc 😬

j0k3r:
> a similar thing can be done when creating an archive release.

You are totaly right, this can be done when we create the archive ready to be installed and attached to each release.
But then, we'll need to improve the documentation about installing tools (node, yarn, etc.) to build assets locally when people _only_ clone the repository.

yguedidi:
@j0k3r for reference there are https://github.com/wallabag/docker/pull/362 for Docker and https://github.com/wallabag/wallabag/pull/6982 for ZIP releases that build assets on the fly for those artifacts.
once accepted we will be able to remove built assets from the git repository!

web/build/118.809c3054.css
yguedidi: yes, but as the goal is to remove prod built assets from the repo the build folder will be ignored :)

.eslintrc.json
```diff
@@ -1,6 +1,9 @@
 {
     "extends": "airbnb-base",
     "parser": "@babel/eslint-parser",
+    "parserOptions": {
+        "requireConfigFile": false
```
yguedidi, line 5: added this because I removed the .babelrc file to favor the configuration through Webpack Encore as from the recipe, see https://github.com/symfony/recipes/blob/main/symfony/webpack-encore-bundle/1.10/webpack.config.js#L54-L57

GNUmakefile
```diff
@@ -40,8 +40,12 @@ run: ## Run the wallabag built-in server
...
 
 build: ## Run webpack
-	@$(YARN) install
-	@$(YARN) build:$(ENV)
+	$(YARN) install
+	@if [ "$(ENV)" = "prod" ]; then \
+		$(YARN) build; \
+	else \
+		$(YARN) dev; \
+	fi;
```
yguedidi, line 48: to me here we should always use dev command, assuming the make file is more for dev purpose

.eslintrc.json
yguedidi: tell me if you prefer it the other way around by keeping the .babelrc file

.github/workflows/continuous-integration.yml
```diff
@@ -76,6 +82,12 @@ jobs:
         with:
           composer-options: "--optimize-autoloader --prefer-dist"
 
+      - name: "Install dependencies with Yarn"
+        run: yarn install
+
+      - name: "Build assets with Yarn"
+        run: yarn build:dev
```
yguedidi, line 89: needed for tests to run as now need the manifest.json file

package.json
```diff
@@ -95,8 +99,9 @@
     "waypoints": "^4.0.1"
   },
   "scripts": {
-    "watch": "webpack-dev-server --env=dev",
-    "build:dev": "webpack --env=dev",
-    "build:prod": "webpack --env=prod"
+    "dev-server": "encore dev-server",
+    "build:dev": "encore dev",
```
yguedidi, line 103: default is `dev`, kept our naming to reduce diff for now

package.json
```diff
@@ -95,8 +99,9 @@
...
+    "watch": "encore dev --watch",
+    "build:prod": "encore production --progress"
```
yguedidi, line 105: default is `build`, kept our naming to reduce diff for now

Kdecherf:
Hello @yguedidi,

I made a quick test on my local setup, looks good :+1:

Kdecherf (approved):
LGTM, would like an additional approval

yguedidi:
Thank you @Kdecherf ! @j0k3r @nicosomb when you have time :)

j0k3r (approved):
Works perfectly locally.

Are you planning to do sth about deprecated & warning messages?

```

 DEPRECATION  The ESLint integration is now deprecated and will be removed in the next major release. Please use the standalone ESLint CLI instead, or use the ESLint Webpack plugin directly.
 WARNING  Webpack Encore requires version ^13.0.0 || ^14.0.0 of sass-loader. Your version 16.0.4 is too new. The related feature *may* still work properly. If you have issues, try downgrading the library, or upgrading Encore.
 WARNING  Webpack Encore requires version ^13.0.0 || ^14.0.0 of sass-loader. Your version 16.0.4 is too new. The related feature *may* still work properly. If you have issues, try downgrading the library, or upgrading Encore.
```

yguedidi:
Yes I will

PR #6850, opened 2023-08-17
Use Symfony Local Web Server

yguedidi:
WebServerBundle is deprecated in Symfony 4.4 and is removed in Symfony 5.
This PR remove it and make use of the Symfony CLI and its Local Web Server to run Wallabag locally.

Just changing the Dockerfile exposed an issue: the IP visible by Symfony is not anymore 127.0.0.1 but the Docker container IP and so `app_dev.php` block all requests.
To solve that, we needed the `TRUSTED_PROXIES` feature to allow Docker container IPs, feature that is available in an up to date `index.php` file instead of both `app.php` and `app_dev.php`.
So I updated some files to match as close as possible to [the framework-bundle recipe in version 4.4](https://github.com/symfony/recipes/tree/main/symfony/framework-bundle/4.4)
If it's requested I can extract those updates in a prerequisite PR to ease the review.


.env
```diff
@@ -0,0 +1,4 @@
+APP_ENV=dev
+APP_SECRET=ch4n63m31fy0uc4n
+TRUSTED_PROXIES=127.0.0.1/8,10.0.0.0/8,172.16.0.0/12,172.17.0.0/16,172.18.0.0/15,172.20.0.0/14,172.24.0.0/13,192.168.0.0/16
```
yguedidi, line 3: original value is `127.0.0.0/8,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16`
updated to add more based on [this article](https://jolicode.com/blog/symfony-reverse-proxies-and-ip-protection#reverse-proxies-in-docker)

.env
```diff
@@ -0,0 +1,4 @@
+APP_ENV=dev
+APP_SECRET=ch4n63m31fy0uc4n
```
yguedidi, line 2: value taken from `docker/php/env.example`

.gitignore
```diff
@@ -20,9 +29,6 @@ docker-compose.override.yml
 # Parameters
 /app/config/parameters.yml
 
-# Managed by Composer
-/vendor/
```
yguedidi, line 24: removed as covered by the ignore rules from FrameworkBundle recipe

app/config/parameters.yml.dist
```diff
@@ -30,9 +30,6 @@ parameters:
 
     locale: en
 
-    # A secret key that's used to generate certain security-related tokens
-    secret: CHANGE_ME_TO_SOMETHING_SECRET_AND_RANDOM
```
yguedidi, line 34: removed as it became an env var

web/index.php
```diff
@@ -0,0 +1,30 @@
...
+
+if ($trustedProxies = $_SERVER['TRUSTED_PROXIES'] ?? false) {
+    Request::setTrustedProxies(explode(',', $trustedProxies), Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO);
+}
+
+if ($trustedHosts = $_SERVER['TRUSTED_HOSTS'] ?? false) {
+    Request::setTrustedHosts([$trustedHosts]);
+}
+
+$kernel = new AppKernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
```
yguedidi, line 22: original is `Kernel` from `App` namespace, adapted to work with existing `AppKernel`

web/index.php
```diff
@@ -0,0 +1,30 @@
...
+//$kernel = new AppCache($kernel);
+
+// When using the HttpCache, you need to call the method in your front controller instead of relying on the configuration parameter
+//Request::enableHttpMethodParameterOverride();
```
yguedidi, line 26: this was not in original from recipe, added because it was in `app.php`

web/.htaccess
```diff
@@ -3,7 +3,7 @@
 # mod_rewrite). Additionally, this reduces the matching process for the
 # start page (path "/") because otherwise Apache will apply the rewriting rules
 # to each configured DirectoryIndex file (e.g. index.php, index.html, index.pl).
-DirectoryIndex app.php
+DirectoryIndex index.php
```
yguedidi, line 6: searched for occurences of `app.php` and `app_dev.php` in the project to use the new common file

config/bootstrap.php
yguedidi, line 1: added as is from the recipe

docker/php/Dockerfile
```diff
@@ -94,6 +100,7 @@ RUN groupmod -g 1000 www-data ; \
     chown -R www-data: /var/www/html \
         /usr/local/etc/php/conf.d/wallabag-php.ini \
         /var/www/.cache \
-        /var/www/.yarnrc
+        /var/www/.yarnrc \
+        /var/www/.symfony5
```
yguedidi, line 104: needed because Symfony CLI writes in it

docker/php/Dockerfile
```diff
@@ -77,10 +80,13 @@ RUN mkdir -p \
         /var/www/html/web/assets \
         /var/www/html/data \
         /var/www/html/data/db \
-        /var/www/.cache
+        /var/www/.cache \
+        /var/www/.symfony5
+
+WORKDIR /var/www/html
```
yguedidi, line 86: needed because it seems `php:8.1` image uses `/` as default working dir

docker/php/Dockerfile
```diff
@@ -15,6 +15,8 @@ RUN apt-get update \
 RUN curl 'https://deb.nodesource.com/gpgkey/nodesource.gpg.key' | apt-key add - \
     && echo "deb https://deb.nodesource.com/node_${NODE_VERSION}.x $(lsb_release -cs) main" > /etc/apt/sources.list.d/nodesource.list
 
+RUN curl -1sLf 'https://dl.cloudsmith.io/public/symfony/stable/setup.deb.sh' | bash
```
yguedidi, line 18: followed default installation instruction for Ubuntu/Debian from the [official website](https://symfony.com/download)

.env.test
```diff
@@ -0,0 +1,5 @@
+KERNEL_CLASS='AppKernel'
```
yguedidi, line 1: original is `App/Kernel`, adapted to work with existing `AppKernel`

.env.test
```diff
@@ -0,0 +1,5 @@
...
+APP_SECRET='$ecretf0rt3st'
+SYMFONY_DEPRECATIONS_HELPER=weak
```
yguedidi, line 3: original is `999999`, update to current configuration

tests/bootstrap.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+require dirname(__DIR__) . '/config/bootstrap.php';
```
yguedidi, line 3: [original](https://github.com/symfony/recipes/blob/main/symfony/phpunit-bridge/4.3/tests/bootstrap.php) wrap this by a condition that lead to a PHPStan error.
as the file exist for sure in our case, I simplified the code to avoid that error

.gitignore
```diff
@@ -13,16 +27,11 @@
 !/bin/console
 .php-cs-fixer.php
 .php-cs-fixer.cache
-.phpunit.result.cache
-phpunit.xml
```
yguedidi, line 17: removed as covered by the ignore rules from phpunit-bridge recipe

.env.test
```diff
@@ -0,0 +1,5 @@
...
+PANTHER_APP_ENV=panther
```
nicosomb, line 4: `symfony\panther` is not used in our project

nicosomb (changes requested):

.env.test
yguedidi: @nicosomb I know, I kept it to be as close as possible to the recipe. Maybe one day it will be used for e2e tests. And as it's a env var I don't think it hurts having it.
Another option would be to make those a comment like TRUSTED_HOSTS, what do you think?
If not OK I can of course remove them :)

j0k3r (commented):
We should warn that people will have to update their server configuration because `web/app.php` is replaced by `web/index.php`.
https://github.com/search?q=repo%3Awallabag%2Fdoc%20app.php&type=code

So maybe we should update the documentation to explain that starting from 2.7.0 version, the index is now `web/index.php`

nicosomb:
@yguedidi do you have some work to do on this PR?

nicosomb:
So, you suggest to release a 2.6.x ?

yguedidi:
@j0k3r @Kdecherf any opinion on is the switch to env vars an issue for a minor/patch release?

PR #6888, opened 2023-08-25
Feat/domain status filters html types

Simounet:
Changed the types and accepted values for status and domain filters (right column). Better for mobile inputs.


nicosomb:
@Simounet could you please fix your PR with php cs ?

j0k3r:
Shouldn't we target 2.6.5 for that?

Simounet:
I'll change the branch and push again.

Simounet:
Working with everything but PHP7.4. Any clues?

j0k3r:
Maybe be cause of https://github.com/wallabag/wallabag/blob/49eea9f92173876213c5cedab1a9df5cdc6c42a9/tests/Wallabag/CoreBundle/Controller/EntryControllerTest.php#L1390

Simounet:
It fails before that test and only with 7.4. 😢

j0k3r:
I've tested locally on PHP 7.4 and got: 
image

Simounet:
Working when value cast as integer.

nicosomb:
PHP 8.x is more restrictive for that.

Simounet:
Yep, I was expecting the opposite. Working on 7.x and not working on 8+. Glad to see it working anyway.

j0k3r (approved):

PR #6906, opened 2023-08-29
Add a validator on URL entity

### LINKED ISSUE #6863: The URL is not checked before saving it
When posting a new URL, we don't check if the URL is correct (by starting with `https://...`, etc). See https://www.php.net/manual/en/filter.filters.validate.php and `FILTER_VALIDATE_URL`. 

This bug is available on API side and web side (because the only check is the `<input type="url"` ... you can easily bypass it).

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    |no
| Deprecations? | no
| Tests pass?   | yes
| Documentation |no
| Translation   |no
| CHANGELOG.md  |no
| License       | MIT

Fixes #6863


Kdecherf (commented):
@nicosomb could you rebase and target the master branch please?

j0k3r (approved):

PR #6909, opened 2023-08-31
Feat/homepage perfs

Simounet:
Some optimisations mostly for the homepage. Started from ~600ms to ~260ms with 16.5k entries and A LOT of untagged entries.
- removed useless title on the footer's stats (called twice)
- untagged entries counter removed (there is no counters on other entries, we should treat it like a classic filter)
- SQL count simplification (not sure about this one because of the Postgres comment you left but it was from 2016 and I'm not using a column in `COUNT()`)


Simounet:
I have to find a solution for Postgres. 😅

nicosomb:
The 2 first changes are ok for me. 

About the last one, I don't understand why we have so many failing tests ...

Simounet:
It's working with separated methods just for COUNT purpose, not using `GROUP BY`.

j0k3r (approved):

nicosomb (approved):
:tada:

PR #6924, opened 2023-09-06
Force secure cookie on HTTPS connection

j0k3r:
Cookie will be set as `secure` when the connection will be on HTTPS otherwise the cookie will be defined as `secure: false`.

I used `mkcert` to  test it locally.

image

See:
- https://symfony.com/doc/4.4/reference/configuration/framework.html#cookie-secure
- [CWE-614: Sensitive Cookie in HTTPS Session Without 'Secure' Attribute ](https://cwe.mitre.org/data/definitions/614.html)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -29,6 +29,7 @@ framework:
         # handler_id set to null will use default session handler from php.ini
         handler_id: session.handler.native_file
         save_path: "%kernel.project_dir%/var/sessions/%kernel.environment%"
+        cookie_secure: auto
     fragments: ~
     http_method_override: true
     assets: ~


.gitignore
```diff
@@ -59,3 +59,6 @@ specialexport.json
 
 # Custom CSS file
 web/custom.css
+
+# Local SSL certificates
+*.pem
```
yguedidi, line 64: should this really be part of the wallabag gitignore? better add it in a global gitignore locally :)
https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

app/config/config.yml
```diff
@@ -29,6 +29,7 @@ framework:
         # handler_id set to null will use default session handler from php.ini
         handler_id: session.handler.native_file
         save_path: "%kernel.project_dir%/var/sessions/%kernel.environment%"
+        cookie_secure: "auto"
```
yguedidi, line 32:
```suggestion
        cookie_secure: auto
```

yguedidi (changes requested):

.gitignore
j0k3r: I was wondering too but I was thinking that if an other developer setup wallabag to run locally, maybe pem files will be defined in the root folder?

yguedidi: > if an other developer setup wallabag to run locally

and configure HTTPS, so multiple steps that reduce the chances that this happen.
to me either wallabag provides HTTPS support directly, either it's up to the dev that manually configure HTTPS to ignore any extra file.

yguedidi (approved):

PR #7115, opened 2023-12-12
Flex structure

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This PR is initiated from the work on #6850 as it needed the some files to be updated from their updated version from Flex recipes, namely the framework-bundle, console and phpunit-bridge recipes.

So I decided to extract that update in a separate PR, but instead of updating just the files the other PR needed, partially taking the content from the recipe, I decided that it could be the opportunity to embrasse the new Flex structure, and update completely the files from above listed recipes, and all this without yet using Flex itself!

As I want this to be accepted in a minor version of wallabag, so that we can advance on upgrading to Symfony 5, I did my best to make the changes in a backward compatible way!
What I mean by "BC way"?
I covered by this (draft) BC policy the case of users using wallabag "as is", not customizing its code or something else, just configure the parameters requested during the setup and that's it.
And I want them to be able to upgrade their wallabag instance to the new minor, is it setup by the Docker, a git clone, or via ZIP file.

So I identified those critical pieces that must stay so that an upgrade setup continues to work:
- parameters.yml file, contains the configuration
- parameters.yml.dist file, model used to generate the configuration
- web/ folder, used as document root by server configuration
- app.php file, used as index file by server configuration
- app_dev.php files, maybe used by server configuration, shouldn't be used in prod!
- var/logs/ folder configured to be writable
- var/cache/ folder configured to be writable
- var/sessions/ folder configured to be writable
- AppCache.php file as maybe app.php was altered to use it
- AppKernel.php file as used by app.php and app_dev.php

Leaving those elements should keep a setup working, even if wallabag internals change!

It may look like a very big PR, 62 commits, 188 files changed... but I don't think it is!
I did my best to tell a story, showing the path I went through to achieve the goal, with baby steps.
This means that even if they are lot of commits, each should be a very easy to review!
So you guessed it, the recommended way to review this is commit by commit, open each in a new tab.
One important thing: I did my best to keep tests green at each commit!
Hope those efforts will really help reviewing this PR :)

I'm opening this PR as draft but it's not, it's just that there are some fixup commit that don't make sense on their own, but are there to highlight some minor changes that are needed by the commit they fix to make it work.
Those changes would have been hard to catch if done directly along other changes of the commit they fix.

I will go through the commits and add some commit related comments to add more details.


app/config/config.yaml
```diff
@@ -25,3 +24,4 @@ imports:
     - { resource: ../../config/packages/twig.yaml }
     - { resource: ../../config/packages/validator.yaml }
     - { resource: ../../config/packages/wallabag.yaml }
+    - { resource: ../../config/services.yaml }
```
yguedidi: this move is done because it's the way the new Kernel will load them

app/AppKernel.php
```diff
@@ -1,101 +1,7 @@
 <?php
 
-use Symfony\Component\Config\Loader\LoaderInterface;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Symfony\Component\HttpKernel\Kernel;
+use Wallabag\Kernel;
```
yguedidi, line 3: decided to use Wallabag as root namespace instead of the default App one.

config/routes/dev/routes.yaml
```diff
@@ -1,10 +1,5 @@
-_wdt:
-    resource: "@WebProfilerBundle/Resources/config/routing/wdt.xml"
-    prefix: /_wdt
-
-_profiler:
-    resource: "@WebProfilerBundle/Resources/config/routing/profiler.xml"
-    prefix: /_profiler
+WebProfilerBundle:
```
yguedidi: using the bundle name here as done in the recipe, I applied this pattern for other routes move, as there is no chances that the bundle name is used as a route name, so no conflicts

src/Wallabag/CoreBundle/Resources/views/Entry/entries.xml.twig
```diff
@@ -27,7 +27,7 @@
         <name>{{ user }}</name>
     </author>
     <icon>{{ asset('favicon.ico') }}</icon>
-    <logo>{{ asset('bundles/wallabagcore/themes/_global/img/logo-square.png') }}</logo>
+    <logo>{{ asset('img/logo-square.png') }}</logo>
```
yguedidi, line 30: not really sure about this one to be honest, but it looks it was broken before as there is no `themes/_global/img/logo-square.png` in CoreBundle

config/services.yaml
```diff
@@ -4,6 +4,7 @@ imports:
     - { resource: parameters_addons.yaml }
 
 parameters:
+    env(APP_SECRET): '%secret%'
```
yguedidi, line 13: this is the trick to preserve BC! if the APP_SECRET env var is not set, it will use the value from the `%secret%` parameter!

config/packages/test/monolog.yaml
yguedidi, line 1: this is a pure copy of the one from dev config, as the test config was extending from dev one, but there one this monolog config and a single setting that needed to be added, so no more extending from dev

config/packages/test/web_profiler.yaml
```diff
@@ -1,5 +1,6 @@
 framework:
     profiler:
+        only_exceptions: false
```
yguedidi, line 3: this comes from the dev config

config/services_test.yaml
```diff
@@ -1,3 +1,6 @@
+imports:
+    - { resource: parameters_test.yaml }
```
yguedidi, line 2: needed so tests get parameters test values to override defaults ones from normal parameters.yml, as we're not using env vars (yet)

.env
```diff
@@ -0,0 +1,6 @@
+###> symfony/framework-bundle ###
```
yguedidi, line 1: even not installed with Flex, used this syntax so Flex is ready to identify those places when we will install it and update recipes

.env
```diff
@@ -0,0 +1,6 @@
...
+APP_ENV=dev
+APP_SECRET=CHANGE_ME_TO_SOMETHING_SECRET_AND_RANDOM
```
yguedidi, line 3: being here it means dev setup will use the env var, so this needs to be updated.
maybe I should remove it? kept it because it impacts only dev setup, not prod ones

config/packages/framework.yaml
```diff
@@ -1,11 +1,19 @@
...
+    # Enables session support. Note that the session will ONLY be started if you read or write from it.
+    # Remove or comment this section to explicitly disable session support.
     session:
         # handler_id set to null will use default session handler from php.ini
         handler_id: session.handler.native_file
         save_path: "%kernel.project_dir%/var/sessions/%kernel.environment%"
         cookie_secure: auto
-    fragments: ~
-    http_method_override: true
+        cookie_samesite: lax
```
yguedidi, line 14: this one I decide to keep it as it's security related

config/routes/dev/framework.yaml
```diff
@@ -0,0 +1,3 @@
+_errors:
+    resource: '@FrameworkBundle/Resources/config/routing/errors.xml'
```
yguedidi, line 2: this should remove some deprecation :)

config/services.yaml
```diff
@@ -37,6 +43,15 @@ services:
...
             $fonts: '%wallabag_core.fonts%'
 
+    # makes classes in src/ available to be used as services
+    # this creates a service per class whose id is the fully-qualified class name
+    # App\:
+    #     resource: '../src/'
+    #     exclude:
+    #         - '../src/DependencyInjection/'
+    #         - '../src/Entity/'
+    #         - '../src/Kernel.php'
```
yguedidi, line 53: kept this as a model for later

config/packages/framework.yaml
```diff
@@ -15,5 +15,5 @@ framework:
 
     #esi: true
     fragments: true
-    php_errors:
-        log: true
+    #php_errors:
+    #    log: true
```
yguedidi, line 19: commented as was not in our config

web/app.php
```diff
@@ -1,15 +1,6 @@
...
-$kernel = new AppKernel('prod', false);
-//$kernel = new CacheKernel($kernel);
-
-// When using the HttpCache, you need to call the method in your front controller instead of relying on the configuration parameter
-//Request::enableHttpMethodParameterOverride();
-$request = Request::createFromGlobals();
-$response = $kernel->handle($request);
-$response->send();
-$kernel->terminate($request, $response);
+require __DIR__ . '/../public/index.php';
```
yguedidi, line 6: this is the trick that keeps BC, keeping the app.php file but internally using the new index one.
small BC break here is for users who have altered their app.php file to use the AppCache, now they'll need to alter the index.php to use CacheKernel.
see https://symfony.com/doc/4.4/http_cache.html#symfony-reverse-proxy

I think it's acceptable as I don't think users do that, maybe I'm wrong

PR #7132, opened 2023-12-21
wallassets folder added to .gitignore

Simounet:
If I get it correctly, the assets are generated during the release process so we should avoid the `wallassets` folder to be commited again. ping @yguedidi


yguedidi (approved):
less error prone this way indeed!

PR #7134, opened 2023-12-22
Update deps & Node 20

j0k3r:
Mostly to free `symfony/phpunit-bridge`.

Also raise minimum version of node to 20, v16 is unmaintained since October 2023.


nicosomb (approved):

lapineige:
Going through the release notes, I found this... Does that mean Wallabag (now ?) required NodeJS to be installed in addition to php ? 🤔

j0k3r:
Not at all. We are using Node to *build* assets, once they are built, you don't need Node anymore.
You are requiring to install Node if you want to work on wallabag using an unreleased version (like cloing the master).

Otherwise, assets are built during the packaging of wallabag. So end user, don't need Node to run wallabag.

lapineige:
Great, thanks for the explanation. So nothing to change on Yunohost side, that's handy :)

PR #7135, opened 2023-12-24
Use wallabag rulerz packages

yguedidi:
To prepare for Symfony 5 upgrade


composer.json
```diff
@@ -108,6 +106,7 @@
         "predis/predis": "^2.0.3",
         "psr/http-message": "^1.0",
         "psr/log": "^1.1",
+        "rulerz-php/doctrine-orm": "dev-master",
```
yguedidi, line 109: added because wallabag rulerz packages are based on master from original packages, not the released 0.x ones.
no big changes, just that the support for databases was extracted in separate packages

composer.json
```diff
@@ -150,6 +149,8 @@
         "twig/string-extra": "^3.4",
         "twig/twig": "^3.4.3",
         "wallabag/phpepub": "^4.0.10",
+        "wallabag/rulerz": "dev-master",
+        "wallabag/rulerz-bundle": "dev-master",
```
yguedidi, line 153: I think it's OK to use dev-master for those, no plan to do a release as we will not develop new features etc, just maintaining support for recent Symfony versions

app/config/config.yml
```diff
@@ -210,7 +210,7 @@ scheb_two_factor:
         template: "@WallabagUser/Authentication/form.html.twig"
         mailer: Wallabag\UserBundle\Mailer\AuthCodeMailer
 
-kphoen_rulerz:
+rulerz:
```
yguedidi, line 213: same here, wallabag rulerz bundle is based on the master from rulerz bundle, that has only a single commit on to if 0.x, that do this change: https://github.com/K-Phoen/RulerZBundle/commit/c9dc901c32cc6049df710dfeb8ceef01153ea3f1#diff-f5f83cd60e81d7bfdde9e349048d2089be00e9173cd2a8eb3af75c426aab8f48R70

j0k3r (approved):

Kdecherf (approved):

PR #7140, opened 2023-12-25
Remove session-based redirection

### LINKED ISSUE #6944: Session-based redirect may redirect to incorrect pages
The PR #6119, which replaces redirects based on client-side referer header with server-side session parameter, introduces a bug.

When having several tabs open on the same browser, you may be incorrectly redirected to a unrelated page, based on the most recent opened tab.

### Environment

* **Version**: 2.6.6
* **Installation**: docker
* **PHP version**: 8.1
* **OS**: Linux
* **Database**: PostgreSQL
* **Parameters**: -

On _tabA_, from _viewA_ you click on _linkA_, then on _tabB_, from _viewB_ you click on _linkB_.
If you delete _linkA_, _tabA_ will be redirected to _linkB_ instead of _viewA_.

This is due to the fact that the server-side session parameter `prevUrl` is shared between all tabs of a user's browser.

A better fix would be to add a query param to all links in order to know where to go when a redirect is needed.

yguedidi:
fixes #6944

easier to review commit by commit 😉

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -46,11 +46,6 @@ parameters:
             path: tests/Wallabag/CoreBundle/Controller/FeedControllerTest.php
 
         -
-            message: "#^Call to method generate\\(\\) on an unknown class PHPUnit_Framework_MockObject_MockObject\\.$#"
-            count: 2
-            path: tests/Wallabag/CoreBundle/Helper/RedirectTest.php
-
-        -
             message: "#^Property Tests\\\\Wallabag\\\\CoreBundle\\\\Helper\\\\RedirectTest\\:\\:\\$routerMock has unknown class PHPUnit_Framework_MockObject_MockObject as its type\\.$#"
             count: 1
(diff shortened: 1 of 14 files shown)


j0k3r:
How is a link generated with the `redirect` attribute in it? I guess it's displayed in the URl? Won't this be too ugly?

yguedidi:
The redirect GET parameter is indeed part of the URL, but it's not likely to be seen by users, because the links are actions that will redirect right after saving, so the parameter will not appear in the address bar as the user will end on the final URL, without the redirect param

j0k3r (approved):

Kdecherf (approved):
Looks good.

It should be noted that `redirect` can be a misleading term here: depending on the value of `config_action_mark_as_read` the redirect path may not be used.

yguedidi:
> It should be noted that `redirect` can be a misleading term here: depending on the value of `config_action_mark_as_read` the redirect path may not be used.

@Kdecherf you're right! merging to unblock the 2.6.8 release, could be renamed later, if we find a better name 😅

PR #7153, opened 2024-01-02
Drop database if exists

yguedidi:
Just because it's safer and avoid to trigger an error, just in case :)


yguedidi:
I understand now why it wasn't there in the first place 😅

PR #7155, opened 2024-01-03
Use PHPUnit directly

yguedidi:


phpstan.neon
```diff
@@ -13,7 +13,4 @@ parameters:
     doctrine:
         objectManagerLoader: tests/object-manager.php
 
-    bootstrapFiles:
-        - vendor/bin/.phpunit/phpunit-8.5-0/vendor/autoload.php
```
yguedidi, line 17: I always get PHPUnit 9.6 locally, so to run PHPStan I had to edit this committed files...

phpunit.xml.dist
```diff
@@ -34,4 +34,8 @@
     <extensions>
         <extension class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension" />
     </extensions>
+
+    <listeners>
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
+    </listeners>
```
yguedidi, line 40: to preserve PHPUnit Bridge deprecation notices after test run

.github/workflows/coding-standards.yml
```diff
@@ -47,9 +47,6 @@ jobs:
       - name: "Generate test cache for PHPStan"
         run: "php bin/console cache:clear --env=test"
 
-      - name: "Install PHPUnit for PHPStan"
-        run: "php bin/simple-phpunit install"
```
yguedidi, line 51: not needed anymore as PHPUnit is already there now

j0k3r (commented):
Is there any benefit to use it directly?

yguedidi:
> Is there any benefit to use it directly?

yes:

> I always get PHPUnit 9.6 locally, so to run PHPStan I had to edit this committed files...

it's not a good DX to always have that uncommitted change when working locally.
also, and I should have mentioned it earlier, currently PHPUnit is a transitive dev dependency, while we use it a lot (test classes).
this makes it a root dev dependency

I think also it's qui standard to have PHPUnit as a dev dependency :)

finally, the bridge is a workaround to add features that are not supported by PHPUnit, and I think it will become more and more useless as PHPUnit get better (I think PHPUnit 10 support showing the depreciation notices)

j0k3r (approved):

PR #7161, opened 2024-01-07
Make tests use dev database config

yguedidi:
Benefits:
- easier testing on different databases locally
- prepare for the use of environment variables for configuration

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -97,5 +97,47 @@ class AppKernel extends Kernel
             // $container->setParameter('container.dumper.inline_class_loader', true);
             $container->addObjectResource($this);
         });
+
+        $loader->load(function (ContainerBuilder $container) {
+            $this->processDatabaseParameters($container);
+
+            if ('test' === $container->getParameter('kernel.environment')) {
+                $this->processDatabaseParameters($container, true);
+            }
+        });
+    }
+
+    private function processDatabaseParameters(ContainerBuilder $container, bool $isTest = false)
+    {
+        $prefix = '';
+
+        if ($isTest) {
+            $prefix = 'test_';
+        }
+
+        switch ($container->getParameter($prefix . 'database_driver')) {
+            case 'pdo_mysql':
+                $scheme = 'mysql';
+                break;
+            case 'pdo_pgsql':
+                $scheme = 'pgsql';
+                break;
+            case 'pdo_sqlite':
+                $scheme = 'sqlite';
+                break;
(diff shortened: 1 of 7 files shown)


app/AppKernel.php
```diff
@@ -97,5 +97,37 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
+                $scheme = 'pgsql';
+                break;
+            case 'pdo_sqlite':
+                $scheme = 'sqlite';
+                break;
+            default:
+                throw new \RuntimeException('Unsupported database driver: ' . $container->getParameter('database_driver'));
+        }
+
+        $container->setParameter('database_scheme', $scheme);
```
yguedidi, line 122: needed because using directly the driver value as a scheme doesn't work

app/AppKernel.php
```diff
@@ -97,5 +97,37 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
+
+        if ('sqlite' === $scheme) {
+            $container->setParameter('database_name', $container->getParameter('database_path'));
+        }
```
yguedidi, line 126: added in order to use only the database_name in the url

app/AppKernel.php
```diff
@@ -97,5 +97,37 @@ public function registerContainerConfiguration(LoaderInterface $loader)
...
+
+        $container->setParameter('database_user', (string) $container->getParameter('database_user'));
+        $container->setParameter('database_password', (string) $container->getParameter('database_password'));
+        $container->setParameter('database_port', (string) $container->getParameter('database_port'));
+        $container->setParameter('database_socket', (string) $container->getParameter('database_socket'));
```
yguedidi, line 131: needed because it doesn't work to concatenate a null value in parameters

app/config/config.yml
```diff
@@ -8,6 +8,7 @@ parameters:
     # Allows to use the live reload feature for changes in assets
     use_webpack_dev_server: false
     craue_config.cache_adapter.class: Craue\ConfigBundle\CacheAdapter\SymfonyCacheComponentAdapter
+    env(DATABASE_URL): '%database_scheme%://%database_user%:%database_password%@%database_host%:%database_port%/%database_name%?unix_socket=%database_socket%&charset=%database_charset%'
```
yguedidi, line 11: where the magic happens, in a backward compatible way :)

j0k3r (approved):

nicosomb (approved):

PR #7163, opened 2024-01-07
Run tests with PHPUnit directly

yguedidi:
Benefits:
- allow to use IDEs PHPUnit integrations
- it's kind of standard to run PHPUnit binary directly in PHP world
- no more tests only exposed Make recipess


tests/bootstrap.php
```diff
@@ -0,0 +1,48 @@
...
+
+(new Filesystem())->remove(__DIR__ . '/../var/cache/test');
+
+(new Process([
+    'php',
+    __DIR__ . '/../bin/console',
+    'doctrine:database:drop',
+    '--force',
+    '--env=test',
+]))->run(function ($type, $buffer) {
```
yguedidi, line 16: used `run` for this one only as it's allowed to fail

j0k3r (commented):
While I think it's a prettier solution, I also think it won't improve DX because locally I often don't want to drop the database, recreate it, run migration every time I launch a single test. Even more when I'm not using SQLite as database.

Can't we use a mix of both solution and only put the fixture command in the bootstrap PHPUnit file?

yguedidi:
@j0k3r could you share with me how you run "a single test" locally in day to day wallabag development?
Me multiple times I had some failing tests because not "clean" test database, so I took the habit to always run `docker compose run --rm php make test` because it solves that.

For context I did some timing checks on my local using Stopwatch component, here the results:
```
sqlite
default/clear cache: 8.00 MiB - 101 ms
default/drop database: 8.00 MiB - 1828 ms
default/create database: 8.00 MiB - 254 ms
default/migrate database: 8.00 MiB - 5512 ms
default/load fixtures: 8.00 MiB - 529 ms
8224 ms

postgres
default/clear cache: 8.00 MiB - 85 ms
default/drop database: 8.00 MiB - 1853 ms
default/create database: 8.00 MiB - 719 ms
default/migrate database: 8.00 MiB - 3866 ms
default/load fixtures: 8.00 MiB - 565 ms
7088 ms
```

Either we accept that less that 10 seconds is an acceptable bootstrap time, either we find a solution to skip database drop+create (I'd keep migration in case the in progress work adds one).
A solution could be to introduce a `WALLABAG_TEST_SKIP_DATABASE_RESET` environment variable for that.
I'd make it an opt-in, so by default always reset, an dev can add that env var in a local `phpunit.xml`

What do you think? Me I'd tend to accept the less that 10 seconds bootstrapping :)

yguedidi:
@j0k3r by the way, the only documented way of running tests is with `make test` if I'm not mistaken. So the scope of this PR is to make running `bin/phpunit` equivalent to the previous `make test`.
Supporting other ways of running tests by calling PHPUnit with a folder, filename, or `--filter` is helpful but should be a dedicated work to think about best way to skip those commands in such situation, and that update the doc accordingly :)

j0k3r:
I almost never use `make test` locally.
I use `make fixtures && php bin/simple-phpunit -v --filter=....` to run a single for what I'm working on.

yguedidi:
@j0k3r any change to accept this PR as is with the less than 10sec extra time when running test, meanwhile I think about a clean solution for partial testing?

j0k3r (approved):

PR #7177, opened 2024-01-12
Skip database reset on partial test run

yguedidi:


yguedidi:
@j0k3r this should support your use case of partial test run ;)

tests/functions.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+function isPartialRun(): bool
+{
+    foreach ($_SERVER['argv'] as $arg) {
+        if (str_starts_with($arg, '--filter')) {
```
yguedidi, line 6: used `str_starts_with` to support the case of `--filter=...` instead of `--filter ...`

composer.json
```diff
@@ -200,7 +200,10 @@
     "autoload-dev": {
         "psr-4": {
             "Tests\\": "tests/"
-        }
+        },
+        "files": [
+            "tests/functions.php"
```
yguedidi, line 205: I preferred a separate file than having the function definition in the bootstrap file

tests/functions.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+function isPartialRun(): bool
```
j0k3r, line 3: Could you add a comment above to explain the goal of the file and the function?

j0k3r (commented):
Overall that's a good idea/workaround 👍🏼

tests/functions.php
yguedidi: sure!

yguedidi: done

j0k3r (approved):

PR #7151, opened 2024-01-01
Move to one bundle

yguedidi:
A prerequisite work to the move to Flex.

- [x] based on #7137

PR #7196, opened 2024-01-21
Improve install command

yguedidi:
This mainly create a dedicated tmp database for each install command test.
Should allow running the tests on more cases instead of skipping them for such an important action that is the installation.

depends on:
- [x] #7216
- [x] #7217
- [x] #7218
- [x] #7220

### DIFF
--- a/src/Command/InstallCommand.php
+++ b/src/Command/InstallCommand.php
@@ -198,13 +198,21 @@ class InstallCommand extends Command
     {
         $this->io->section('Step 2 of 4: Setting up database.');
 
+        $conn = $this->entityManager->getConnection();
+        $databasePlatform = $conn->isConnected() ? $conn->getDatabasePlatform() : null;
+
         // user want to reset everything? Don't care about what is already here
         if (true === $this->defaultInput->getOption('reset')) {
             $this->io->text('Dropping database, creating database and schema, clearing the cache');
 
+            $this->runCommand('doctrine:schema:drop', ['--force' => true, '--full-database' => true]);
+
+            if (!$databasePlatform instanceof PostgreSQLPlatform) {
+                $this->runCommand('doctrine:database:drop', ['--force' => true]);
+                $this->runCommand('doctrine:database:create');
+            }
+
             $this
-                ->runCommand('doctrine:database:drop', ['--force' => true])
-                ->runCommand('doctrine:database:create')
                 ->runCommand('doctrine:migrations:migrate', ['--no-interaction' => true])
                 ->runCommand('cache:clear')
             ;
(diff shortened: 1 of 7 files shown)


src/Wallabag/CoreBundle/Command/InstallCommand.php
```diff
@@ -415,14 +421,47 @@ private function isDatabasePresent()
...
+                }
+
+                try {
+                    if (isset($associationMapping['joinTable']['name'])) {
+                        $schemaManager->dropTable($databasePlatform->quoteIdentifier($associationMapping['joinTable']['name']));
+                    }
+                } catch (TableNotFoundException $exception) {
+                }
+            }
+        }
```
yguedidi, line 465: all this is needed because the Doctrine metadata are not in sync with the schema created by migrations...
run `bin/console doctrine:schema:validate` to see it broken.
I plan to sync metadata and schema but this will require a dedicated work that will need discussion as it may involve big migrations (run `bin/console doctrine:schema:validate -v` or `bin/console doctrine:migrations:diff` to see the inconsistencies)
for example using PostgreSQL the command `doctrine:schema:drop` doesn't remove 2 tables

src/Wallabag/CoreBundle/Command/InstallCommand.php
```diff
@@ -415,14 +421,47 @@ private function isDatabasePresent()
...
+
+        $connection = $this->entityManager->getConnection();
+        $databasePlatform = $connection->getDatabasePlatform();
         $schemaManager = $this->entityManager->getConnection()->createSchemaManager();
 
-        return \count($schemaManager->listTableNames()) > 0 ? true : false;
+        try {
+            $schemaManager->dropTable($databasePlatform->quoteIdentifier($this->tableMetadataStorageConfiguration->getTableName()));
+        } catch (TableNotFoundException $exception) {
+        }
```
yguedidi, line 444: I discovered that I could empty the migration table by running `doctrine:migrations:version --all --delete` so I will change this, so that injecting the Doctrine service is not needed anymore.
also, will rebase on top of #7151 so that it's ready to be merged after :)

yguedidi: as I use the Doctrine migration table name to check for the existence of the schema, I need the service anyway, so I'll keep it as is

src/Command/InstallCommand.php
```diff
@@ -198,13 +193,21 @@ private function setupDatabase()
...
+        $conn = $this->entityManager->getConnection();
+        $databasePlatform = $conn->isConnected() ? $conn->getDatabasePlatform() : null;
+
         // user want to reset everything? Don't care about what is already here
         if (true === $this->defaultInput->getOption('reset')) {
             $this->io->text('Dropping database, creating database and schema, clearing the cache');
 
+            $this->runCommand('doctrine:schema:drop', ['--force' => true, '--full-database' => true]);
+
+            if (!$databasePlatform instanceof PostgreSQLPlatform) {
```
j0k3r, line 205: Why?

yguedidi: because PostgreSQL always create the database, and you can't delete the database you're connected too

j0k3r (commented):
Looks ok to me then

j0k3r (approved):

PR #7218, opened 2024-01-27
Fix doctrine:schema:drop

yguedidi:
Because of custom names of those foreign key constraints, they were not dropped by doctrine:schema:drop leading to some tables not being dropped because of them still existing.

schema is still not valid and will be fixed in another PR, for now we just fix what is needed to drop the full wallabag schema


yguedidi:
thanks for the review @j0k3r ! I think in the end better completely fix the schema validation, so I'm closing this in favor of #7220

PR #7242, opened 2024-02-05
Secure all actions with IsGranted

yguedidi:


yguedidi:
closing as will do it controller by controller

PR #7247, opened 2024-02-08
Make DownloadImages use HttpClient

yguedidi:
first step toward removing Guzzle

### DIFF
--- a/tests/Wallabag/CoreBundle/Helper/DownloadImagesTest.php
+++ b/tests/Wallabag/CoreBundle/Helper/DownloadImagesTest.php
@@ -2,11 +2,11 @@
 
 namespace Tests\Wallabag\CoreBundle\Helper;
 
-use GuzzleHttp\Psr7\Response;
-use Http\Mock\Client as HttpMockClient;
 use Monolog\Handler\TestHandler;
 use Monolog\Logger;
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\HttpClient\MockHttpClient;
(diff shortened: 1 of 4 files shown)


tests/Wallabag/CoreBundle/Helper/DownloadImagesTest.php
```diff
@@ -240,44 +236,27 @@ public function testEnsureOnlyFirstOccurrenceIsReplaced()
...
-        $httpMockClient->addResponse(new Response(200, ['content-type' => 'image/png'], file_get_contents(__DIR__ . '/../fixtures/unnamed.png')));
-
-        $logHandler = new TestHandler();
-        $logger = new Logger('test', [$logHandler]);
-
-        $download = new DownloadImages($httpMockClient, sys_get_temp_dir() . '/wallabag_test', 'http://wallabag.io/', $logger);
-        $res = $download->processSingleImage(123, '', 'https://example.com/unnamed.png');
-
-        $this->assertStringContainsString('/assets/images/9/b/9b0ead26/66953334.png', $res, "Fetch client didn't follow the HTTP redirection");
-    }
```
yguedidi, line 282: removed because [HttpClient always follow redirection](https://symfony.com/doc/5.x/http_client.html#redirects), and so this test become like testing the framework.
also, it can't work this way as if I force the first response to be a redirect, it will be the one DOwnloadImages service will receive and so will fail.

Kdecherf (approved):

PR #7248, opened 2024-02-08
Make PocketImport use HttpClient

yguedidi:
second step toward removing Guzzle

- [x] based on #7247
- [x] based on #7265

### DIFF
--- a/src/Wallabag/CoreBundle/Import/PocketImport.php
+++ b/src/Wallabag/CoreBundle/Import/PocketImport.php
@@ -2,22 +2,16 @@
 
 namespace Wallabag\CoreBundle\Import;
 
-use Http\Client\Common\HttpMethodsClient;
-use Http\Client\Common\Plugin\ErrorPlugin;
-use Http\Client\Common\PluginClient;
-use Http\Client\Exception\RequestException;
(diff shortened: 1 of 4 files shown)


yguedidi:
@j0k3r @nicosomb @Kdecherf I don't understand why those tests fail, it seems they fail on master too on my local, will investigate

src/Wallabag/CoreBundle/Import/PocketImport.php
```diff
@@ -81,18 +77,20 @@ public function getRequestToken($redirectUri)
...
+            $this->accessToken = $response->toArray()['access_token'];
+        } catch (ExceptionInterface $e) {
             $this->logger->error(sprintf('PocketImport: Failed to authorize client: %s', $e->getMessage()), ['exception' => $e]);
 
             return false;
         }
 
-        $this->accessToken = $this->jsonDecode($response)['access_token'];
-
         return true;
```
j0k3r, line 94: Why don't you also move that line inside the `try`?

yguedidi: no particular reason, makes sense so done now

j0k3r (approved):

PR #7265, opened 2024-02-18
Make retries an instance property in AuthenticatorSubscriber

yguedidi:
- [x] remove the first empty commit before merge!!! ⚠️ 

see the first commit that is empty has [failing tests](https://github.com/wallabag/wallabag/actions/runs/7951883836) on master..


j0k3r:
This is weird. How does it started failing now?

j0k3r (approved):

yguedidi:
> This is weird. How does it started failing now?

@j0k3r no idea to be honest.. must be something external to the code, so either an endpoint that is used in tests triggers the retry, and because it's a static property the value increase. but didn't investigate to confirm

PR #7268, opened 2024-02-18
Move templates

yguedidi:
- [x] based on #7265

A small and simple move toward Flex struture that simplify a bit the code.

places to update found with a simple but efficient search and replace of `@WallabagCore/` to nothing


app/config/config.yml
```diff
@@ -43,6 +43,7 @@ framework:
 
 # Twig Configuration
 twig:
+    default_path: '%kernel.project_dir%/templates'
```
yguedidi, line 46: here where the magic happen, from https://github.com/symfony/recipes/blob/main/symfony/twig-bundle/5.4/config/packages/twig.yaml

j0k3r (approved):

PR #7274, opened 2024-02-19
Move to bundle-less

yguedidi:
Another move toward Flex struture that simplify a lot the code structure.

Better reviewed commit by commit.
Fixup commits are there to highlight some changes that would have been hard to catch in the big change commits.
They will be squashed before merge as they don't make sense on there own and the commit they fix is a broken one without them.

Mostly search and replace too.

- [x] squash fixup commits before merge


Kdecherf (approved):
LGTM

PR #7290, opened 2024-02-24
Use lighter PDF file in Pinboard tests

yguedidi:
From [4m04s](https://github.com/wallabag/wallabag/actions/runs/8018668664) on latest master CI run to now [3m14s](https://github.com/wallabag/wallabag/actions/runs/8027642340): ~ -50s


Kdecherf (commented):
Could we commit the file to the repository in order to avoid making network calls during tests?

yguedidi:
> Could we commit the file to the repository in order to avoid making network calls during tests?

@Kdecherf  Maybe possible, but then it will not anymore look like to be an import from a Pinboard file that contains URLs.
Also, we already do many network walls during the tests that uses real URL
So more a work for later I'd say

Kdecherf (approved):

PR #7360, opened 2024-03-20
Use IsGranted in UserController

yguedidi:
extracted from #7242, added tests and updated on latest master

### DIFF
--- a/src/Controller/UserController.php
+++ b/src/Controller/UserController.php
@@ -10,6 +10,7 @@ use Pagerfanta\Doctrine\ORM\QueryAdapter as DoctrineORMAdapter;
 use Pagerfanta\Exception\OutOfRangeCurrentPageException;
 use Pagerfanta\Pagerfanta;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Google\GoogleAuthenticatorInterface;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Form\Form;
 use Symfony\Component\Form\FormInterface;
@@ -41,6 +42,7 @@ class UserController extends AbstractController
      * Creates a new User entity.
      *
      * @Route("/users/new", name="user_new", methods={"GET", "POST"})
+     * @IsGranted("CREATE_USERS")
      */
     public function newAction(Request $request, UserManagerInterface $userManager, EventDispatcherInterface $eventDispatcher)
     {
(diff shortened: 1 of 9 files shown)


templates/User/edit.html.twig
```diff
@@ -66,11 +66,17 @@
...
+                                        <button onclick="return confirm('{{ 'user.form.delete_confirm'|trans|escape('js') }}')" type="submit" class="btn waves-effect waves-light red">{{ 'user.form.delete'|trans }}</button>
+                                    {{ form_end(delete_form) }}
+                                {% else %}
+                                    <button disabled="disabled" type="button" class="btn waves-effect waves-light red">{{ 'user.form.delete'|trans }}</button>
+                                {% endif %}
                             </p>
-                            <p><a class="waves-effect waves-light btn blue-grey" href="{{ path('user_index') }}">{{ 'user.form.back_to_list'|trans }}</a></p>
+                            {% if is_granted('LIST_USERS') %}
+                                <p><a class="waves-effect waves-light btn blue-grey" href="{{ path('user_index') }}">{{ 'user.form.back_to_list'|trans }}</a></p>
+                            {% endif %}
```
yguedidi, line 79: @j0k3r @nicosomb @Kdecherf need your opinion on this: should I strictly wrap every link everywhere like now? or should we assume that if you got access to the edit template it's likely that you have access to the list already, and so no need to check?
The 2 approach are valid IMO, at work we use the later for less `is_granted` in case some are complex to compute, but I feel the former is "more secure" in case of code copy/paste or future usages that will make the assumption wrong

j0k3r: I think we can assume that if the user can edit a user, it has access to the list as well.

src/Controller/UserController.php
```diff
@@ -77,6 +79,7 @@ public function newAction(Request $request, UserManagerInterface $userManager, E
      * Displays a form to edit an existing User entity.
      *
      * @Route("/users/{id}/edit", name="user_edit", methods={"GET", "POST"})
+     * @IsGranted("EDIT", subject="user")
```
j0k3r, line 82: The `user` `subject` means it has to use the `UserVoter`?

yguedidi: kind of yes! it will use the `UserVoter` because it will be the one supporting a `User` entity as subject

templates/User/edit.html.twig
yguedidi: @nicosomb @Kdecherf any other opinion? meanwhile I implemented @j0k3r what wants :)

Kdecherf: Seems reasonable

yguedidi: I just removed the is_granted for LIST_USERS on new and edit pages

j0k3r (approved):

PR #7375, opened 2024-03-23
Use IsGranted in EntryController

yguedidi:
Better reviewed commit by commit.
Left as a draft PR so that it's not accidentaly merged before squashing in one commit after approve :)

For EntryVoter, all attributes are using the same rule: being the user of the entry.
But because it's about security, I prefered to be as granular as possible, in case each actions grant rules change in the future.
That's why I added an attribute per action, and updated the templates accordingly.

Same for MainVoter, that will own general actions not on a particular subject.
Just used the same LIST_ENTRIES for all entries listing actions as each is just a specific filtering.


yguedidi:
Thanks @j0k3r! squashing to a single commit before merging

j0k3r (approved):

PR #7376, opened 2024-03-23
Use IsGranted in SiteCredentialController

yguedidi:
- [x] based on #7375, review only last commit

### DIFF
--- a/src/Controller/SiteCredentialController.php
+++ b/src/Controller/SiteCredentialController.php
@@ -4,6 +4,7 @@ namespace Wallabag\Controller;
 
 use Craue\ConfigBundle\Util\Config;
 use Doctrine\ORM\EntityManagerInterface;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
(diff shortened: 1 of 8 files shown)


j0k3r (commented):
Looks ok as well

yguedidi:
@j0k3r sorry for asking review again, seems the workflow keeping the PR in draft so that it's not merged accidentally before rebasing/squashing is not perfect.. I guess you simply can't approve when it's in draft which is expected, but this leads to spamming you, sorry.
should I keep that workflow?

j0k3r:
> should I keep that workflow?

Maybe you can just open a PR and put the _Work in Progress_ label instead of using draft PR?
So if someone approuve the PR, you'll still be able to squash it without losing the approval.

j0k3r (approved):

yguedidi:
> Maybe you can just open a PR and put the Work in Progress label instead of using draft PR?

@j0k3r this will not protect from an accidental click on merge, but look like to be the best compromise, so OK!

PR #7577, opened 2024-07-12
Fix tests

j0k3r:
Looks like lemonde.fr doesn't work anymore in wallabag for test. Switching to an other great media.

Also update site configs.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4567,16 +4567,16 @@
         },
         {
             "name": "j0k3r/graby-site-config",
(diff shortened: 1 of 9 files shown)


nicosomb (dismissed):

nicosomb:
some tests are 🔴

Kdecherf (changes requested):
We should keep 2 entries in the import tests (I guess that's why they're failing btw)

Kdecherf (approved):

PR #7664, opened 2024-09-06
Add `graby-site-config` as direct deps

j0k3r:
So it'll get updated regularly by dependabot.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,6 +81,7 @@
         "html2text/html2text": "^4.3.1",
         "incenteev/composer-parameter-handler": "^2.2",
         "j0k3r/graby": "^2.4.5",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -81,6 +81,7 @@
         "html2text/html2text": "^4.3.1",
         "incenteev/composer-parameter-handler": "^2.2",
         "j0k3r/graby": "^2.4.5",
+        "j0k3r/graby-site-config": "^1.0",
```
yguedidi, line 84: just curious, why not dev-master?

yguedidi (approved):

composer.json
j0k3r: I don't know, just to follow the monthly release instead of having a dependabot PR every time the master change

PR #7701, opened 2024-09-27
fix: inappropriate handling of escaped query arguments

### LINKED ISSUE #6928: Tagging multiple entries through the API
### Environment

* **Version**: 2.6.5
* **Installation**: docker image
* **PHP version**: Not sure
* **OS**: Debian Stable
* **Database**: SQLite
* **Parameters**: This is my docker-compose.yml

<details>
  <summary>My <code>docker-compose.yml</code> is:</summary>
```yaml
version: '3'
services:
  wallabag:
    image: wallabag/wallabag
    container_name: wallabag
    environment:
      - SYMFONY__ENV__DATABASE_NAME=wallabag
      - SYMFONY__ENV__DATABASE_USER=wallabag
      - SYMFONY__ENV__SECRET=HIDDEN
      - SYMFONY__ENV__DOMAIN_NAME="HIDDEN"
      - SYMFONY__ENV__SERVER_NAME="HIDDEN"
    ports:
      - "8012:80"
    volumes:
      - ./data:/var/www/wallabag/data
      - ./images:/var/www/wallabag/web/assets/images
    healthcheck:
      test: ["CMD", "wget" ,"--no-verbose", "--tries=1", "--spider", "http://localhost"]
      interval: 100m
      timeout: 3s
```
</details>

I'm not sure if this is a bug or if I'm just misunderstanding the API docs.  Here is the documentation for the API endpoint `/api/entries/tags/lists`:

```
Handles an entries list and add tags to them.

list: Urls (as an array) to handle. A JSON array of urls [{'url': 'http://...','tags': 'tag1, tag2'}, {'url': 'http://...','tags': 'tag1, tag2'}]
```

It's not clear to me what this means.  I took this to mean I could specify multiple existing entries and supply their urls along with the tags to add to this API endpoint.

To test this, I picked an existing entry with an `id` of 9940 and the `url` field given below in the curl command.  I want to add two tags ("later" and "testtag") to this entry.  (NOTE: I'm aware that I can do this for a single entry via the endpoint `/api/entries/{entry}/tags`, I would like to apply tags to multiple entries at once.)

From a local machine, I then run
```bash
curl -X POST \
         -H "Content-Type: application/json"
         --data '[{"url":"https://www.escapistmagazine.com/why-transparency-is-such-a-big-issue-for-the-strike-and-for-streaming/","tags":"later,testtag"}]' \
         "$WALLABAG_HOST/api/entries/tags/lists?access_token=$WALLABAG_TOKEN"
```
where `WALLABAG_HOST` and `WALLABAG_TOKEN` have their obvious meanings.

The result is:
```json
{
  "type": "https://tools.ietf.org/html/rfc2616#section-10",
  "title": "An error occurred",
  "status": 500,
  "detail": "Internal Server Error"
}
```

Am I misunderstanding how this works?  I tried asking on Gitter yesterday but there were no responses.

Other questions about the tagging API:

1. Why is the DELETE version of this endpoint `/api/entries/tags/list` (not `lists`, like above)?  Is this a typo in the docs?
2. I can add multiple tags to a specific entry via `/api/entries/{entry}/tags`.  But there is no equivalent to removing multiple tags from a specific entry: `DELETE /api/entries/{entry}/tags/{tag}` removes only a single tag.  Is there a symmetric version of the add tags feature?

Oppen:
The list(s) endpoints receive a JSON encoded list of URLs as a query
param. Because it's a query param, it's escaped, so the `json_decode`
method can't decode it without first calling `urldecode`.
Also, the `POST` versions were accessing the elments as if they were the
strings, rather than objects of the shape `{'url':'<url>'}`.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | yes/no
| Deprecations? | yes/no
| Tests pass?   | yes/no
| Documentation | yes/no
| Translation   | yes/no
| CHANGELOG.md  | yes/no
| License       | MIT

Fixes #6928
Tested locally on my instance. I hope the CI will catch any pieces lost in translation :)

### DIFF
--- a/src/Controller/Api/EntryRestController.php
+++ b/src/Controller/Api/EntryRestController.php
@@ -1289,7 +1289,7 @@ class EntryRestController extends WallabagRestController
     {
         $this->validateAuthentication();
 
-        $urls = json_decode(urldecode($request->query->get('list', '[]')));
+        $list = json_decode(urldecode($request->query->get('list', '[]')));
 
         if (empty($list)) {
             return $this->sendResponse([]);
@@ -1356,7 +1356,7 @@ class EntryRestController extends WallabagRestController
     {
         $this->validateAuthentication();
 
-        $urls = json_decode(urldecode($request->query->get('list', '[]')));
+        $list = json_decode(urldecode($request->query->get('list', '[]')));
 
         if (empty($list)) {
             return $this->sendResponse([]);
--- a/src/Controller/Api/EntryRestController.php
+++ b/src/Controller/Api/EntryRestController.php
@@ -477,7 +477,7 @@ class EntryRestController extends WallabagRestController
     {
         $this->validateAuthentication();
 
-        $urls = json_decode($request->query->get('urls', []));
+        $urls = json_decode(urldecode($request->query->get('urls', '[]')));
 
         if (empty($urls)) {
             return $this->sendResponse([]);
@@ -487,6 +487,7 @@ class EntryRestController extends WallabagRestController
 
         // handle multiple urls
         foreach ($urls as $key => $url) {
+            $url = $url->url;
             $entry = $entryRepository->findByUrlAndUserId(
                 $url,
                 $this->getUser()->getId()
@@ -537,7 +538,7 @@ class EntryRestController extends WallabagRestController
     {
         $this->validateAuthentication();
 
-        $urls = json_decode($request->query->get('urls', []));
+        $urls = json_decode(urldecode($request->query->get('urls', '[]')));
 
         $limit = $this->getParameter('wallabag.api_limit_mass_actions');
 
@@ -552,6 +553,7 @@ class EntryRestController extends WallabagRestController
 
         // handle multiple urls
         foreach ($urls as $key => $url) {
+            $url = $url->url;
             $entry = $entryRepository->findByUrlAndUserId(
                 $url,
                 $this->getUser()->getId()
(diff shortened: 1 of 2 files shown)


src/Controller/Api/EntryRestController.php
```diff
@@ -1287,7 +1289,7 @@ public function deleteEntriesTagsListAction(Request $request, TagRepository $tag
     {
         $this->validateAuthentication();
 
-        $list = json_decode($request->query->get('list', []));
+        $urls = json_decode(urldecode($request->query->get('list', '[]')));
```
Oppen, line 1292:
```suggestion
        $list = json_decode(urldecode($request->query->get('list', '[]')));
```

src/Controller/Api/EntryRestController.php
```diff
@@ -1354,7 +1356,7 @@ public function postEntriesTagsListAction(Request $request, EntryRepository $ent
...
```
Oppen, line 1359:
```suggestion
        $list = json_decode(urldecode($request->query->get('list', '[]')));
```

j0k3r:
I've checked it locally and in fact, it's just the OpenAPI example which is wrong.
It's not `[{'url': 'http://...'}, {'url': 'http://...'}]` but it must be `['http://...', 'http://...']` instead.

The array must be an array of urls.
Try it on your side with the proper array and you'll see it'll works as expected.

All in all, you spotted a problem in the API document. Can you convert bad array to the proper one?
Thanks

yguedidi:
@Oppen @j0k3r do we agree that this, "as is", should be closed, and a PR fixing the API documentation should replace it?

j0k3r:
@yguedidi yep!

PR #7753, opened 2024-10-31
Replaced gitter with matrix

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  |no
| BC breaks?    |no
| Deprecations? |no
| Tests pass?   | yes
| Documentation |no
| Translation   | yes
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/translations/messages.cs.yml
+++ b/translations/messages.cs.yml
@@ -343,7 +343,7 @@ tag:
 quickstart:
     support:
         title: Podpora
-        gitter: Na Gitteru
+        matrix: Na Gitteru
         email: E-mailem
         github: Na GitHubu
         description: Pokud potřebujete pomoc, jsme tu pro vás.
--- a/translations/messages.cs.yml
+++ b/translations/messages.cs.yml
@@ -343,7 +343,6 @@ tag:
 quickstart:
     support:
         title: Podpora
-        matrix: Na Gitteru
         email: E-mailem
         github: Na GitHubu
         description: Pokud potřebujete pomoc, jsme tu pro vás.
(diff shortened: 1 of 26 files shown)


translations/messages.pl.yml
```diff
@@ -474,7 +474,7 @@ quickstart:
         description: Jeżeli potrzebujesz pomocy, jesteśmy tutaj dla ciebie.
         github: na GitHubie
         email: przez e-mail
-        gitter: na Gitterze
+        matrix: na Gitterze
```
Kdecherf, line 477: I'm not sure we should keep the translation in the languages were there is a doubt on the update to make. I would go with letting a missing translation entry in this situation

Kdecherf (changes requested):

translations/messages.pl.yml
nicosomb: you'd prefer to remove these lines?

Kdecherf: Yes I would prefer

nicosomb: Done.

translations/messages.cs.yml
```diff
@@ -343,7 +343,7 @@ tag:
 quickstart:
     support:
         title: Podpora
-        gitter: Na Gitteru
+        matrix: Na Gitteru
```
Kdecherf, line 346: Remove

translations/messages.fa.yml
```diff
@@ -270,7 +270,7 @@ quickstart:
         description: به کمک نیاز دارید؟ ما پشتیبان شما هستیم.
         github: روی گیت‌هاب
         email: با ایمیل
-        gitter: روی گیتر
+        matrix: روی گیتر
```
Kdecherf, line 273: Delete

Kdecherf (changes requested):

Issue #7762, opened 2024-11-01
Import page is broken on master branch


nicosomb:
### Environment

* **Version**: master branch
* **Installation**: git clone 
* **PHP version**: 8.1
* **OS**: macos
* **Database**: sqlite
* **Parameters**:

<details>
  <summary>My <code>app/config/parameters.yml</code> is:</summary>

  ```
# This file is auto-generated during the composer install
parameters:
    database_driver: pdo_sqlite
    database_host: 127.0.0.1
    database_port: null
    database_name: wallabag
    database_user: root
    database_password: null
    database_path: '%kernel.project_dir%/data/db/wallabag.sqlite'
    database_table_prefix: wallabag_
    database_socket: null
    database_charset: utf8mb4
    domain_name: 'http://wllbg-dev.test'
    server_name: 'Your wallabag instance'
    mailer_dsn: 'smtp://127.0.0.1'
    locale: en
    secret: CHANGE_ME_TO_SOMETHING_SECRET_AND_RANDOM
    twofactor_sender: no-reply@wallabag.org
    fosuser_registration: true
    fosuser_confirmation: true
    fos_oauth_server_access_token_lifetime: 3600
    fos_oauth_server_refresh_token_lifetime: 1209600
    from_email: no-reply@wallabag.org
    rss_limit: 50
    rabbitmq_host: localhost
    rabbitmq_port: 5672
    rabbitmq_user: guest
    rabbitmq_password: guest
    rabbitmq_prefetch_count: 10
    redis_scheme: tcp
    redis_host: localhost
    redis_port: 6379
    redis_path: null
    redis_password: null
    sentry_dsn: null

  ```
</details>

### What steps will reproduce the bug?

When I'm on `http://localhost/import/`, I have this error: `stream_socket_client(): Unable to connect to tcp://localhost:5672 (Connection refused)`

@yguedidi I don't know if it's related to the great work you made to change the structure of the project.

PR #7137, opened 2023-12-24
Upgrade to Symfony 5

yguedidi:
requires those PRs to be merged first:
- [x] #7135
- [x] #7136
- [x] #7140
- [x] release 2.6.8
- [x] merge `2.6` in `master`, #7152 
- [x] rebase

better be reviewed commit by commit

| Production Changes                   | From    | To      | Compare                                                                               |
|--------------------------------------|---------|---------|---------------------------------------------------------------------------------------|
| doctrine/annotations                 | 1.14.3  | 2.0.1   | [...](https://github.com/doctrine/annotations/compare/1.14.3...2.0.1)                 |
| doctrine/doctrine-bundle             | 2.7.2   | 2.11.1  | [...](https://github.com/doctrine/DoctrineBundle/compare/2.7.2...2.11.1)              |
| doctrine/doctrine-migrations-bundle  | 3.2.5   | 3.3.0   | [...](https://github.com/doctrine/DoctrineMigrationsBundle/compare/3.2.5...3.3.0)     |
| enshrined/svg-sanitize               | 0.15.4  | 0.16.0  | [...](https://github.com/darylldoyle/svg-sanitizer/compare/0.15.4...0.16.0)           |
| friendsofsymfony/rest-bundle         | 3.5.0   | 3.6.0   | [...](https://github.com/FriendsOfSymfony/FOSRestBundle/compare/3.5.0...3.6.0)        |
| incenteev/composer-parameter-handler | v2.1.5  | v2.2.0  | [...](https://github.com/Incenteev/ParameterHandler/compare/v2.1.5...v2.2.0)          |
| jms/serializer-bundle                | 5.3.1   | 5.4.0   | [...](https://github.com/schmittjoh/JMSSerializerBundle/compare/5.3.1...5.4.0)        |
| lcobucci/jwt                         | 4.1.5   | 4.3.0   | [...](https://github.com/lcobucci/jwt/compare/4.1.5...4.3.0)                          |
| lexik/form-filter-bundle             | v7.0.3  | REMOVED |                                                                                       |
| nelmio/api-doc-bundle                | v4.11.1 | v4.18.1 | [...](https://github.com/nelmio/NelmioApiDocBundle/compare/v4.11.1...v4.18.1)         |
| nelmio/cors-bundle                   | 2.3.1   | 2.4.0   | [...](https://github.com/nelmio/NelmioCorsBundle/compare/2.3.1...2.4.0)               |
[...]

PR #7768, opened 2024-11-03
Restored lazy connections for RabbitMQ

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    |no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | yes/no
| License       | MIT

Should fix https://github.com/wallabag/wallabag/issues/7762

The bug was introduced here: https://github.com/wallabag/wallabag/commit/423f2f792d57386defce3369b8e780f1aac0b428 in this huge PR: https://github.com/wallabag/wallabag/pull/7137

RabbitMqBundle doesn't support lazy services, but in the documentation, it's written: `but you can set lazy: true in your connection configuration to avoid unnecessary connections to your message broker in every request`
I think that this setting is necessary for us, when the RabbitMQ server is not up. 

Our testsuite didn't warn us, because our testsuite environment installs a RabbitMQ server.
That's why I duplicated the workflow (and I removed RabbitMQ and Redis dependencies), but I don't know if it's the good way.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -148,3 +148,57 @@ jobs:
 
       - name: "Run PHPUnit"
         run: "php bin/phpunit -v"
+
+  phpunit-without-rmq-redis:
+    name: "PHP ${{ matrix.php }} using ${{ matrix.database }} without Rabbit & Redis"
+    runs-on: "ubuntu-20.04"
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php:
+          - "8.3"
+        database:
+          - "sqlite"
+          - "mysql"
+          - "pgsql"
+
+    steps:
+      - name: "Checkout"
+        uses: "actions/checkout@v4"
+        with:
+          fetch-depth: 2
+
+      - name: "Install PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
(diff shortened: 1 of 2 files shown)


j0k3r:
> Our testsuite didn't warn us, because our testsuite environment installs a RabbitMQ server.
> That's why I duplicated the workflow (and I removed RabbitMQ and Redis dependencies), but I don't know if it's the good way.

Maybe you can just add a new job in the current CI for only one version of PHP where you don't install rabbit & redis. We don't need that part to be tested on all PHP versions.

.github/workflows/continuous-integration.yml
```diff
@@ -148,3 +148,56 @@ jobs:
 
       - name: "Run PHPUnit"
         run: "php bin/phpunit -v"
+phpunit-without-rmq-redis:
```
j0k3r, line 151:
```suggestion

  phpunit-without-rmq-redis:
```

j0k3r (changes requested):

j0k3r (approved):

j0k3r (approved):

yguedidi (approved):
🤞🏼

PR #7797, opened 2024-11-11
Change NB_ELEMENTS in pocket importer to 30 to comply with Pocket API

### LINKED ISSUE #7635: Pocket imports are not working at all
### Environment
* **Version**: 2.6.9
* **Installation**: Docker
* **PHP version**: 8.1
* **OS**: Debian 12
* **Database**: SQLite / MySQL
* **Parameters**:  Not relevant

### What is the bug?
Wallabag fails to import Pocket links.

When trying to import using the Pocket API, the import fails randomly. Sometimes it imports nothing, sometimes it imports 100-200 links. After that it fails with a cryptic error message such as:

An exception occurred while executing a query: SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xF0\x9F\x91\x87</...' for column `wallabag`.`wallabag_entry`.`content` at row 1

Even though the imported links do not contain any UTF-8 codepoints or smilies.

Converting the exported file to JSON and then importing that using the Wallabagv2 JSON importer also fails for similar reasons.
 
- Install Wallabag in a Docker container
- Try to import Pocket links using the API
- It fails

If you need anymore information please let me know.

ruffinge:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes #7635

Change NB_ELEMENTS in pocket importer to 30 to comply with Pocket API restrictions.

Tested on my instance and verified that import from pocket now works correctly.

### DIFF
--- a/src/Import/PocketImport.php
+++ b/src/Import/PocketImport.php
@@ -9,7 +9,7 @@ use Wallabag\Entity\Entry;
 
 class PocketImport extends AbstractImport
 {
-    public const NB_ELEMENTS = 5000;
+    public const NB_ELEMENTS = 30;
     /**
      * @var HttpClientInterface
      */


j0k3r:
Can give us the mention about the Pocket API?

ruffinge:
Based on that wording, it would seem that it should work with more, but in my testing, the response from pocket's api was HTTP 413, with no data. That empty data was what caused the error in wallabag during any import attempt with more entries (`500: Internal Server Error Wallabag\ImportBundle\Import\AbstractImport::parseEntriesForProducer(): Argument #1 ($entries) must be of type array, null given, called in /var/www/wallabag.it/app2019/src/Wallabag/ImportBundle/Import/PocketImport.php on line 134`).

lcd1232:
@j0k3r could we merge this fix?

j0k3r:
@ruffinge Could you target the 2.6 branch instead of master?

j0k3r (approved):

PR #7799, opened 2024-11-12
Add Markdown export

### LINKED ISSUE #4243: Export to Markdown
Wallabag can export to various formats like PDF, EPub and others. Could be added a new option to export in markdown format? For taking notes and create brief documents, MD is usually preferred than ODT and HTML. It's a light format, there are a ton of apps compatible with it, and you can edit them in anything that can edit plain text.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    |no
| Deprecations? | no
| Tests pass?   | yes
| Documentation |no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Should fix #4243

### DIFF
--- a/migrations/Version20241112193044.php
+++ b/migrations/Version20241112193044.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Wallabag\Doctrine\WallabagMigration;
+
+/**
+ * Added the internal setting to export articles in markdown.
+ */
+final class Version20241112193044 extends WallabagMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('INSERT INTO ' . $this->getTable('internal_setting') . " (name, value, section) VALUES ('export_md', '1', 'export');");
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('DELETE FROM' . $this->getTable('internal_setting') . " WHERE name = 'export_md';");
+    }
+}
--- a/src/Helper/EntriesExport.php
+++ b/src/Helper/EntriesExport.php
@@ -5,6 +5,7 @@ namespace Wallabag\Helper;
 use Html2Text\Html2Text;
 use JMS\Serializer\SerializationContext;
 use JMS\Serializer\SerializerBuilder;
+use League\HTMLToMarkdown\HtmlConverter;
 use PHPePub\Core\EPub;
 use PHPePub\Core\Structure\OPF\DublinCore;
 use Symfony\Component\HttpFoundation\Response;
@@ -129,10 +130,8 @@ class EntriesExport
 
     /**
      * Use PHPePub to dump a .epub file.
-     *
-     * @return Response
      */
-    private function produceEpub()
+    private function produceEpub(): Response
     {
         $user = $this->tokenStorage->getToken() ? $this->tokenStorage->getToken()->getUser() : null;
         \assert($user instanceof User);
@@ -249,10 +248,8 @@ class EntriesExport
 
     /**
      * Use TCPDF to dump a .pdf file.
-     *
-     * @return Response
      */
-    private function producePdf()
+    private function producePdf(): Response
     {
         $user = $this->tokenStorage->getToken() ? $this->tokenStorage->getToken()->getUser() : null;
         \assert($user instanceof User);
@@ -326,10 +323,8 @@ class EntriesExport
 
     /**
      * Inspired from CsvFileDumper.
-     *
-     * @return Response
      */
-    private function produceCsv()
+    private function produceCsv(): Response
     {
         $delimiter = ';';
         $enclosure = '"';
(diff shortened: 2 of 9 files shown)


src/Helper/EntriesExport.php
```diff
@@ -432,6 +433,34 @@ private function produceTxt()
...
     }
 
+    /**
+     * Dump a Markdown file.
+     *
+     * @return Response
+     */
+    private function produceMd()
+    {
+        $user = $this->tokenStorage->getToken() ? $this->tokenStorage->getToken()->getUser() : null;
```
yguedidi, line 443: This seems unused after

src/Helper/EntriesExport.php
```diff
@@ -432,6 +433,34 @@ private function produceTxt()
...
+        \assert($user instanceof User);
+
+        $content = '';
+        $converter = new HtmlConverter();
+        $converter->getConfig()->setOption('strip_tags', true);
+        foreach ($this->entries as $entry) {
```
yguedidi, line 449: Does it make sense to support multiple entries for markdown export? I guess users will expect multiple md files, not all in a single one

templates/Entry/entries.html.twig
```diff
@@ -105,6 +105,7 @@
...
             {% if craue_setting('export_txt') %}<li class="bold"><a class="waves-effect" href="{{ path('export_entries', {'category': current_route, 'format': 'txt', 'tag': current_tag, 'search_entry[term]': export_search_term, 'currentRoute': previous_route, 'entry': entry}) }}">TXT</a></li>{% endif %}
             {% if craue_setting('export_xml') %}<li class="bold"><a class="waves-effect" href="{{ path('export_entries', {'category': current_route, 'format': 'xml', 'tag': current_tag, 'search_entry[term]': export_search_term, 'currentRoute': previous_route, 'entry': entry}) }}">XML</a></li>{% endif %}
+            {% if craue_setting('export_md') %}<li class="bold"><a class="waves-effect" href="{{ path('export_entries', {'category': current_route, 'format': 'md', 'tag': current_tag, 'search_entry[term]': export_search_term, 'currentRoute': previous_route, 'entry': entry}) }}">MD</a></li>{% endif %}
```
yguedidi, line 108: Is MD clear enough to end users?

tests/Controller/ExportControllerTest.php
```diff
@@ -307,6 +307,23 @@ public function testXmlExport()
         $this->assertNotEmpty('updated_at', (string) $content->entry[0]->updated_at);
     }
 
+    public function testMdExport()
```
yguedidi, line 310: I'd add an assertion on the content

yguedidi (changes requested):

yguedidi:
Reviewed from phone 😅

src/Helper/EntriesExport.php
nicosomb: Why not, but we need to rewrite the Export system. That's how we handle other exports.

tests/Controller/ExportControllerTest.php
```diff
@@ -307,6 +307,25 @@ public function testXmlExport()
...
+    {
+        $this->logInAs('admin');
+        $client = $this->getTestClient();
+
+        ob_start();
+        $crawler = $client->request('GET', '/export/all.md');
+        ob_end_clean();
```
yguedidi, line 317: I don't get why ob_* functions and a $crawler variable are needed here, I think they can be removed, just keep the call to request

yguedidi (changes requested):

tests/Controller/ExportControllerTest.php
```diff
@@ -307,6 +307,23 @@ public function testXmlExport()
         $this->assertNotEmpty('updated_at', (string) $content->entry[0]->updated_at);
     }
 
+    public function testMdExport()
+    {
+        $this->logInAs('admin');
+        $client = $this->getTestClient();
+
+        $crawler = $client->request('GET', '/export/all.md');
```
yguedidi, line 315: I would not store the crawler in a variable as it's unused

yguedidi (dismissed):
One last comment, approved anyway as not critical so you can merge after the change

yguedidi (dismissed):

migrations/Version20241112193044.php
```diff
@@ -0,0 +1,24 @@
...
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Wallabag\Doctrine\WallabagMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
```
j0k3r, line 11: Can you instead explain what it does?

src/Helper/EntriesExport.php
```diff
@@ -432,6 +433,31 @@ private function produceTxt()
         );
     }
 
+    /**
+     * Dump a Markdown file.
+     *
+     * @return Response
+     */
+    private function produceMd()
```
j0k3r, line 441: Can't we apply typing for new method?
```suggestion
     */
    private function produceMd(): Response
```

j0k3r (approved):

PR #7806, opened 2024-11-14
Add assets build during install and update

### LINKED ISSUE #7788: Assets bug after migration from 2.6 to 2.7
In order to plan the 2.7 release, I tried the migration (from 2.6.10 to master). 
All was fine. 

An issue: the webfonts (see https://github.com/wallabag/wallabag/pull/6773) weren't available. 
I had to execute `yarn build:prod` to fix that.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes/no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fix #7788

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -39,3 +39,5 @@ TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
 git checkout $TAG
 SYMFONY_ENV=$ENV $COMPOSER_COMMAND install --no-dev -o --prefer-dist
 php bin/console wallabag:install --env=$ENV
+yarn install
+yarn build:$ENV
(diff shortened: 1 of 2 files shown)


nicosomb:
@Kdecherf we provide an archive for those who don't want to build assets (no need to have composer for example)

these updated files are available for those who want to install wallabag manually (with composer, and yarn, etc.)

yguedidi (approved):

BrandonNolet:
How will this affect those with docker installs?

yguedidi:
@BrandonNolet Docker is based on the ZIP file (see [here](https://github.com/wallabag/docker/blob/master/Dockerfile#L69)) and it has the assets already, we build them on release (see [here](https://github.com/wallabag/wallabag/blob/master/scripts/release.sh#L13-L14))

yguedidi:
> To me here the underlying decision is more: is git a recommended way to install wallabag for end users? or is it only the dev repo?
> Those scripts are git-based, so if we support the git install method, then I guess we should update those scripts and make sure the doc is clear about requirements (node and yarn).
> Personal opinion: I don't think git is a way to install an app :)

@nicosomb @j0k3r @Kdecherf any opinion?

PR #7816, opened 2024-11-16
fix(TaggingRule): allowed variable should match docs

rosenstrauch:
add missing 'r' here to allow creating tagging rules using the documented isStarred filter.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes/no
| Documentation | yes/no
| Translation   | no
| CHANGELOG.md  | yes/no
| License       | MIT


nicosomb (approved):
good catch!

yguedidi:
@nicosomb shouldn't this target 2.6 instead? Now that it's merged it must be backported I think

PR #7831, opened 2024-11-19
Fix #7746 API Document for Entry 

### LINKED ISSUE #7746: PATCH API endpoint can't edit/unstar articles
### Environment

* **Version**: 2.6.9
* **Installation**: Docker

According to [the API docs](https://app.wallabag.it/api/doc/), the correct way to programmatically unstar an entry is with PATCH:

```bash
curl -X 'PATCH' \
      "$server/api/entries/$entry_id?starred=0&is_starred=0" -H 'accept: */*' \
      -H "Authorization: Bearer $token"
```

But when I run the above command, I get a entry response that isn't updated, as if I had issued a GET command:

```text
curl -X 'PATCH' \
      "$server/api/entries/$entry_id?starred=0&is_starred=0" -H 'accept: */*' \
      -H "Authorization: Bearer $token" | jq '.is_starred'

  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

1
```

And the wallabag logs show it is receiving a PATCH command, so it's not a failing curl invocation:

```text
172.31.0.4 - - [30/Oct/2024:05:22:32 +0000] "GET /api/entries/2819 HTTP/1.0" 200 10260 "-" "curl/8.10.1"
172.31.0.4 - - [30/Oct/2024:05:22:32 +0000] "PATCH /api/entries/2819?starred=0 HTTP/1.0" 499 0 "-" "curl/8.10.1"
```

EINDEX:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Tests pass?   | yes
| Documentation | yes
| Translation   |no
| CHANGELOG.md  | no
| License       | MIT

Fixes #7746 

This PR just fixing the API document payload to correct JSON body from params.

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -558,124 +558,81 @@ class EntryRestController extends WallabagRestController
      * @Operation(
      *     tags={"Entries"},
      *     summary="Create an entry.",
-     *     @OA\Parameter(
-     *         name="url",
-     *         in="query",
-     *         description="Url for the entry.",
+     *     @OA\RequestBody(
      *         required=true,
-     *         @OA\Schema(
-     *             type="string",
-     *             example="http://www.test.com/article.html"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="title",
-     *         in="query",
-     *         description="Optional, we'll get the title from the page.",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="string",
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="tags",
-     *         in="query",
-     *         description="a comma-separated list of tags.",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="string",
-     *             example="tag1,tag2,tag3"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="archive",
-     *         in="query",
-     *         description="entry already archived",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="integer",
-     *             enum={"1", "0"},
-     *             default="0"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="starred",
-     *         in="query",
-     *         description="entry already starred",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="integer",
-     *             enum={"1", "0"},
-     *             default="0"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="content",
-     *         in="query",
-     *         description="Content of the entry",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="string"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="language",
-     *         in="query",
-     *         description="Language of the entry",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="string"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="preview_picture",
-     *         in="query",
-     *         description="Preview picture of the entry",
-     *         required=false,
-     *         @OA\Schema(
-     *             type="string"
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="published_at",
-     *         in="query",
-     *         description="Published date of the entry",
-     *         required=false,
-     *
-     *         @OA\Schema(
-     *             type="string",
-     *             format="YYYY-MM-DDTHH:II:SS+TZ or a timestamp (integer)",
-     *         )
-     *     ),
-     *     @OA\Parameter(
-     *         name="authors",
-     *         in="query",
-     *         description="Authors of the entry",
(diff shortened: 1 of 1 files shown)


src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -558,124 +558,81 @@ public function postEntriesListAction(Request $request, EntryRepository $entryRe
      * @Operation(
      *     tags={"Entries"},
      *     summary="Create an entry.",
-     *     @OA\Parameter(
-     *         name="url",
-     *         in="query",
```
yguedidi, line 563: I think the right fix is to remove this `in="query"` as suggested by @j0k3r in https://github.com/wallabag/wallabag/issues/7746#issuecomment-2454063764
the endpoints are not taking a JSON but normal POST

yguedidi (changes requested):

src/Wallabag/ApiBundle/Controller/EntryRestController.php
EINDEX: Hi @yguedidi 

I tried that, but both are not working, did you have any other suggestions?

```xml
<args>
<item key="0">string</item>
<item key="1">
<![CDATA[ Missing required field "in" for @OA\Parameter(name="url") in \Wallabag\ApiBundle\Controller\EntryRestController->postEntriesAction() in /Users/eindex/repos/wallabag/src/Wallabag/ApiBundle/Controller/EntryRestController.php ]]>
</item>
</args>
```

image

```xml
<args>
<item key="0">string</item>
<item key="1">
<![CDATA[ @OA\Parameter(name="url",in="body")->in "body" is invalid, expecting "query", "header", "path", "cookie" in \Wallabag\ApiBundle\Controller\EntryRestController->postEntriesAction() in /Users/eindex/repos/wallabag/src/Wallabag/ApiBundle/Controller/EntryRestController.php ]]>
</item>
</args>
```

image

yguedidi: not an expert on OpenAPI, but quickly checking the doc and it seems we should replace `@OA\Parameter` by [`@OA\RequestBody`](https://zircote.github.io/swagger-php/reference/annotations.html#requestbody)

yguedidi (changes requested):

EINDEX:
Hi @yguedidi 

That document says:
> Describes a single request body.

image


So this means that does not support multi-fields, like what we are using.

The reason I changed to JSON as payload, wallabag supports JSON payload. 
But I am not sure about another payload format, like form-url-encoded, or multipart.

yguedidi:
@EINDEX looking at `\Wallabag\Controller\Api\EntryRestController::retrieveValueFromRequest()` it clearly uses `$request->request->get(`, that is the equivalent for `$_POST`, so for normal post submissions, not JSON ones 😕

EINDEX:
I understand your point, the other API read JSON is using code like this: https://github.com/wallabag/wallabag/blob/89db5690a0df40348f5ff13a95f230dedb8f2e57/src/Wallabag/AnnotationBundle/Controller/WallabagAnnotationController.php#L64C9-L64C59

However I have tested the API, and on the behavior side,  https://github.com/wallabag/wallabag/issues/7746 can prove the API accepts JSON.
And it also accepts the `Content-Type: application/x-www-form-urlencoded`

Both below ways work.
```bash
curl -X "PATCH" "https://<wallabag>/api/entries/1136" \
     -H 'Authorization: Bearer <token>' \
     -H 'Content-Type: application/x-www-form-urlencoded; charset=utf-8' \
     --data-urlencode "archive=1"
```

```bash
curl -X "PATCH" "https://<wallabag>/api/entries/1136" \
     -H 'Authorization: Bearer <token>' \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "archive": "1"
}'
```

I think the current gap, from `swagger-php` to the actual library `http_foundation` does not match, about the payload part in the document of `swagger-php` does not show any way to add `Content-Type: application/x-www-form-urlencoded`.

I am new to PHP, this confuses me.
Maybe you can test, it and give me some suggestions?

Thank you very much.

yguedidi:
Maybe @j0k3r @nicosomb @Kdecherf know better than be about API documentation?

j0k3r:
I finally took time to test it locally and it's handle pretty well.
Data is well retrieve when we send a json in the patch request. Maybe at some point the request is converted so that `$request->request->get(` can retrieve data from the body json.

Anyway, it looks ok to me.

I can see one more method that will require to be converted from `Parameter` to `RequestBody`, it's `postEntriesTagsAction`. Tags list must be given as json instead of in the query.
Maybe `postEntriesTagsListAction` too.

j0k3r:
@EINDEX can you the two remaining method to `RequestBody`? Also, can you run `php bin/php-cs-fixer fix` locally to fix CS issues?

Issue #7848, opened 2024-11-22
RSS feeds now link to wallabag instead of actual article URL


anarcat:
i use wallabag with [livemarks](https://addons.mozilla.org/en-US/firefox/addon/livemarks/) to have wallabag bookmarks in my web browser, it's pretty neat!

it used to work pretty well too, i could add a link to wallabag with a certain tag and poof, it would show up in my browser after the next RSS pull. neato!

but at some point, this stopped working: now links, instead of going to (say) `example.com`, they go to `wallabag.example.com/view/1243`. then i need to click *again* to get to the real site.

### Environment

* **Version**: 2.5.4
* **Installation**: ?
* **PHP version**: unknown, i don't run the server, but i assume it's whatever debian stable ships
* **OS**: Debian, probably stable
* **Database**: probably mysql
* **Parameters**: no access

### What steps will reproduce the bug?

  1. add a URL to wallabag
  2. add RSS feed to livemarks
  3. look at the feed

It looks like this is actually a regression that was added in the RSS to Atom switch, in this massive commit (https://github.com/wallabag/wallabag/commit/531c8d0a5c55fa93438e227a7d349235fbd31d28#diff-a073911454063f5c33d4a63105ce63ca3c8b22ff5068737163950a290cf951b1R35)

The `<item>` element used to look like:

```
            <item>
                <title><![CDATA[{{ entry.title|e }}]]></title>
                <source url="{{ url('view', { 'id': entry.id }) }}">wallabag</source>
                <link>{{ entry.url }}</link>
                <guid>{{ entry.url }}</guid>
```

now it looks like this:

```
    <entry>
        <title><![CDATA[{{ entry.title|e }}]]></title>
        <link href="{{ entry.url }}"/>
        <link rel="alternate" type="text/html"
              href="{{ url('view', {'id': entry.id}) }}"/>
        <link rel="via"
              href="{{ entry.url }}"/>
```

I'm not sure why livemarks is using the "alternate" entry instead of the normal entry, that said. 

I think the correct way to do this would be to put the wallabag URL in the `via` field instead.

Note that I first filed this as an issue in Livemarks, they claim they are following standards and are just trying to comply with Feedburner, see https://github.com/nt1m/livemarks/issues/486.

PR #7849, opened 2024-11-22
restore pre-Atom behavior of linking directly to the article

anarcat:
RFC4287 section 4.2.7.2 specifies that "rel=alternate" is effectively the default for the link element:

   If the "rel" attribute is not present, the link
   element MUST be interpreted as if the link relation type is
   "alternate".

So having a plain `<link>` and a `<link rel="alternate">` is kind of weird, *especially* if they point to different resources. So we just remove the plain entry and *replace* it with the rel=alternate, which is really the default here.

The sample Atom feeds in RFC4287 (section 1.1) do give an example *only* with `rel="alternate"`:

     <entry>
       <title>Atom draft-07 snapshot</title>
       <link rel="alternate" type="text/html"
        href="http://example.org/2005/04/02/atom"/>
       <link rel="enclosure" type="audio/mpeg" length="1337"
        href="http://example.org/audio/ph34r_my_podcast.mp3"/>

To refer to the actual Wallabag URL, we use the "via", which is defined in the RFC as:

   5.  The value "via" signifies that the IRI in the value of the href attribute identifies a resource that is the source of the information provided in the containing element.

I'm not sure how widely used that tag is, but I feel that the distinction between `rel="alternate"` is weird at best, and buggy (and certainly introducing unpleasantness in my usage) at worse.

Before:

```xml
<link href="{{ entry.url }}"/>
<link rel="alternate" type="text/html"
      href="{{ url('view', {'id': entry.id}) }}"/>
<link rel="via"
      href="{{ entry.url }}"/>
```

That is:

```xml
<link href="http://example.com/"/>
<link rel="alternate" type="text/html"
      href="http://wallabag.example.com/view/1"/>
<link rel="via"
      href="http://example.com/"/>
```

After:

```xml
<link rel="alternate" href="{{ entry.url }}"/>
<link rel="via" type="text/html"
      href="{{ url('view', {'id': entry.id}) }}"/>
```

That is:

```xml
<link rel="alternate" href="http://example.com"/>
<link rel="via" type="text/html"
      href="http://wallabag.example.com/view/1"/>
```
Closes: #7848

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | ???
| Deprecations? | no?
| Tests pass?   | ?
| Documentation | no
| Translation   | no
| CHANGELOG.md  | TODO?
| License       | MIT

### DIFF
--- a/templates/Entry/entries.xml.twig
+++ b/templates/Entry/entries.xml.twig
@@ -31,11 +31,9 @@
     {% for entry in entries %}
     <entry>
         <title><![CDATA[{{ entry.title|e }}]]></title>
-        <link href="{{ entry.url }}"/>
-        <link rel="alternate" type="text/html"
+        <link rel="alternate" href="{{ entry.url }}"/>
+        <link rel="via" type="text/html"
               href="{{ url('view', {'id': entry.id}) }}"/>
-        <link rel="via"
-              href="{{ entry.url }}"/>
         <id>wallabag:{{ domainName|removeScheme|removeWww }}:{{ user }}:entry:{{ entry.id }}</id>
         <updated>{{ entry.updatedAt|date('c') }}</updated>
         <published>{{ entry.createdAt|date('c') }}</published>


j0k3r (approved):
Looks ok, thanks for the explanation.

PR #7887, opened 2024-12-10
Define a custom user-agent for wallabag

e-adrien:
Hello,

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This PR is related to #7878
The idea is to use a user-agent that describe the Wallabag instance so it's easier for websites to identify and allow the requests.
`Wallabag/%wallabag.version% (Guzzle/5 +%domain_name%)`

I wanted to fetch the Guzzle version from the composer lockfile but I don't know if we can do that.

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -31,6 +31,7 @@ services:
             $supportUrl: '@=service(''craue_config'').get(''wallabag_support_url'')'
             $fonts: '%wallabag.fonts%'
             $defaultIgnoreOriginInstanceRules: '%wallabag.default_ignore_origin_instance_rules%'
+            $defaultUserAgent: "%wallabag.user_agent%"
 
     Wallabag\:
         resource: '../../src/*'
@@ -206,6 +207,8 @@ services:
             $config:
                 error_message: '%wallabag.fetching_error_message%'
                 error_message_title: '%wallabag.fetching_error_message_title%'
+                http_client:
+                    ua_browser: "%wallabag.user_agent%"
             $client: '@psr18.wallabag.client'
         calls:
             - [ setLogger, [ "@logger" ] ]
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -31,7 +31,7 @@ services:
             $supportUrl: '@=service(''craue_config'').get(''wallabag_support_url'')'
             $fonts: '%wallabag.fonts%'
             $defaultIgnoreOriginInstanceRules: '%wallabag.default_ignore_origin_instance_rules%'
-            $defaultUserAgent: "%wallabag.user_agent%"
+            $defaultUserAgent: "%wallabag_user_agent%"
 
     Wallabag\:
         resource: '../../src/*'
@@ -208,7 +208,7 @@ services:
                 error_message: '%wallabag.fetching_error_message%'
                 error_message_title: '%wallabag.fetching_error_message_title%'
                 http_client:
-                    ua_browser: "%wallabag.user_agent%"
+                    ua_browser: "%wallabag_user_agent%"
             $client: '@psr18.wallabag.client'
         calls:
             - [ setLogger, [ "@logger" ] ]
--- a/app/config/wallabag.yml
+++ b/app/config/wallabag.yml
@@ -183,4 +183,3 @@ parameters:
         - 'text/html'
         - 'application/vnd.ms-excel'
     wallabag.resource_dir: "%kernel.project_dir%/web/uploads/import"
-    wallabag.user_agent: "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/535.2 (KHTML, like Gecko) Chrome/15.0.874.92 Safari/535.2"
--- a/app/config/wallabag.yml
+++ b/app/config/wallabag.yml
@@ -183,3 +183,4 @@ parameters:
         - 'text/html'
         - 'application/vnd.ms-excel'
     wallabag.resource_dir: "%kernel.project_dir%/web/uploads/import"
+    wallabag.user_agent: "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/535.2 (KHTML, like Gecko) Chrome/15.0.874.92 Safari/535.2"
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -28,6 +28,8 @@ parameters:
 
     mailer_dsn: smtp://127.0.0.1
 
+    wallabag_user_agent: "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/535.2 (KHTML, like Gecko) Chrome/15.0.874.92 Safari/535.2"
+
     locale: en
 
     # A secret key that's used to generate certain security-related tokens
--- a/src/SiteConfig/LoginFormAuthenticator.php
+++ b/src/SiteConfig/LoginFormAuthenticator.php
@@ -14,6 +14,7 @@ class LoginFormAuthenticator
     public function __construct(
         private readonly HttpBrowser $browser,
         AuthenticatorProvider $authenticatorProvider,
+        private readonly string $defaultUserAgent,
     ) {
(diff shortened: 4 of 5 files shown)


src/Helper/HttpClientFactory.php
```diff
@@ -64,6 +75,14 @@ public function createClient(array $config = [])
...
         }
 
+        if (!isset($config['defaults']['headers'])) {
+            $config['defaults']['headers'] = [];
+        }
+        if (!isset($config['defaults']['headers']['User-Agent'])) {
+            // need to set the user-agent
+            $config['defaults']['headers']['User-Agent'] = $this->userAgent;
+        }
+
```
j0k3r, line 85: Maybe this should be moved before line 67 so if admin didn't enabled paywalled website, they'll still got the default user agent.

Kdecherf (changes requested):
We should probably avoid putting the version of wallabag as it could expose sensitive information in case of vulnerabilities

app/config/services.yml
```diff
@@ -191,6 +191,7 @@ services:
     Graby\Graby:
         arguments:
             $config:
+                http_client: { "ua_browser": "Wallabag/%wallabag.version% (Guzzle/5 +%domain_name%)" }
```
j0k3r, line 194: I think that's a good idea because "webmaster" will be able to filter request from wallabag but I'm wondering how this is going to behave more globally because most websites might block request because they didn't know that _browser_.

The default UA is more likely to be accepted because it looks like a browser.

I'm wondering if this shouldn't be defined for each site which can create a dedicated rule to allow wallabag to request their content.

e-adrien:
> We should probably avoid putting the version of wallabag as it could expose sensitive information in case of vulnerabilities
> 
> Edit: we also should reconsider the use of `domain_name` here as it could disclose private instances

I was trying to replicate Mastodon or Pleroma UA strings:
```
Pleroma 2.6.2; https://pleroma.chtisurel.net <matlag@chtisurel.net>; Bot
Pleroma 2.7.0-299-gbc223acf; https://social.linuxine.net <>; Bot
Mastodon/4.3.2 (http.rb/5.2.0; +https://toot.renn.es/)
Mastodon/4.3.0 (http.rb/5.2.0; +https://plumes.io/)
```
but a less specific string will be fine too.

Should I go for `Wallabag/2 (Guzzle/5)` and keep only the major or just `Wallabag (Guzzle)`?

yguedidi:
For info I'm in the process of migrating from Guzzle to Symfony HttpClient. So the UA will certainly change later :)

src/Helper/HttpClientFactory.php
e-adrien: It's done

app/config/services.yml
e-adrien: Agreed, it can go both ways.

We are using Cloudflare and it sends JS challenges to Wallabag requests with the current default UA.
If someone didn't send me a post on Mastodon about this issue, I wouldn't have noticed that Wallabag traffic was blocked.

I think we can also argue that it's more fair to declare Wallabag instead of trying to hide it behind a browser.

app/config/services.yml
```diff
@@ -191,6 +191,7 @@ services:
     Graby\Graby:
         arguments:
             $config:
+                http_client: { "ua_browser": "Wallabag (Guzzle/5)" }
```
yguedidi, line 194:
```suggestion
                http_client:
                    ua_browser: "%wallabag.user_agent%"
```
better use a parameter IMO

app/config/services.yml
```diff
@@ -218,6 +219,7 @@ services:
 
     Wallabag\Helper\HttpClientFactory:
         calls:
+            - ['setUserAgent', ["Wallabag (Guzzle/5)"]]
```
yguedidi, line 222: having the user agent as a parameter, you could inject the parameter directly to the HttpClientFactory service, no need for the setter

yguedidi (changes requested):
I'd use the existing user agent as a default value for the parameter. users will be able to change it if they want.
maybe deserves some documentation

e-adrien:
Hi @yguedidi, Thank you for your time. I made the requested changes. 
Do you want me to open a PR on https://github.com/wallabag/doc ?

yguedidi:
@e-adrien thanks for the work and sorry for the delay!
For 2.7.0 we want #7947 that is likely to be merged soon as already approved once (will just require a rebase to fix composer.lock conflicts)
And it will remove the HttpClientFactory you update here.. Maybe worth investigating how you could introduce this new feature on top of #7947 ?

yguedidi:
@e-adrien you can rebase this and adapt! 🎉

e-adrien:
Hi @yguedidi, it's done if you want to have a look.
The only thing is that this PR changes the same line of code than #7920 but, I will rebase this PR or #7920 later.

j0k3r:
I still have mixed feeling about changing the default UA for all requests from wallabag.
wallabag isn't a browser and some websites might have different behavior about that. And I think wallabag should hide it behing a browser UA to ensure the maximum compatibility with websites.

FMPOV it's far from a _simple_ change.

Thoughts @yguedidi @Kdecherf @nicosomb?

yguedidi:
I agree with @j0k3r ! to me goal here would be more to add a feature to allow power users to change the UA if they want, but the default value for `wallabag.user_agent` should be the original one

HolgerAusB:
> Can we have an option in the website configuration (for example `use_browser_ua`, default value to true) to allow the website to choose between a browser UA and an explicit Wallabag UA?

would be nice to set this as a free string via UI globaly with a reset button to roll-back to 'Wallabag (FooBar/5)'
and maybe in the credentials area an override UA per domain option, while not necessary to have username/password then.

j0k3r:
> to me goal here would be more to add a feature to allow power users to change the UA if they want

@yguedidi for the whole instance or only for a given user (for its scope only)?

> Can we have an option in the website configuration (for example `use_browser_ua`, default value to true) to allow the website to choose between a browser UA and an explicit Wallabag UA?

@e-adrien you mean, in the site credentials page of wallabag?

> and maybe in the credentials area an override UA per domain option

@HolgerAusB for me, this is something that should be defined in the siteconfig, like it's already the case

yguedidi:
> for the whole instance or only for a given user (for its scope only)?

@j0k3r to me for the whole instance. I see it this way:

- wallabag has its default existing UA
- power users can change that value if they want through parameter (env var in a near future)
- site configs can customize it if needed to override the UA value from wallabag, be it the default one or the power users customized one

e-adrien:
Hi @yguedidi 
I finally found some time to rebase this PR.
I think it respects all of your three points (I took the user-agent from Graby's source code). Let me know if you want me to change anything.

nicosomb:
@e-adrien Is that possible to write some lines in our documentation https://github.com/wallabag/doc/ to explain how to use this new configuration?

app/config/wallabag.yml
```diff
@@ -183,3 +183,4 @@ parameters:
         - 'text/html'
         - 'application/vnd.ms-excel'
     wallabag.resource_dir: "%kernel.project_dir%/web/uploads/import"
+    wallabag.user_agent: "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/535.2 (KHTML, like Gecko) Chrome/15.0.874.92 Safari/535.2"
```
j0k3r, line 186: Can you define that value in `app/config/parameters.yml.dist` so that admin user will be able to change it without touching the code.

e-adrien: It's done, let me know if you prefer a different name for the parameter.

e-adrien:
@nicosomb I've just opened a PR for the documentation : https://github.com/wallabag/doc/pull/204

nicosomb (approved):

PR #7900, opened 2024-12-14
Create .dockerignore

RJGhugo:
The Docker Image seems to have a lot of unnecessary files.

Please add files to be ignored.


ccoenen:
This is not how pull requests are supposed to work.

I think this can be closed. The added file does not contribute anything. This should probably be an issue, not a pull-request.

Issue #7987, opened 2025-01-28
"With annotations" menu count is different from real number of entries


50LUD4:
Hello everyone.

Entries "With annotations" left menu count is different from real number of entries. I've a self-hosted instance.

Thanks

![Image](https://github.com/user-attachments/assets/b79ec7cd-5fab-4aca-abd2-2399df22d4ee)

j0k3r:
Good catch, dunno how can we have missed that.

PR #7999, opened 2025-02-10
Fix entries counter for annotated entries in the menu

j0k3r:
The query were badly made and return all annotations for the current user instead of the total of entries with annotation(s).

Fix https://github.com/wallabag/wallabag/issues/7987


nicosomb (approved):
👏

Kdecherf (approved):

PR #8030, opened 2025-02-23
Back-end hide/show of random, export and filters

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Better reviewed commit by commit

### DIFF
--- a/templates/Entry/entries.html.twig
+++ b/templates/Entry/entries.html.twig
@@ -114,7 +114,7 @@
 
     <!-- Export -->
     {% if has_exports and is_granted('EXPORT_ENTRIES') %}
-        <div id="export" class="sidenav">
+        <div id="export" class="sidenav" data-edge="right">
         {% set current_tag = null %}
         {% if tag is defined %}
             {% set current_tag = tag.slug %}
@@ -140,7 +140,7 @@
 
     <!-- Filters -->
     {% if has_filters %}
-    <div id="filters" class="sidenav">
+    <div id="filters" class="sidenav" data-edge="right">
         <form action="{{ path('all') }}">
 
             <h4 class="center">{{ 'entry.filters.title'|trans }}</h4>
--- a/templates/Entry/entries.html.twig
+++ b/templates/Entry/entries.html.twig
@@ -262,7 +262,7 @@
                 </div>
 
                 <div class="col s6">
-                    <a href="#!" class="center waves-effect waves-green btn-flat" id="clear_form_filters">{{ 'entry.filters.action.clear'|trans }}</a>
+                    <button type="reset" class="center waves-effect waves-green btn-flat">{{ 'entry.filters.action.clear'|trans }}</button>
                 </div>
 
                 <div class="col s6">
--- a/templates/Entry/entries.html.twig
+++ b/templates/Entry/entries.html.twig
@@ -1,5 +1,8 @@
 {% extends "layout.html.twig" %}
 
+{% set has_filters = form is not null %}
+{% set has_exports = craue_setting('export_epub') or craue_setting('export_pdf') or craue_setting('export_json') or craue_setting('export_csv') or craue_setting('export_txt') or craue_setting('export_xml') or craue_setting('export_md') %}
+
 {% block head %}
     {{ parent() }}
     {% if tag is defined and app.user.config.feedToken %}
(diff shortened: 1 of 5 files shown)


templates/Entry/entries.html.twig
```diff
@@ -1,5 +1,8 @@
 {% extends "layout.html.twig" %}
 
+{% set has_filters = form is not null %}
```
yguedidi, line 3: check on `is_granted('LIST_ENTRIES')` got removed because now all actions using this template are protected by the security check

templates/Entry/entries.html.twig
```diff
@@ -18,6 +21,28 @@
     {% include "Entry/_title.html.twig" with {'filter': filter} %}
 {% endblock %}
 
+{% block nav_panel_extra_actions %}
+    <li>
+        <a class="waves-effect tooltipped js-random-action" data-position="bottom" data-delay="50" data-tooltip="{{ 'menu.top.random_entry'|trans }}" href="{{ path('random_entry', {'type': active_route}) }}">
+            <i class="material-icons">casino</i>
+        </a>
+    </li>
```
yguedidi, line 29: wrapping check `active_route and is_granted('LIST_ENTRIES')` removed as now the button is only on the entries page and that condition is always true there

templates/Entry/entries.html.twig
yguedidi: kept the `form is not null` condition because it's false on the `showEntriesForTagAction`

templates/Entry/entries.html.twig
```diff
@@ -86,7 +111,8 @@
     {% endif %}
 
     <!-- Export -->
-    <div id="export" class="sidenav">
+    {% if has_exports %}
```
yguedidi, line 114: this is a non critical bugfix as before the panel was always displayed even when no export optin enabled

templates/Entry/entries.html.twig
```diff
@@ -233,7 +260,7 @@
                 </div>
 
                 <div class="col s6">
-                    <a href="#!" class="center waves-effect waves-green btn-flat" id="clear_form_filters">{{ 'entry.filters.action.clear'|trans }}</a>
+                    <button type="reset" class="center waves-effect waves-green btn-flat">{{ 'entry.filters.action.clear'|trans }}</button>
```
yguedidi, line 263: again small non critical bugfix as the JS implementation was buggy, better rely on standards

templates/layout.html.twig
```diff
@@ -1,5 +1,23 @@
...
+   {% set active_route = 'annotated' %}
+{% elseif current_route == 'archive' or current_route_from_query_params == 'archive' %}
+   {% set active_route = 'archive' %}
+{% elseif current_route == 'starred' or current_route_from_query_params == 'starred' %}
+   {% set active_route = 'starred' %}
+{% elseif current_route == 'unread' or current_route == 'homepage' or current_route_from_query_params == 'unread' %}
+   {% set active_route = 'unread' %}
+{% elseif current_route == 'untagged' %}
+   {% set active_route = 'untagged' %}
+{% endif %}
```
yguedidi, line 19: moved here so that `active_route` is accessible to children templates

Kdecherf (changes requested):
A quick local test seems to show that `has_filters` and `has_exports` don't work. I don't see the export and filter anymore, however I expected to see at least the export button as I have at least one export enabled.

yguedidi:
> A quick local test seems to show that `has_filters` and `has_exports` don't work. I don't see the export and filter anymore, however I expected to see at least the export button as I have at least one export enabled.

@Kdecherf on which page are you? the buttons should appear only when on entries listing pages.
The previous JS code was showing the buttons only when the `#filters` or `#export` elements are on the page, and those are defined on `entries.html.twig`.
It is expected to not have the buttons on quickstart or config pages.

yguedidi:
@j0k3r @Kdecherf @nicosomb could you review this one please? :)

Kdecherf (approved):
@yguedidi looks good, it seems that I messed up with my local env

PR #8031, opened 2025-02-23
Use Stimulus

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Use Stimulus for JS behaviors.
Goal is to have them isolated, attached to the right elements, and properly clean any side effect, so that later we unblock the usage of Turbo Drive for smooth navigation.

Also, as each commit was simple and understandable enough, I took the opportunity to migrate away from jQuery, and managed to finally remove it 🎉 

- [x] Materialize
  - [x] Sidenav
  - [x] Tabs
  - [x] Tooltip
  - [x] Collapsible
  - [x] Dropdown
  - [x] Form Select
  - [x] FAB
  - [x] Toast
- [x] QR code
- [x] Highlight
- [x] Scroll indicator
- [x] Scroll Storage
- [x] Annotations
- [x] Clipboard
- [x] Config
- [x] Fake Radio
- [x] Tag
- [x] Batch Edit
- [x] Topbar
- [x] Sticky Nav
- [x] Leftbar
- [x] Add Tag
- [x] Dark Theme
- [x] Shortcuts
- [x] Entries Navigation


yguedidi:
@j0k3r @Kdecherf @nicosomb @Simounet any early feedback on the approach? :)

assets/controllers/qrcode_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from '@hotwired/stimulus';
+import jrQrcode from 'jr-qrcode';
+
+export default class extends Controller {
+  static values = { url: String }
```
j0k3r, line 5: So there is magic somewhere where `this.values.url` is being converted into `this.urlValue`?

yguedidi: Yes, stimulus "feature"

Simounet:
It looks like one more tool to understand and maintain for only 3 page types. For projects with not so many maintainers, I'm kind of very strict about dependencies. Webpack already made my head hurts a few times for not so many code and features to bundle.
The changes you're providing here are not that complex, but I know nothing about Stimulus. Hope this will stay that way. 🤞

yguedidi:
> It looks like one more tool to understand and maintain for only 3 page types. For projects with not so many maintainers, I'm kind of very strict about dependencies. Webpack already made my head hurts a few times for not so many code and features to bundle.
> The changes you're providing here are not that complex, but I know nothing about Stimulus. Hope this will stay that way. 🤞

Thanks @Simounet for your feedback! To be honest not the kind of feedback I expected but I respect your point of view!

Do you know about [Turbo Drive](https://turbo.hotwired.dev/)? in simple words it's improve the navigation UX on an app by avoiding full page reloads, giving a feeling of faster page loads, like with Single Page Apps.
I think improving navigation speed perception is a good goal for wallabag.

A prerequisite to setup Turbo is to have any JS side effect be loaded and unloaded properly when elements get added and removed on the page, to avoid memory leaks (which in itself is a good practice, even when not expecting to use Turbo).

[Stimulus](https://stimulus.hotwired.dev/) is a lightweight JS framework that help doing that in a simple way, by adding JS behavior attached to HTML elements, and it manages for us the loading and unloading of JS side effects.
It follows the "JS for progressive enhancements" philosophy I'd say, which is exactly why we have some JS in wallabag!

To me, using Stimulus will:

- allow using Turbo Drive for a better UX
- better architecture the front-end code by have small specific controllers
- allow to remove jQuery, as many things jQuery provided are now doable with vanilla JS, and Stimulus helps here in attaching JS behaviors to elements only
- open the possibility to use Symfony UX packages

In case you want to understand better the Stimulus philosophy, you can read the [The Origin of Stimulus](https://stimulus.hotwired.dev/handbook/origin), an interesting read 🙂 

Hope I managed to convince you

yguedidi:
@Simounet I just you you reacted with a thumbup, I'll take it as an OK so I'll continue my work on this then 🙂 
@j0k3r @Kdecherf @nicosomb any blockage from you?

Simounet:
Yep, looks good to me.

assets/index.js
```diff
@@ -274,13 +274,6 @@ const articleScroll = () => {
 $(document).ready(() => {
   $('select').formSelect();
 
-  $('.dropdown-trigger').dropdown({ hover: false });
```
yguedidi, line 286: not migrated as there are no such dropdowns

assets/controllers/materialize/form_select_controller.js
```diff
@@ -0,0 +1,12 @@
+import { Controller } from '@hotwired/stimulus';
+import M from '@materializecss/materialize';
+
+export default class extends Controller {
+  connect() {
+    this.instance = M.FormSelect.init(this.element.querySelector('select'));
```
yguedidi, line 6: Stimulus controller applied on a parent element and get the select element to pass it to Materialize because of the magic Materialize is doing with the select element, it is altered by Materialize and made Stimulus do an infinite loop when applied directly on the select element

templates/Entry/entry.html.twig
```diff
@@ -330,7 +330,7 @@
             {{ entry.content|raw }}
         </article>
 
-        <div class="fixed-action-btn js-fixed-action-btn horizontal click-to-toggle hide-on-large-only">
+        <div class="fixed-action-btn horizontal click-to-toggle hide-on-large-only" data-controller="materialize--fab" data-action="scroll@window->materialize--fab#autoDisplay">
```
yguedidi: registering the scroll action this way allows Stimulus handle attaching/detaching the event listener automatically

assets/index.js
```diff
@@ -255,25 +255,13 @@ const articleScroll = () => {
       const s = $(window).scrollTop();
       const d = $(document).height();
       const c = $(window).height();
-      const articleElBottom = articleEl.offset().top + articleEl.height();
       const scrollPercent = (s / (d - c)) * 100;
       $('.progress .determinate').css('width', `${scrollPercent}%`);
-      const fixedActionBtn = $('.js-fixed-action-btn');
-      const toggleScrollDataName = 'toggle-auto';
-      if ((s + c) > articleElBottom) {
-        fixedActionBtn.data(toggleScrollDataName, true);
```
yguedidi, line 264: no more need for a data attribute, managed directly by a controller property

assets/controllers/qrcode_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from '@hotwired/stimulus';
+import jrQrcode from 'jr-qrcode';
+
+export default class extends Controller {
+  static values = { url: String };
+
+  connect() {
```
yguedidi, line 7: nothing to disconnect here

assets/controllers/highlight_controller.js
```diff
@@ -0,0 +1,11 @@
+import { Controller } from '@hotwired/stimulus';
+import 'highlight.js/styles/atom-one-light.css';
+import hljs from 'highlight.js';
+
+export default class extends Controller {
+  connect() {
+    this.element.querySelectorAll('pre code').forEach((element) => {
```
yguedidi, line 7: changed from `pre` to `pre code` following HighlightJS documentation

assets/controllers/scroll_storage_controller.js
```diff
@@ -0,0 +1,19 @@
...
+      top: window.innerHeight * localStorage[`wallabag.article.${this.entryIdValue}.percent`],
+      behavior: 'smooth',
+    });
+  }
+
+  saveScroll() {
+    const scrollPercent = window.scrollY / window.innerHeight;
+    const scrollPercentRounded = Math.round(scrollPercent * 100) / 100;
+
+    localStorage[`wallabag.article.${this.entryIdValue}.percent`] = scrollPercentRounded;
```
yguedidi, line 17: assumed localStorage to always be available as widely supported now

assets/controllers/annotations_controller.js
```diff
@@ -0,0 +1,57 @@
...
+        destroy: this.destroyUrlValue,
+        search: this.searchUrlValue,
+      },
+      entryId: this.entryIdValue,
+      onError(msg, xhr) {
+        if (!Object.prototype.hasOwnProperty.call(xhr, 'responseJSON')) {
+          annotator.notification.banner('An error occurred', 'error');
+          return;
+        }
+        Object.values(xhr.responseJSON.children).forEach((v) => {
```
yguedidi, line 39: used `Object.values` to replace jQuery `$.each`

templates/Developer/client_parameters.html.twig
```diff
@@ -18,14 +18,14 @@
                         <td>{{ 'developer.client_parameter.field_id'|trans }}</td>
                         <td>
                             <strong><code>{{ client_id }}</code></strong>
-                            <button class="btn" data-clipboard-text="{{ client_id }}">{{ 'developer.client.copy_to_clipboard'|trans }}</button>
+                            <button class="btn" data-controller="clipboard" data-clipboard-text="{{ client_id }}">{{ 'developer.client.copy_to_clipboard'|trans }}</button>
```
yguedidi, line 21: kept the `data-clipboard-text` attribute not migrated to a Stimulus value because it is used directly by ClipboardJS, and not configurable

package.json
```diff
@@ -102,7 +102,7 @@
     "build:dev": "encore dev",
     "watch": "encore dev --watch",
     "build:prod": "encore production --progress",
-    "lint:js": "eslint assets/*.js assets/js/*.js assets/js/**/*.js assets/controllers/*.js",
```
yguedidi: no more JS files there so remove to keep ESLint running successfully

templates/Tag/tags.html.twig
```diff
@@ -19,18 +19,19 @@
...
+                    <a href="{{ path('tag_entries', {'slug': tag.slug}) }}" class="card-tag-link" data-tag-target="link">
                         {{ tag.label }}&nbsp;({{ nbEntries }})
                     </a>
                     {% if renameForms is defined and renameForms[tag.id] is defined and is_granted('EDIT', tag) %}
-                    <form class="card-tag-form hidden" data-handle="tag-rename-form" action="{{ path('tag_rename', {'slug': tag.slug, redirect: current_path}) }}" method="POST">
-                        {{ form_widget(renameForms[tag.id].label, {'attr': {'value': tag.label}}) }}
+                    <form class="card-tag-form hidden" data-tag-target="form" action="{{ path('tag_rename', {'slug': tag.slug, redirect: current_path}) }}" method="POST">
+                        {{ form_widget(renameForms[tag.id].label, {'attr': {'value': tag.label, 'data-tag-target': 'input'}}) }}
                         {{ form_rest(renameForms[tag.id]) }}
+                        <button type="submit"><i class="material-icons">done</i></button>
```
yguedidi, line 30: better make visible a real functional form with a submit button than some JS magic

assets/controllers/batch_edit_controller.js
```diff
@@ -0,0 +1,15 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+  static targets = ['item', 'tagAction'];
+
+  toggleSelection(e) {
+    this.itemTargets.forEach((item) => {
+      item.checked = e.currentTarget.checked; // eslint-disable-line no-param-reassign
```
yguedidi, line 8: add the ESLint ignore because apparently the current config forbid to edit parameters of a closure

templates/Entry/entries.html.twig
```diff
@@ -77,15 +77,15 @@
...
                 <button class="mass-action-button btn cyan darken-1" type="submit" name="delete" onclick="return confirm('{{ 'entry.confirm.delete_entries'|trans|escape('js') }}')" title="{{ 'entry.list.delete'|trans }}"><i class="material-icons">delete</i></button>
             </div>
 
             <div class="mass-action-tags">
-                <button class="btn cyan darken-1 mass-action-button mass-action-button--tags" type="submit" name="tag" title="{{ 'entry.list.add_tags'|trans }}"><i class="material-icons">label</i></button>
-                <input type="text" class="mass-action-tags-input" name="tags" placeholder="{{ 'entry.list.mass_action_tags_input_placeholder'|trans }}" />
+                <button class="btn cyan darken-1 mass-action-button mass-action-button--tags" type="submit" name="tag" title="{{ 'entry.list.add_tags'|trans }}" data-batch-edit-target="tagAction"><i class="material-icons">label</i></button>
+                <input type="text" class="mass-action-tags-input" name="tags" placeholder="{{ 'entry.list.mass_action_tags_input_placeholder'|trans }}" data-action="keydown.enter->batch-edit#tagSelection:prevent:stop" />
```
yguedidi, line 88: I found it better to have `:prevent:stop` here, equivalent of `e.preventDefault()` and `e.stopPropagation()` in JS

assets/controllers/topbar_controller.js
```diff
@@ -0,0 +1,31 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+  static targets = ['addUrl', 'addUrlInput', 'search', 'searchInput', 'actions'];
+
+  showAddUrl() {
+    this.actionsTarget.style.display = 'none';
+    this.addUrlTarget.style.display = 'flex';
```
yguedidi, line 8: they are actually flex containers, using the usual `block` breaks things

assets/controllers/materialize/fab_controller.js
```diff
@@ -22,6 +22,10 @@ export default class extends Controller {
     }
   }
 
+  click() {
+    this.dispatch('click');
```
yguedidi, line 26: this will dispatch a `materialize--fab:click` event that will bubble up to window ...

templates/Entry/entry.html.twig
```diff
@@ -320,7 +320,7 @@
             </div>
 
             {% if is_granted('TAG', entry) %}
-                <div class="input-field nav-panel-add-tag" style="display: none">
+                <div class="input-field nav-panel-add-tag hidden" data-controller="add-tag" data-action="leftbar:toggleAddTagForm@window->add-tag#toggle">
```
yguedidi, line 323: same trick here about dispatching an event and listening to it on window from another controller.
it's a way to interact between separated controllers

templates/base.html.twig
```diff
@@ -4,7 +4,7 @@
 <!--[if lte IE 7]><html class="no-js ie7 ie67 ie678"{% if lang is not empty %} lang="{{ lang }}"{% endif %}><![endif]-->
 <!--[if IE 8]><html class="no-js ie8 ie678"{% if lang is not empty %} lang="{{ lang }}"{% endif %}><![endif]-->
 <!--[if gt IE 8]><html class="no-js"{% if lang is not empty %} lang="{{ lang }}"{% endif %}><![endif]-->
-<html{% if lang is not empty %} class="{{ theme_class() }}" lang="{{ lang }}"{% endif %}>
+<html{% if lang is not empty %} lang="{{ lang }}"{% endif %} class="{{ theme_class() }}" data-controller="dark-theme">
```
yguedidi: extracted the class attribute from the if condition as I see no reason for it to not be there when lang is empty

templates/layout.html.twig
```diff
@@ -134,19 +134,19 @@
                         <li class="divider"></li>
                     {% endif %}
                     <li>
-                        <a href="#" class="js-theme-toggle" data-theme="light">
+                        <a href="#" data-action="click->dark-theme#useLight:prevent">
```
yguedidi, line 137: `:prevent` is used to not navigate to `#`

yguedidi: technically those should be buttons, not links, but let's keep that for a future improvement (there are way more places to update..)

assets/controllers/dark_theme_controller.js
```diff
@@ -0,0 +1,39 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
+  connect() {
+    this.#choose();
+
+    this.mql = window.matchMedia('(prefers-color-scheme: dark)');
```
yguedidi, line 7: assumed `matchMedia` is always present as widely available now

assets/controllers/dark_theme_controller.js
```diff
@@ -0,0 +1,39 @@
...
+    this.element.classList.add('dark-theme');
+    document.cookie = 'theme=dark;samesite=Lax;path=/;max-age=31536000';
+  }
+
+  useAuto() {
+    document.cookie = 'theme=auto;samesite=Lax;path=/;max-age=0';
+    this.choose();
+  }
+
+  #choose() {
```
yguedidi, line 26: just in case `#` is used to make the function private to the class

assets/controllers/dark_theme_controller.js
```diff
@@ -0,0 +1,39 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
```
yguedidi, line 3: all JS code reworked in a simpler and more readable way

templates/base.html.twig
```diff
@@ -4,7 +4,7 @@
 <!--[if lte IE 7]><html class="no-js ie7 ie67 ie678"{% if lang is not empty %} lang="{{ lang }}"{% endif %}><![endif]-->
 <!--[if IE 8]><html class="no-js ie8 ie678"{% if lang is not empty %} lang="{{ lang }}"{% endif %}><![endif]-->
 <!--[if gt IE 8]><html class="no-js"{% if lang is not empty %} lang="{{ lang }}"{% endif %}><![endif]-->
-<html{% if lang is not empty %} lang="{{ lang }}"{% endif %} class="{{ theme_class() }}" data-controller="dark-theme">
+<html{% if lang is not empty %} lang="{{ lang }}"{% endif %} class="{{ theme_class() }}" data-controller="dark-theme shortcuts">
```
yguedidi: decided to have this controller registered at the root level because Mousetrap has no clean way to destroy itself.
for now its has no impact as we still full page reload, but when I'll introduce Turbo soon after this, it means the Mousetrap instance will be kept across navigate.
that's why I prepared already the code for that by mapping all shortcuts at the root level, and make then do something or not if possible, like if the target link or button is on the page

assets/controllers/shortcuts_controller.js
```diff
@@ -100,5 +100,16 @@ export default class extends Controller {
       e.preventDefault();
       this.showActionsTarget.click();
     });
+
+    const originalStopCallback = Mousetrap.prototype.stopCallback;
+
+    Mousetrap.prototype.stopCallback = (e, element, combo) => {
+      // allow esc key to be used in input fields of topbar
+      if (combo === 'esc' && element.dataset.topbarTarget !== undefined) {
```
yguedidi, line 110: this allows the ESC key to bubble to Mousetrap when the element is a topbar target, means either the add new entry input or the search input. so that the ESC shortcut to close them and show back the top bar title works

assets/controllers/entries_navigation_controller.js
```diff
@@ -0,0 +1,58 @@
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
```
yguedidi, line 3: this got reworked a bit to be simpler.
also, before it was applied on all card list, list the imports page, where it has no sense. now it's applied only on the entries pages displayed as cards.
also, cross pages navigation was broken when using key shortcuts, now fixed

package.json
```diff
@@ -102,7 +102,7 @@
     "build:dev": "encore dev",
     "watch": "encore dev --watch",
     "build:prod": "encore production --progress",
-    "lint:js": "eslint assets/*.js assets/js/**/*.js assets/controllers/*.js",
```
yguedidi: same here, not more files there so removed to keep ESLint running

package.json
```diff
@@ -88,8 +88,6 @@
     "hammerjs": "^2.0.8",
     "highlight.js": "^11.11.1",
     "icomoon-free-npm": "^0.0.0",
-    "jquery": "^3.7.1",
```
yguedidi, line 89: still downloaded by annotator, but it's already a win not to use it in our own code 🎉

yguedidi:
@j0k3r @Kdecherf @nicosomb @Simounet ready for a review! clearly it must be reviewed commit by commit!
I took the time to add some explanation comments on each commit to facilitate the review, enjoy 🙂

Simounet (approved):

yguedidi:
Thank you @Simounet !! merging it

PR #8075, opened 2025-03-12
Convert 403 errors to 404 errors

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | maybe
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This is a good security practice that avoids leaking that some resources exist on the platform.
Maybe a BC break if some relies on the returns HTTP status code, but worth it IMO


Kdecherf (dismissed):
I agree with the good practice to reduce risk of entity enumeration, I think we should mark this as a real BC

yguedidi:
> I think we should mark this as a real BC

@Kdecherf how? I added a marker on the PR title at least, any other way? and is it OK to ship this in 2.7 exceptionally? 🤞🏼

yguedidi:
@Kdecherf updated the changelog so we don't forget communicating about it on 2.7.0 release 🙂

yguedidi:
@j0k3r @Kdecherf @nicosomb my point of view about the BC break: as wallabag is a final product and not a dependency for other packages, I think we don't have the requirement to strictly follow SemVer, voila! 😁

j0k3r:
I dont think it's a BC because people should not have to use wallabag http response progammatically. We do have an API to handle most of the work. If people rely on that kind of thing, they must switch to the API or request the missing feature.

j0k3r (approved):

PR #8079, opened 2025-03-14
Update Dependabot configuration

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -40,10 +40,8 @@ updates:
   - yguedidi
   - Kdecherf
   ignore:
-  # until we add support for Symfony 5+
   - dependency-name: symfony/*
-    versions:
-    - ">= 5.0.0"
+    update-types: [ "version-update:semver-major" ]
 - package-ecosystem: github-actions
   directory: "/"
   schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,7 +21,7 @@ updates:
 - package-ecosystem: composer
   directory: "/"
   schedule:
-    interval: daily
+    interval: weekly
     time: "04:00"
     timezone: Europe/Paris
   open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -21,7 +21,7 @@ updates:
 - package-ecosystem: composer
   directory: "/"
   schedule:
-    interval: daily
+    interval: weekly
```
yguedidi, line 24: I think it's better to be weekly, so we focus on it once a week :)

.github/dependabot.yml
```diff
@@ -40,13 +40,8 @@ updates:
...
   - Kdecherf
   ignore:
-  - dependency-name: lcobucci/jwt
-    versions:
-    - ">= 4.2.0"
-  # until we add support for Symfony 5+
   - dependency-name: symfony/*
-    versions:
-    - ">= 5.0.0"
+    update-types: [ "version-update:semver-major" ]
```
yguedidi, line 44: Updated this way so that the rule can remain across all coming Symfony upgrades, as we want to handle major version upgrade manually

.github/dependabot.yml
```diff
@@ -40,13 +40,8 @@ updates:
   - yguedidi
   - Kdecherf
   ignore:
-  - dependency-name: lcobucci/jwt
-    versions:
-    - ">= 4.2.0"
```
yguedidi, line 45: After bump to minimum PHP 8.2 version the currently required version is `^4.3`

j0k3r (approved):

.github/dependabot.yml
j0k3r: I was ok with daily, but ok

yguedidi: also linked a bit to the rythme we do releases 😅 not worth having it daily IMO
weekly looks to be a good compromise

PR #8093, opened 2025-03-16
Add IsGranted to TagController

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

More precise security

### DIFF
--- a/src/Controller/TagController.php
+++ b/src/Controller/TagController.php
@@ -6,10 +6,12 @@ use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\QueryBuilder;
 use Pagerfanta\Adapter\ArrayAdapter;
 use Pagerfanta\Exception\OutOfRangeCurrentPageException;
+use Sensio\Bundle\FrameworkExtraBundle\Configuration\IsGranted;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Routing\Annotation\Route;
+use Symfony\Component\Security\Core\Security;
 use Symfony\Contracts\Translation\TranslatorInterface;
 use Wallabag\Entity\Entry;
 use Wallabag\Entity\Tag;
@@ -26,16 +28,19 @@ class TagController extends AbstractController
     private EntityManagerInterface $entityManager;
     private TagsAssigner $tagsAssigner;
     private Redirect $redirectHelper;
+    private Security $security;
 
-    public function __construct(EntityManagerInterface $entityManager, TagsAssigner $tagsAssigner, Redirect $redirectHelper)
+    public function __construct(EntityManagerInterface $entityManager, TagsAssigner $tagsAssigner, Redirect $redirectHelper, Security $security)
     {
         $this->entityManager = $entityManager;
         $this->tagsAssigner = $tagsAssigner;
         $this->redirectHelper = $redirectHelper;
+        $this->security = $security;
     }
 
     /**
      * @Route("/new-tag/{entry}", name="new_tag", methods={"POST"}, requirements={"entry" = "\d+"})
+     * @IsGranted("TAG", subject="entry")
      *
      * @return Response
      */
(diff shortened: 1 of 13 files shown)


src/Controller/TagController.php
```diff
@@ -87,13 +87,13 @@ public function addTagFormAction(Request $request, Entry $entry, TranslatorInter
      * Removes tag from entry.
      *
      * @Route("/remove-tag/{entry}/{tag}", name="remove_tag", methods={"GET"}, requirements={"entry" = "\d+", "tag" = "\d+"})
+     * @IsGranted("UNTAG", subject="entry")
+     * @IsGranted("DELETE", subject="tag")
```
yguedidi, line 91: added this because the tag can be removed completely when no more entries

src/Repository/TagRepository.php
```diff
@@ -85,6 +88,36 @@ public function findAllFlatTagsWithNbEntries($userId)
...
     }
 
+    /**
+     * Find all tags per user with nb entries.
+     *
+     * @param int $userId
+     *
+     * @return array<array{tag: Tag, nbEntries: int}>
+     */
+    public function findAllTagsWithNbEntries($userId)
```
yguedidi, line 98: I absolutely wanted that the TagVoter votes on real entity instances.
But the list of tags used `findAllFlatTagsWithNbEntries` that returns an array, making the is granted calls for edit and delete attributes fail.
so introduced this that returns a real entity instance, in addition to the number of entries.

Side note: on fact the tag entity is not used at all in the voter, as for now the grant is given only because being logged in. But in the futur this will certainly change when tags will be owned by the user, see #6784

src/Repository/TagRepository.php
```diff
@@ -13,8 +14,10 @@
  */
 class TagRepository extends ServiceEntityRepository
 {
-    public function __construct(ManagerRegistry $registry)
-    {
+    public function __construct(
+        ManagerRegistry $registry,
+        private string $tablePrefix,
```
yguedidi, line 19: we can now start using constructor promoted properties 🤗

templates/Tag/tags.html.twig
```diff
@@ -16,12 +16,14 @@
                     <a href="{{ path('untagged') }}">{{ 'tag.list.untagged'|trans }} ({{ nbEntriesUntagged }})</a>
                 </li>
             {% endif %}
-            {% for tag in tags %}
-                <li title="{{ tag.label }} ({{ tag.nbEntries }})" id="tag-{{ tag.id }}" class="chip">
+            {% for tagWithNbEntries in allTagsWithNbEntries %}
+                {% set tag = tagWithNbEntries.tag %}
```
yguedidi, line 20: real Tag entity instance here, to make is_granted calls work

src/Controller/TagController.php
j0k3r: Shouldn't this comment be in the code instead?

src/Repository/TagRepository.php
```diff
@@ -85,6 +88,36 @@ public function findAllFlatTagsWithNbEntries($userId)
...
+
+        $sql = <<<SQL
+            SELECT DISTINCT {$rsm->generateSelectClause()}, count(e.id) as nbEntries
+            FROM {$this->tablePrefix}tag t
+            LEFT JOIN {$this->tablePrefix}entry_tag et ON et.tag_id = t.id
+            JOIN {$this->tablePrefix}entry e ON e.id = et.entry_id
+            WHERE e.user_id = :userId
+            GROUP BY t.id
+            ORDER BY t.label
+        SQL;
```
j0k3r, line 113: Just to be sure, this is working properly on each database?

yguedidi: I'd say yes, as it's used only by the `/tag/list` route, that is called multiple times in TagControllerTest, and CI proves it runs properly on each DBMS

src/Controller/TagController.php
yguedidi: good idea

yguedidi:
@j0k3r you can check this [force-push diff](https://github.com/wallabag/wallabag/compare/2826c90a8650e74b20187c81a8f2186da2890c89..943bfd9162d0ab33719c9bed05fb9364c2e894ba) to review the changes. the previous force-push is just a rebase

PR #8098, opened 2025-03-17
Add IsGranted to EntryRestController

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

More precise security.

Replacing the entity manager clear by creating a new client is needed because without it tests about deleting an entry failed.
It's because of the clear, the user entity instance that is inside the entry and the one used in security became different instances, so the `===` check on user in EntryVoter failed.


Kdecherf (approved):
LGTM

PR #8117, opened 2025-04-02
Fix some depreciation notices

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Thanks to PHPStorm inspection and test deprecation notices


tests/Event/Subscriber/TablePrefixSubscriberTest.php
```diff
@@ -115,6 +116,6 @@ public function testPrefixManyToMany()
 
         $this->assertSame('yo_entry', $metaDataEvent->getClassMetadata()->getTableName());
         $this->assertSame('yo_entry_tag', $metaDataEvent->getClassMetadata()->associationMappings['tags']['joinTable']['name']);
-        $this->assertSame('yo_entry', $metaDataEvent->getClassMetadata()->getQuotedTableName(new MySQLPlatform()));
+        $this->assertSame('yo_entry', (new DefaultQuoteStrategy())->getTableName($metaClass, new MySQLPlatform()));
```
j0k3r, line 119: What if we switch to an other quote strategy in the entity?

yguedidi: I did it this simple way because we're not changing the quote strategy in the config, and to keep it a unit test, otherwise it would have been an integration test with KernelTestCase at least

j0k3r (approved):

PR #8203, opened 2025-05-26
Add availability to disable an importer

### LINKED ISSUE #8196: Add a configurator to enable / disable importers
As a wallabag instance administrator, I want to enable or disable a specific importer in the project configuration.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes https://github.com/wallabag/wallabag/issues/8196


yguedidi (commented):
One remark; I'd not use the security system for such checks, I'd reserve it for authorization things, based on the user. For feature availability, better just check the availability of the feature and if not throw the right exception (maybe a 404 in controllers for example)

And a side note: I'd really like that at some point we get ride of those configurable internal settings..! They really looks to me as "one shot settings" you configure once for all, and that are not likely to change once configured. So I think they should become env vars :)

yguedidi (changes requested):
Following up on my previous comment about treating importer availability more as static instance configuration than runtime internal settings:

If we go in that direction, I wonder if importer availability should eventually be handled at the importer registry level, rather than only by setting an `enabled` flag on every importer service.

For example, with env/config-based enabled importers, a compiler pass could filter the `wallabag.import` tagged services so that `ImportChain` only contains actually enabled importers. That would make the import index naturally reflect the configured instance capabilities.

That said, doing this fully would require checking the current direct dependencies on concrete importer services in controllers, so it may be too large for this PR.

The related point I would not want us to miss is the queue status code: `ImportController::checkQueueAction()` and `RabbitMQConsumerTotalProxy` still rely on a separate hardcoded list of importer aliases/queues. Once importer availability becomes configurable, this second list can easily drift from the actual enabled importer registry.

So I see this as either:

* something to address in this PR if we want importer availability to be registry-driven from the start; or
* a follow-up issue explicitly linked to this PR, to make queue counters derive from the configured/enabled importer registry instead of a fixed list.

j0k3r:
@yguedidi I'm more for the second one (and you did already open the issue)

PR #8220, opened 2025-05-29
Show display name on shared entries instead of username

### LINKED ISSUE #8219: Do not disclose username on shared entry page (use display name instead)
When sharing an entry the top of the page shows who has originally shared this entry. However it does so by disclosing the username, which cannot be changed and is used for login. I think it would be better for it to show the display name that users can change in their settings.

> This article has been shared by *** with [wallabag](https://app.wallabag.it/).

Seems like this is the only line that needs a change : 

https://github.com/wallabag/wallabag/blob/cad5a24fb6c2f42c65ad0caf0cffc0fe1ed0c523/templates/Entry/share.html.twig#L27

Aldarone:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | no
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes #8219

### DIFF
--- a/templates/Entry/share.html.twig
+++ b/templates/Entry/share.html.twig
@@ -24,7 +24,7 @@
         <header class="block">
             <h1>{{ entry.title|e|raw }}</h1>
             <a href="{{ entry.url|e }}" target="_blank" rel="noopener" title="{{ 'entry.view.original_article'|trans }} : {{ entry.title|e|raw }}" class="tool">{{ entry.domainName|removeWww }}</a>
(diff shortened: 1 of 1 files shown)


nicosomb (approved):
Thank you and congratulations for your first pull request on our project 🎉 !

Kdecherf (approved):

Issue #6856, opened 2023-08-19
Feature Request: Add an option to toggle whether to confirm the deletion


LumenYoung:
**Is your feature request related to a problem? Please describe.**

As suggested on the title, we have no option to toggle whether we want an extra confirmation to deletion. This creates a lot of friction to the batch processing (mostly deleting unwanted entries) of entries.   

**Describe the solution you'd like**

Solution would be: create a new option to toggle whether you want to have a pop-up to confirm your deletion of an entry.

**Describe alternatives you've considered**

Alternatively we can directly add a checkbox on the pop-up which is used to check if there should be confirmation on the later session.

Also, if only considering the batch processing of entries in my case, there is also another solution to multi-select all the entries and delete them in one fell swoop. The problem here is that the **multi-selection** mode in Wallabag allows you only to click on the checkbox, which is relatively small and requires more focus when you want to select. **Therefore, an alternative approach** to solve this could be that to make the clicking on the thumbnail in multi-selection mode not to enter the entry, rather to check the click box. 

I think the second alternative can be useful in boarder context and scenarios, where your batch processing is not only about deletion but also labelling or etc. But this is also harder to implement as I think about it.  

**Additional context**

Since I'm pretty new to Wallabag, I don't know yet whether this confirmation should be added at the config or the internal setting page, these two terms are overlapping concepts in my perception. Therefore I don't know where this setting, if implemented, should be placed.

Kdecherf:
Hello @LumenYoung

> Since I'm pretty new to Wallabag, I don't know yet whether this confirmation should be added at the config or the internal setting page, these two terms are overlapping concepts in my perception

Config is for per-user settings whereas internal setting is for instance-wide settings. See the internal setting page as a setting page for administrators.

> The problem here is that the multi-selection mode in Wallabag allows you only to click on the checkbox, which is relatively small and requires more focus when you want to select. Therefore, an alternative approach to solve my problem could be that to make the clicking on the thumbnail in multi-selection mode not to enter the entry, rather to check the click box

I can't agree more, we should create a dedicated issue for that

@wallabag/core I think we should change the milestone to 3.0 to move faster on flex and sf5 topics.

LumenYoung:
Hello, @Kdecherf,

Thank you for your reply and your explaination. I'm glad that you also agree on the multi-selection part of my argument. I also agree it would be really nice to have #6971 on Wallabag!

PR #8247, opened 2025-06-04
Keep track of entry deletions and expose it in the API

casimir:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes (TODO)
| Translation   | no
| CHANGELOG.md  | no (?)
| License       | MIT

A proposition to fix the *exhaustive syncing problem* for API clients, as discussed in #2906.
This problem is a pain point, sometime of blocker, for third party clients (e.g.: https://github.com/wallabag/wallabag/issues/2906#issuecomment-2912852397, https://github.com/wallabag/android-app/issues/768, https://github.com/wallabag/ios-app/issues/185, [Read-It-Later#32](https://gitlab.gnome.org/World/read-it-later/-/issues/32)). 

*Disclaimer, my last experience with Symfony goes back to the tutorial for version 1.2. I have a lot of experience with Django though and tried to transpose what I know of web frameworks. Beware of non-idiomatic things.* 

This new feature revolves around a log of entry deletions. Deletions are recorded using `EntryDeletedEvent` and are exposed in the API with a new endpoint `/api/entry-deletions`.
The API endpoint is heavily inspired by `/api/entry` and the lifecycle logic is based on how downloading image is handled.
Just like the downloaded image, this feature comes with a management command to purge the expired data `wallabag:purge-entry-deletions`.

The expiration time is configured with a parameters (in days) `wallabag.entry_deletion_expiration_days`. It makes it a little more tricky for third party apps but I thought that it respects the spirit of "you instance, your tunning".
I also wanted to expose the cutoff date in `/api/info` to help clients transition but couldn’t find how at that moment. I don't know if this would be useful anymore.

I tried to cover all the new code with tests, including specific tests for all the edge cases I could think of (there are some...)

I might have been a little carried away with the WOA annotations (`src/OpenApi`). Depending on how you prefer to include adjacent features, it could be split in another PR.
As a developer consuming a good part of the API it has been very frustrating to not have the responses schema.

If this PR lands, I plan on proposing another update to the documentation. It would go into https://github.com/wallabag/doc right?

Finally, I tried to split the several aspects of this changeset in dedicated commits. I know how hard big changesets are to review but feel free to ask any question! I am also lurking on the wallabag matrix channel.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -156,7 +156,8 @@
         "wallabag/rulerz": "dev-master",
         "wallabag/rulerz-bundle": "dev-master",
         "willdurand/hateoas": "^3.12",
-        "willdurand/hateoas-bundle": "^2.7"
+        "willdurand/hateoas-bundle": "^2.7",
+        "zircote/swagger-php": "^4.11.1 || ^5.0"
     },
     "require-dev": {
         "dama/doctrine-test-bundle": "^8.2.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "00b8f95df6ec0572c06ad6f34847405c",
+    "content-hash": "b47338ba84f717112fd3f36632c65e03",
     "packages": [
         {
             "name": "babdev/pagerfanta-bundle",
@@ -19449,6 +19449,6 @@
         "ext-tokenizer": "*",
         "ext-xml": "*"
     },
-    "platform-dev": [],
-    "plugin-api-version": "2.3.0"
+    "platform-dev": {},
+    "plugin-api-version": "2.6.0"
 }


yguedidi:
@casimir what a big contribution, thank you!

before reviewing the code, I really need to understand the need, the use case..

if I understand correctly, on third party apps, there is a sync issue. for example deleted entries on the server are still present on third party apps on sync, right?

at first sight, I'm not a big fan of tracking events for such thing, so I'd better look for other kind of solutions.
what are the exact sync needs? I'd say the apps would like to know what changed since the last sync, right? in a performance way?
what about tracking an "updated at" in entries each time there is a change in the DB row?
then a new dedicated endpoint for "entries stats", we pass a list of entry ids and a date, and it returns the ids grouped by "same", "changed", and "deleted"?

Kdecherf:
Hello @casimir,

Thanks for your change proposal. As a user of the android app, I feel the pain of the cleanup phase because of this limitation.
However I'm not comfortable with the idea of pushing a history table for this use-case.

@nicosomb following discussions on #2906 I would like to reconsider the option of a soft delete for entries and a new deleted_at field in the entry table.
Thus, deleting an entry would not immediately delete the actual data from the database. At most we could remove data like title and content while keeping the rest of the record with the date of deletion. At the end we could retrieve the list of IDs of articles deleted since a given date (the last sync of the third-party app for example).

The soft delete option may help us complete #6856 in the process.

cc @yguedidi @j0k3r

casimir:
@yguedidi the need is to address the difficulties for third-parties to sync the entries state of an account, be it a full offline DB or an incremental cache. As of now, it is not easy to do without a lot of API calls.

At first glance, `/api/entries?since=XXXX` is made exactly for that but it is not possible to implement such a workflow with it.
This is because it only concerns the entries in server's DB at the moment of the call and not the moment of the last client sync. In short, every entry deleted from another client since the last sync are simply forgotten by the server for ever. Meaning that server and client(s) can't reconciliate the broken timeline, leaving the client with no other choice but to either sync everything again or try to compute a delta and find what it should evict in its data.

Just for the exercice, let's imagine an account with 2k entries (archived or not is irrelevant here). You will need to send 40 requests to sync the whole entries list with a pagination of 50 (the default I saw in a few apps). That's objectively a lot for a simple entry deletion.
This is also detrimental to the clients UX since a server round trip is far from instantaneous. For example, https://app.wallabag.it responds on average around 250 ms on this endpoint (scientifically eyeballed by myself).

Some mobile apps (at least frigoligo and android-app) mitigate that by:
1. Getting the total count using the lightest query possible (`/api/entries?perPage=1&detail=metadata`).
2. Comparing this count with the count of local entries.
3. If different, fetching the exhaustive list of IDs from the server to compute a diff. (frigoligo use `perPage=100` here)
4. Find the deleted entry ID and replicate the suppression locally.
This technique allows to avoid a lot of the requests but at the cost of complexity. Notably, this algo is transactional and don't like network errors.

In short, the lack of means to sync with the single source of truth makes the implementation of 3p clients more complex than usual for this kind of thing and the server round-trips needed for that make devs and users of third-party apps frustrated.

---

@Kdecherf in retrospect, I am not a fan of the history table either. Honestly, the reason why I went this way was because I wasn't confident to tamper with critical code. Particularly `Entry` and everything that goes around it. I went through years of conversation on GitHub around this subject but couldn't find a preferred solution. So I went with the simplest approach I could think of at the time. Which was of course more complex once the implementation started 😅.

Technical aspects set aside, here is why I went this way.
First, I think I am better at communicating with algorithm and architecture than with prose and I wanted to break the status quo.
Second, as I stated just before, I wasn't confident at all to mingle with existing code of a stack I have close to 0 experience with.
Third, it took me several attempts at implementing an [external solution](https://github.com/casimir/freon/?tab=readme-ov-file#api-inconsistencies-in-wallabag) before I had the motivation to contribute to Symfony code.
Fourth, because I didn't know the tech, I went all-in on "add things instead of changing things". I put a lot of thought into the design and potential scaling issues. I was particularly concerned about problems on big instances, like storage buildup and network usage patterns. Maybe that's not an aim for the project though.

In the end I settled for a good but not perfect solution, certainly not the simplest one. My main goal was to re-ignite the discussion about this issue and here we are. Thanks for your consideration btw.
IMHO the soft delete solution is the right one, it's almost the state of the art for this kind of use case. However, I am not familiar with the design of the server and I have no clue about the details.

casimir:
For reference here a similar project with the same issue and its own solution: https://codeberg.org/readeck/readeck/pulls/678 . It goes one step further and provides a dedicated endpoint for incremental syncing. A little like `/api/entries?detail=metatada` with an infinite page size.
I almost went this way at first and I like the separation of endpoints for "entry data fetching" and "sync metadata listing". WDYT?

nicosomb:
@casimir thank you so much for this great contribution. 

I agree with @Kdecherf about soft delete.

casimir:
If a consensus is reached between the maintainers, I might be able to allocate some time to propose a PR for the soft delete feature in the coming days, provided it's not already being worked on.

casimir:
Hello again! To be clear, I'd like to know if a PR for the soft delete feature we discussed has a chance to be included while the topic is still 'hot'. I'd like to work on this, but it's not clear to me what requirements need to be met to get maintainers/reviewers final approval.
Until then, I'll keep this PR open if it's okay with you.

yguedidi:
Hello @casimir ! And thanks for pushing for this, there is clearly something to solve there!

I would tend to agree with the soft delete solution too.

But before any PR, I think it's better to open an issue about it, with a proposed specification on how to implement it :)

As I like to split problems to better solve them, here I see at least 3 sub topics:

- tracking properly soft deletes, hiding deleted entries from relevant places
- exposing the right data for sync purpose
- cleaning of the deleted entries in some way

Maybe an issue for each of them are worth it as solutions are multiple.

Few insights on the first one: it's likely to be an addition of a deleted_at field to Entry, but the hard part would be to identify all places that needs to use it to filter entries, and decided on how we handle edge cases (like what if a user adds an entry that is soft deleted?)

yguedidi:
Meanwhile, I think this PR "as is" can be closed

j0k3r:
I agree with the overall idea of introducing a soft delete.
Maybe a main issue with some sub-issues listed by @yguedidi could be a good starting point.

Let us know @casimir if you want to create these issues or if you want us to create them.

casimir:
I don't have a lot of available bandwidth these days. You can create the issues and I'm still willing to work on the implementation.

PR #8291, opened 2025-06-12
Hide sensitive information in production

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

`templates/bundles/TwigBundle/Exception/error.html.twig` is only used in production env. 
And it displays sensitive information, like the database name. 

We have to remove that. 

We can also enhance the message, with a link to the `wallabag_support_url` setting. 

## Before the PR 

Capture d’écran 2025-06-12 à 21 51 44

## After the PR 

Capture d’écran 2025-06-12 à 22 05 14

### DIFF
--- a/tests/Controller/ConfigControllerTest.php
+++ b/tests/Controller/ConfigControllerTest.php
@@ -1130,7 +1130,7 @@ class ConfigControllerTest extends WallabagTestCase
 
         $this->assertSame(400, $client->getResponse()->getStatusCode());
         $this->assertGreaterThan(1, $body = $crawler->filter('body')->extract(['_text']));
-        $this->assertStringContainsString('Bad CSRF token.', $body[0]);
+        $this->assertStringContainsString('400: Bad Request', $body[0]);
     }
 
     public function testChangeLocaleWithReferer()
--- a/tests/Controller/EntryControllerTest.php
+++ b/tests/Controller/EntryControllerTest.php
@@ -732,7 +732,7 @@ class EntryControllerTest extends WallabagTestCase
 
         $this->assertSame(400, $client->getResponse()->getStatusCode());
         $this->assertGreaterThan(1, $body = $crawler->filter('body')->extract(['_text']));
-        $this->assertStringContainsString('Bad CSRF token.', $body[0]);
+        $this->assertStringContainsString('400: Bad Request', $body[0]);
     }
 
     /**


yguedidi (approved):
CI to be fixed

j0k3r:
During we are looking for an error message which got removed by your PR. You need to update tests.

nicosomb:
in fact, when I see the tests (which have messages with useful information, even in production environment), I don't know what to do whis this PR.  

@j0k3r @yguedidi @Kdecherf are you OK with the idea of this PR ?

Kdecherf:
@nicosomb here's a quick ExceptionListener example, it may need changes as I didn't have time to test this in a prod setup. The idea is to ensure that we have the exception_id attribute shown in both twig template and app logs

`src/Event/ExceptionListener.php`:

``` php
<?php

namespace Wallabag\Event\Listener;

use Symfony\Component\HttpKernel\Event\ExceptionEvent;
use Psr\Log\LoggerInterface;

class ExceptionListener
{
    private LoggerInterface $logger;

    public function __construct(LoggerInterface $logger)
    {
        $this->logger = $logger;
    }

    public function onKernelException(ExceptionEvent $event)
    {
        $request = $event->getRequest();

        $exceptionId = bin2hex(random_bytes(10));

        $request->attributes->set('exception_id', $exceptionId);

        // If the exception_id parameter is not reachable from the twig template, we may need to log the exception ourselves?
        // $this->logger->error('Exception ID: ' . $exceptionId, ['exception' => $event->getThrowable()]);
    }
}
```

`app/config/services.yml`:

``` yaml

    Wallabag\Event\Listener\ExceptionListener:
        tags:
            - { name: kernel.event_listener, event: kernel.exception }

```

yguedidi:
@Kdecherf nice idea, but to me deserves its own PR, so we can do the specifications on how best to implement it.
better to keep this PR scoped to the removal of sensitive information

j0k3r (approved):

Kdecherf:
@yguedidi fine for me, @nicosomb let's move forward with your change
However I would add something like "contact your server admin or check logs"

yguedidi (approved):
As we're likely to release directly a 3.0 after the 2.7, shouldn't this target 2.7?

PR #8332, opened 2025-06-29
Fix reading time computation for short entries

### LINKED ISSUE #1831: RSS feeds paging
Currently, when querying a RSS feed you can limit the number of result through the limit= parameter (or in global settings).
However, there is no way to get the older content over that limit via RSS feed. 
It would be nice to have some paging, thanks to a ?page=x or ?paged=x parameter as in Wordpress, to make every link reachable.

Once implemented, it would be easy to add a "My favorite readings" page on a website. A kind of Shaarli that could be filled in a much frictionless way - with no comments.

andreadecorte:
Time computation for short entries may give incorrect results with reading speed > 400.

Reading speed = 420
Reading time = 1 min
Displayed reading time = 0 min

With this change this will be correctly displayed as 1 min also in this case

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes #1831

### DIFF
--- a/templates/Entry/_reading_time.html.twig
+++ b/templates/Entry/_reading_time.html.twig
@@ -1,6 +1,6 @@
 {% set reading_time = entry.readingTime / app.user.config.readingSpeed * 200 %}
 <i class="material-icons grey-text">timer</i>
-{% if reading_time > 0 %}
+{% if reading_time|round > 0 %}
     <span>{{ 'entry.list.reading_time_minutes_short'|trans({'%readingTime%': reading_time|round}) }}</span>
 {% else %}
     <span>{{ 'entry.list.reading_time_less_one_minute_short'|trans|raw }}</span>


casimir:
@nicosomb that would be nice for mobile apps too.

yguedidi:
As a bugfix, this PR should target 2.6 branch, not master.

And I'd suggest an alternative solution, that will centralize the computation and allow to expose the value in the API: introduce a `getUserReadingTime()` in `Entry` and use it in the template.

@andreadecorte are you willing to rework this PR or should we close it and someone else open a new one?

andreadecorte:
> As a bugfix, this PR should target 2.6 branch, not master.
> 
> And I'd suggest an alternative solution, that will centralize the computation and allow to expose the value in the API: introduce a `getUserReadingTime()` in `Entry` and use it in the template.
> 
> @andreadecorte are you willing to rework this PR or should we close it and someone else open a new one?

ok so basically move the computation logic to Entry`, including the rounding logic, so it will fix the bug. Yeah I can take a stab at it and target 2.6 branch too

yguedidi:
Perfect, thanks!

andreadecorte:
@yguedidi I just noticed that 2.6 is quite behind, so master doesn't apply in a clean way. I am curious, should I duplicate the change in the 2 branches? as it stands there is not much possibility to backport there

yguedidi:
@andreadecorte i think the best is to reset your branch on 2.6 and redo the fix on top of it. We will handle the up merge of 2.6 in master and handle conflicts, so no need to apply the fix on master ;)

andreadecorte:
@yguedidi I did the suggested changes against 2.6. It includes a minor refactoring to extract the 200 value which was in several places, let me know what you think

src/Wallabag/CoreBundle/Tools/Utils.php
```diff
@@ -4,6 +4,9 @@
 
 class Utils
 {
+
```
yguedidi, line 7: extra line to remove

src/Wallabag/CoreBundle/Tools/Utils.php
```diff
@@ -4,6 +4,9 @@
...
+    public const DEFAULT_WORDS_PER_MINUTE = 200;
```
yguedidi, line 8: I would have extracted this refactor to a dedicated commit

src/Wallabag/CoreBundle/Tools/Utils.php
```diff
@@ -28,6 +31,6 @@ public static function generateToken($length = 15)
      */
     public static function getReadingTime($text)
     {
-        return floor(\count(preg_split('~([^\p{L}\p{N}\']+|(\p{Han}|\p{Hiragana}|\p{Katakana}|\p{Hangul}){1,2})~u', strip_tags($text))) / 200);
+        return floor(\count(preg_split('~([^\p{L}\p{N}\']+|(\p{Han}|\p{Hiragana}|\p{Katakana}|\p{Hangul}){1,2})~u', strip_tags($text))) / Utils::DEFAULT_WORDS_PER_MINUTE);
```
yguedidi, line 34:
```suggestion
        return floor(\count(preg_split('~([^\p{L}\p{N}\']+|(\p{Han}|\p{Hiragana}|\p{Katakana}|\p{Hangul}){1,2})~u', strip_tags($text))) / self::DEFAULT_WORDS_PER_MINUTE);
```

src/Wallabag/CoreBundle/Resources/views/Entry/_reading_time.html.twig
```diff
@@ -1,7 +1,7 @@
-{% set reading_time = entry.readingTime / app.user.config.readingSpeed * 200 %}
+{% set reading_time = entry.getUserReadingTime %}
```
yguedidi, line 1: this should work too:
```suggestion
{% set reading_time = entry.userReadingTime %}
```

src/Wallabag/CoreBundle/Helper/RuleBasedTagger.php
```diff
@@ -133,7 +133,7 @@ private function getRulesForUser(User $user)
     private function fixEntry(Entry $entry)
     {
         $clonedEntry = clone $entry;
-        $clonedEntry->setReadingTime($entry->getReadingTime() / $entry->getUser()->getConfig()->getReadingSpeed() * 200);
+        $clonedEntry->setReadingTime($entry->getUserReadingTime());
```
yguedidi, line 136: note that before the set value is not rounded, while now with the new function it is.
wondering if this has an impact on if entries satisfies or not the rules.

andreadecorte: yep, it changes slightly the logic as we are rounding before comparing vs taking the float value without.
I think new behavior makes sense, as it aligns with UI behaviour like in the case below:
```
article reading time 17 min
user reading speed 400
article user reading time 9
```
This gives:
- before: a rule with >= 9 reading speed wouldn't have matched, despite the UI showing 9 so possibly user is confused
- after: a rule with >= 9 reading speed would now match the article

But let me know if you prefer keeping old behaviour

yguedidi: Thanks for the detailed explanation! To me it seems in the scope of the bugfix so fine for me!

src/Wallabag/CoreBundle/Resources/views/Entry/_reading_time.html.twig
andreadecorte: changed

src/Wallabag/CoreBundle/Tools/Utils.php
andreadecorte: extracted this part in a dedicated commit

j0k3r (approved):

yguedidi (approved):

PR #8400, opened 2025-08-09
Fix async browser import with nested children

tsipinakis:
The `parseEntry` function recurses into children and dispatches their processing. However, validate will reject any entry that does not have a url e.g. parents.

We need to call parseEntries before so that the processing of children is dispatched, otherwise nothing is imported.


j0k3r:
Do you have an example? Because if we move the validation _after_ the parsing the parsing might fail because the validation wasn't done.

Issue #7180, opened 2024-01-14
class name conflicts in wallabag's css


HolgerAusB:
### Environment

* **Version**: 2.6.7
* **Installation**: GIT

### What steps will reproduce the bug?

Recently I saw several websites which uses the same class names as they are in wallabag's material.css which caused conflicts in the visualization. So I needed to string_replace these class-names in the site specific configs. But this could lead to problems if these class names appear in the normal body text of the article.

### Suggestion
 
So we need either more unique classnames as `wallabag-text-primary` instead of `text-primary`

or

we need a crawler that is renaming all class (and id?) elements from fetched html which are used in our css files: e.g.
`text-primary` -> `notwallabag-text-primary`

I'd prefer the first option, because the second one might break some site rules

#### * usbeketrica.com
uses `.text-primary`. That results in a white text-color, which goes invisible on the light-theme

#### * apple.com
uses `.video-container` which causes wallabag to enlarge the element, while the video could not be embed.
e.g. https://www.apple.com/newsroom/2023/10/apple-supercharges-24-inch-imac-with-new-m3-chip/

#### * telepolis.de
uses` <h3 class="subheading">`. While this class is not part of materials css, these subheaders where stripped from wallabag. After replacing the class in site config they remain in text. So might be a different problem here.

#### * next.ink
uses `<h3 class="wp-block-heading">`. This class is also not part of material.css but subheadings get stripped by wallabag.

j0k3r:
That's a good point.
Is it easily doable @Simounet?

yguedidi:
just curious, I'm wondering why do we have remaining classes from the original content? isn't the fetch content stripped from all non necessary things to be displayed inside wallabag in "reader mode"?

Simounet:
It's doable but like @yguedidi I think it could be great to strip all classes before storing the content into our db.

HolgerAusB:
hmmm, I see what you mean, @Simounet. But on the other hand, it could be helpful to keep them, if somebody uses custom.css. So I wrote a still dirty mastodon-wrapper (python cgi), which get toots including all ascendants and descendants via api and construct a plain html out of this, so I can catch that with wallabag. And I am pushing own classes for some elements to build a nicer view to that 'articles'. That would be gone then.

Maybe we could store store the classes in the db but have a toggle to optionally remove them when reading from db and showing in the UI and/or API. Maybe with persistent toggle per article or domain.

HolgerAusB:
[...]

PR #8505, opened 2025-10-12
feat: Render articles within their own (shadow) DOM

quent1-fr:
This pull request introduces a new feature that renders articles within their own shadow DOM. The shadow DOM provides encapsulation for the article's styles and structure, ensuring that there are no conflicts with the main document or other components on the page.

I'm open to all comments and suggestions, so please feel free to share your thoughts or any improvements you think could be made.


j0k3r:
That's interesting.
Can you share a screenshot before/after when a problem of CSS conflicts for example?

yguedidi:
Interesting approach!
Will try to give feedback tomorrow

yguedidi:
@quent1-fr do you have any examples of URLs that lead to styling issues? just so that we align on the problem we're trying to solve with this PR before continuing

quent1-fr:
@yguedidi Unfortunately, I don’t have specific worst-case examples at hand. Most of the time, the issues I encounter are minor, such as strange rendering glitches. However, I occasionally experience more critical problems, like a gray overlay blocking the entire page - including the navbar - which can be quite annoying.

One notable instance was when saving a page on [LeBonCoin.fr](https://www.leboncoin.fr/):
image

Simounet:
Hi there, seems like a good idea but I'm kind of worried about putting the main content on the JS side instead of plain HTML. Did you think about using CSS `@layer`? To me, we should be able to prefix the wallabag's classes, or unset the style inside `<article>`. Sometimes it's a `<button>` or an `` that takes wrong styles.

yguedidi:
To be honest I'm mixed feelings about this PR... on one side I'm impressed by the creativity of such solution! on the other side I don't think we should need it 😅 

From my point of view, the rendered reading mode could almost be a pure Mardown, or only semantic HTML tags, no need for classes at all, not even for styling.
IMO we should have controlled reading mode styles that applies to those semantic HTML tags (headings, paragraphs, links, quotes, lists, etc).

There is https://github.com/wallabag/wallabag/issues/7180 that mention potential use cases for keeping CSS classes:

- building/debugging site configs
- very advanced generated content

To me the first one is needed, but it's clearly a different use case from the rendered reading mode 99% of users will use.
We may need something like a site config builder helper, or a debug mode in reading mode preserving the classes? (I prefer the former, more "explicit" for the use case)

I have no special point regarding the other advanced use case of @HolgerAusB.
At first sight it looks a bit weird to custom style elements that are part of a reading mode, I mean other that the global styling of the reading mode itself (like titles styles, paragraph styles, links styles, etc)

What do you think?

PR #8585, opened 2025-12-03
Prefetch content ISO to fetch URL content

Simounet:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Trying to be ISO when we are submitting the content from an external source (wallabagger for instance) using Graby `setContentAsPrefetched`. At the moment, the external content is not processed as a normal submitted URL.

### DIFF
--- a/src/Helper/ContentProxy.php
+++ b/src/Helper/ContentProxy.php
@@ -44,21 +44,31 @@ class ContentProxy
     {
         $this->graby->toggleImgNoReferrer(true);
         if (!empty($content['html'])) {
+            $this->graby->setContentAsPrefetched($content['html']);
             $content['html'] = $this->graby->cleanupHtml($content['html'], $url);
         }
 
-        if ((empty($content) || false === $this->validateContent($content)) && false === $disableContentUpdate) {
-            $fetchedContent = $this->graby->fetchContent($url);
+        if (false === $disableContentUpdate) {
+            $fetchedContent = array_filter($this->graby->fetchContent($url), static function ($value, $key) {
+                if ('html' === $key) {
+                    return true;
+                }
 
-            $fetchedContent['title'] = $this->sanitizeContentTitle(
-                $fetchedContent['title'],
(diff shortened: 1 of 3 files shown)


Simounet:
Hi @nicosomb @yguedidi @j0k3r @Kdecherf what do you think about this failing test?
````
1) Tests\Wallabag\Controller\Import\WallabagV1ControllerTest::testImportWallabagWithFile
Mimetype for http://www.framablog.org/ is empty
Failed asserting that a string is empty.
````

Imported content from a file is handled as prefetched content. Instead of an empty string, we've got `text/html` as mimetype because it's the default content type's value. Sounds ok to me but I might be wrong.

Kdecherf:
Hi @Simounet, would you mind rebasing your PR, I'll take time to review it afterwards

Kdecherf:
> Hi [snipped] what do you think about this failing test?
> 
> ```
> 1) Tests\Wallabag\Controller\Import\WallabagV1ControllerTest::testImportWallabagWithFile
> Mimetype for http://www.framablog.org/ is empty
> Failed asserting that a string is empty.
> ```
> 
> Imported content from a file is handled as prefetched content. Instead of an empty string, we've got `text/html` as mimetype because it's the default content type's value. Sounds ok to me but I might be wrong.

Sounds good to update the expected value of the test

Simounet:
Hey @Kdecherf , thanks for looking at my PR. Everything is update-to-date.

Kdecherf (approved):
LGTM

Kdecherf:
Thanks @Simounet

yguedidi:
@Simounet @Kdecherf shouldn't the target 2.6? Seems like a bugfix

Simounet:
Yep, should be backported.

PR #8712, opened 2026-03-10
Centralize Materialize select rendering in a global form theme

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Centralize collapsed select rendering in a shared Twig form theme and route the config page selects plus the entry filter language select through it.

This fixes the hidden entry language filter select by moving it off the unsupported plain Materialize select path that applied `opacity: 0`. It also moves the remaining repo-owned select renderings onto the same theme-driven Materialize path.

This is the first part of a broader effort to move form component rendering into Twig form themes in follow-up pull requests.

### DIFF
--- a/templates/Entry/entries.html.twig
+++ b/templates/Entry/entries.html.twig
@@ -228,13 +228,11 @@
                     </label>
                 </div>
 
-                <div class="col s12">
-                    {{ form_label(form.language) }}
-                </div>
-
-                <div class="input-field col s12">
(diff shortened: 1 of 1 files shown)


yguedidi:
@j0k3r possible to have a new look to this please?

PR #8716, opened 2026-03-12
Move Doctrine test fixtures under tests/fixtures

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Move the Doctrine fixture classes that are only loaded in `APP_ENV=test` out of the root `fixtures/` directory and into `tests/fixtures/Doctrine`, and retarget Composer and test service discovery to the `Wallabag\Tests\Fixtures\Doctrine` namespace.

This is preparation work for future dev-environment fixtures so local development can start from seeded data instead of an empty database, while keeping the current test-only Doctrine fixtures separate from that upcoming dev-only setup.

### DIFF
--- a/composer-dependency-analyser.php
+++ b/composer-dependency-analyser.php
@@ -13,7 +13,6 @@ $config
     ->addPathToScan(__DIR__ . '/migrations', false)
     ->addPathToScan(__DIR__ . '/src', false)
     ->addPathToScan(__DIR__ . '/web', false)
-    ->addPathToScan(__DIR__ . '/fixtures', true)
     ->addPathToScan(__DIR__ . '/tests', true)
     ->ignoreErrorsOnPackages([
         'doctrine/common',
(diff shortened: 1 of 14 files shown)


composer-dependency-analyser.php
```diff
@@ -13,7 +13,6 @@
     ->addPathToScan(__DIR__ . '/migrations', false)
     ->addPathToScan(__DIR__ . '/src', false)
     ->addPathToScan(__DIR__ . '/web', false)
-    ->addPathToScan(__DIR__ . '/fixtures', true)
```
yguedidi: safe to remove as already covered by the line below for the tests folder

PR #8725, opened 2026-03-15
Upgrade CI workflows off Node.js 20 deprecated actions

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

GitHub Actions is deprecating Node.js 20 as a runner runtime, with a forced cutover on June 2, 2026. Two deprecation warnings were present, each sourced from a different third-party action's internal dependency:

- `actions/cache@v4` (Node.js 20), used inside `ramsey/composer-install@v3`
- `actions/github-script@v7` (Node.js 20), used inside `codecov/codecov-action@v5`

Updates `ramsey/composer-install` from v3 to v4, which adopts `actions/cache@v5` (Node.js 24) internally.

Adds `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` to the CI workflow env, since `codecov/codecov-action` (latest: v5.5.2) has no Node.js 24 release yet. Also adds the same env to the dependabot automerge workflow as a forward-compatibility measure.

Upgrades `shivammathur/setup-php` from the `v2` floating tag to `2.37.0`, which updates the action's own runtime from Node.js 20 to Node.js 24.

> **Note on remaining annotation:** the `actions/github-script@v7` warning from `codecov/codecov-action@v5` will still appear in CI even with `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` set. This is a known GitHub runner bug ([actions/runner#4295](https://github.com/actions/runner/issues/4295)) where the annotation fires based on the action's declared `using: node20` before the env var is checked. The action actually executes on Node.js 24 at runtime — the annotation is a false positive. No fix is available upstream until either codecov updates their action or GitHub resolves the runner bug.

best reviewed commit by commit

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -12,6 +12,9 @@ env:
   DOCTRINE_DEPRECATIONS: none
   PGPASSWORD: wallabagrocks
   COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  # codecov/codecov-action@v5 internally calls actions/github-script@v7 (Node.js 20);
+  # force Node.js 24 until codecov releases a fix
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 jobs:
   phpunit:


j0k3r (approved):
It's weird because they should be updated automatically by dependabot?
https://github.com/wallabag/wallabag/blob/0a7837dbd76f6e7ccd74c375447982355fb3dfd6/.github/dependabot.yml#L44-L50

Or it's only related to _official GitHub images_? 🤔

yguedidi:
@j0k3r thanks for the review and approve!

In the end I'll close the PR I think, let's rely on dependabot.
Especially that apparrently `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` doesn't remove the annotation, so not worth it.

I think dependabot didn't triggered the update just because it's weekly and the releases are quite recent:
- ramsey/composer-install@v4 released 1h ago
- shivammathur/setup-php@v2.37 released 8h ago

PR #8735, opened 2026-03-16
Define DATABASE_URL directly in CI and remove per-database parameter files

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Instead of copying a database-specific YAML file to `app/config/parameters_test.yml` at the start of each CI job, each job now sets `DATABASE_URL` directly via a GitHub Actions job-level `env:` block. A conditional expression selects the right value from the matrix `database` variable, using `github.workspace` for the absolute path in the SQLite case.

The three per-database parameter files under `app/config/tests/` are no longer needed and have been removed along with the now-empty directory.

### DIFF
--- a/app/config/config_prod.yml
+++ b/app/config/config_prod.yml
@@ -1,6 +1,9 @@
 imports:
     - { resource: config.yml }
 
+parameters:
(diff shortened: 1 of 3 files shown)


yguedidi:
replaced by #8743

PR #8741, opened 2026-03-17
Expose CI validation commands in the Makefile

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

The CI `coding-standards` and `translations` workflows run several validation commands that had no Makefile equivalent, requiring developers to memorise raw tool invocations or read CI YAML to reproduce failures locally.

This PR adds the missing targets to the Makefile. All new targets use the existing `$(PHP_NO_XDEBUG)`, `$(COMPOSER)`, `$(PHP)`, and `$(YARN)` execution variables, so they route through the Docker Compose `php` service automatically when the stack is running, or fall back to local binaries otherwise.

## New targets

- `check-cs` — PHP CS Fixer dry-run (reports issues without auto-fixing, matching what CI does; complements the existing `fix-cs` which auto-fixes)
- `twigcs` — TwigCS with `--severity=error --display=blocking` on `app/` and `src/`
- `composer-validate` — `composer validate`
- `composer-normalize` — `composer normalize --dry-run --no-check-lock`
- `composer-dependency-analyser` — `bin/composer-dependency-analyser`
- `lint-translations` — `bin/console lint:yaml translations -v`
- `lint` — composite that runs all six above plus `phpstan`, `lint-js`, and `lint-scss`, mirroring the full CI coding-standards job in one command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,9 @@ test-integration: ## Launch integration testsuite (append test paths to narrow s
 test-functional: ## Launch functional testsuite (append test paths to narrow scope)
 	@$(PHP_NO_XDEBUG) -dmemory_limit=-1 bin/phpunit --testsuite functional -v $(PHPUNIT_PATH_ARGS)
 
+check-cs: ## Check PHP coding standards without making changes
+	@$(PHP_NO_XDEBUG) bin/php-cs-fixer fix --verbose --dry-run
+
 fix-cs: ## Run PHP-CS-Fixer
 	@$(PHP_NO_XDEBUG) bin/php-cs-fixer fix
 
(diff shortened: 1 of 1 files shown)


yguedidi:
@Kdecherf I'm planning to migrate the doc and website code in the main repository here, so that such changes can update cross layers!
Help maintaining an up to date doc and site with evolutions on the project.
Docker is also in my roadmap.
I can surely do the PR for the import of doc repo code and adapt it to be ready for the Netlify config, but then I'll need @nicosomb I think to change the Netlify live configuration

PR #8745, opened 2026-03-19
Allow WALLABAG_SITE_CONFIG_FOLDERS to configure an external site config folder

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

The [wiki](https://github.com/wallabag/wallabag/wiki/Auto%E2%80%90Updating-wallabag%27s-site%E2%80%90depended-config) documents a manual process where users must edit `app/config/services.yml` to point `Graby\SiteConfig\ConfigBuilder` at an external site-config folder (typically a live clone of ftr-site-config). This PR eliminates that manual edit by introducing a `WALLABAG_SITE_CONFIG_FOLDERS` environment variable.

When the variable is unset or empty the behaviour is identical to before: Graby uses only its bundled config. When set to a directory path, that path is passed to `ConfigBuilder` as the `site_config` value — exactly as the wiki currently instructs users to do by hand.

Two Symfony DI features work together: `env(WALLABAG_SITE_CONFIG_FOLDERS): ''` in the `parameters` block provides a safe empty-string default so the variable is optional, and a `%env()%`-backed named parameter exposes it to a `@=` Expression Language expression on the `ConfigBuilder` argument. The compiled container resolves this to a `getEnv()` call at runtime, not at `cache:clear` time.

### DIFF
--- a/app/config/services.yml
+++ b/app/config/services.yml
@@ -5,6 +5,8 @@ imports:
 
 parameters:
     spiriit_form_filter.get_filter.doctrine_orm.class: Wallabag\Event\Subscriber\CustomDoctrineORMSubscriber
+    env(WALLABAG_SITE_CONFIG_FOLDERS): ''
+    wallabag.site_config_folders: '%env(csv:WALLABAG_SITE_CONFIG_FOLDERS)%'
 
 services:
     _defaults:
@@ -225,7 +227,7 @@ services:
 
     Graby\SiteConfig\ConfigBuilder:
         arguments:
-            $config: {}
+            $config: "@=parameter('wallabag.site_config_folders') and parameter('wallabag.site_config_folders')[0] ? {'site_config': parameter('wallabag.site_config_folders')} : {}"
 
     Wallabag\SiteConfig\GrabySiteConfigBuilder:
         tags:


yguedidi:
@HolgerAusB you may like this one ☺️ 

Could you test it on you instance and confirm it works as expected? Thanks

HolgerAusB:
I'll defiantly love it, thank you @yguedidi . But I am not sure, if I have the skills to test this, as I am not a developer/programmer.

Changing 3 lines in services.yml of my current release version should not be the problem.
But how would I start wallabag now (GIT install/Apache2)? Where do I define the env, so it will be persistent over the next wallabag release?

Is the env able to contain a list (or is it named an array) of folders? Do I have to use square brackets then:
`WALLABAG_SITE_CONFIG_FOLDER=['/var/www/path_to/ftr/site_config/standard', '/var/www/path_to/ftr/site_config/custom']`

HolgerAusB:
Thanks to AI, I was able to test this.

Adding this line to `/etc/apache2/sites-available/wallabag.conf`:
```
SetEnv WALLABAG_SITE_CONFIG_FOLDER "['/var/www/path_to/ftr/site_config/standard', '/var/www/path_to/ftr/site_config/custom']"
```

I changed the author selector for one site config file to a literal test string in `.../standard` and `.../custom` folder. And as expected, the last folder in the list/array wins. It worked. Thank you so much, @yguedidi ❤️💕❤️

This apache2 config should go to the documentation.

yguedidi:
@HolgerAusB thanks for the quick answer!

On Apache2, to set an env var I think you can use the [`SetEnv` directive](https://httpd.apache.org/docs/current/en/mod/mod_env.html) in your apache configuration file, where you define the wallabag host configuration

> Is the env able to contain a list (or is it named an array) of folders?

Was not, but just updated the PR to normally support multiple folders!
The env var supports a comma-separated list of folders now.
Not that I renamed the variable to WALLABAG_SITE_CONFIG_FOLDERS (with the trainling "S")

In your case I think you should add this to your apache config, for now:
```
SetEnv WALLABAG_SITE_CONFIG_FOLDERS /var/www/path_to/ftr/site_config/standard,/var/www/path_to/ftr/site_config/custom
```

yguedidi:
@HolgerAusB note that I also changed the parameter in services.yml to `wallabag.site_config_folders: '%env(csv:WALLABAG_SITE_CONFIG_FOLDERS)%'` (adding the `csv:` part) to support multiple folders

HolgerAusB:
Two destination folders worked already with the first version and it is also working with the changes for `csv:` and the trainling "S"

yguedidi:
@HolgerAusB updated version that makes tests pass, could you test it?

Also, to be sure the support for multiple folders works, could you have a site config only in a folder and another in another folder, and test that both are taken into account? Thanks in advance 🙂

yguedidi:
we commented at the same time haha, happy that it works! Thanks for testing it

HolgerAusB:
The last folder in the list which contains a domain matching config wins. That was the behavior in my wiki solution and it is also the same with your patch. Configs never merging together (and should not). So everything is fine for me.

PR #8755, opened 2026-03-20
Fix XDebug connectivity in Docker and switch to trigger mode

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

XDebug was unable to connect to the IDE debugger because `xdebug.client_host` defaults to `localhost`, which inside a Docker container resolves to the container itself rather than the host machine. Every PHP invocation emitted "Could not connect to debugging client. Tried: localhost:9003".

## What changed

- **`compose.yaml`**: add `extra_hosts: host.docker.internal:host-gateway` to the PHP service so the container can resolve the Docker host IP on all platforms (Docker Desktop, native Linux Docker, WSL2). Also remove the stale `# - 9000:9000` port binding comment — XDebug 3 makes outbound connections from the container to the IDE, so no port publishing is needed.
- **`docker/php/xdebug`**: set `client_host=host.docker.internal` in `XDEBUG_CONFIG` and switch to trigger mode (`start_with_request=trigger`). Remove `XDEBUG_TRIGGER=yes` — that line set the env var unconditionally, which would have caused XDebug to always fire even in trigger mode, defeating its purpose.

## Trigger mode

With `start_with_request=trigger`, XDebug only initiates a debug session when explicitly requested — no more terminal noise on every PHP invocation.

- **Web**: use the [Xdebug Helper by JetBrains](https://chromewebstore.google.com/detail/xdebug-helper-by-jetbrain/aoelhdemabeimdhedkidlnbkfhnhgnhm) browser extension, or append `?XDEBUG_TRIGGER=1` to any URL.
- **CLI**: prefix the command with `XDEBUG_TRIGGER=1`: `docker compose exec php env XDEBUG_TRIGGER=1 php bin/console …`

### DIFF
--- a/docker/php/xdebug
+++ b/docker/php/xdebug
@@ -1,3 +1,2 @@
 XDEBUG_MODE=debug
-XDEBUG_TRIGGER=yes
-XDEBUG_CONFIG=client_host=host.docker.internal
+XDEBUG_CONFIG=client_host=host.docker.internal start_with_request=trigger


yguedidi:
~needs more testing 😅~ my bad, forgot to start listening on VSCode... 🤦🏼

yguedidi:
thanks @j0k3r ! tested locally and it works, managed to set breakpoints and have the execution to stop on them

PR #8767, opened 2026-03-26
Make compromised-password validation opt-in

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

This PR makes wallabag's compromised-password validation opt-in. Admins can enable the HIBP-backed check with `WALLABAG_NO_COMPROMISED_PASSWORDS=1`; when the variable is absent, wallabag skips the extra network call and leaves password validation unchanged.

The same opt-in flag now controls `NewUserType`, `ChangePasswordType`, and the production installer. The committed development `.env` enables the check for contributors, and the admin parameters documentation explains how to enable it in real deployments.

When the check is enabled, wallabag still uses Symfony's `NotCompromisedPassword` constraint with `skipOnError: true`, so temporary HIBP failures do not block offline or CI workflows.


Kdecherf (changes requested):
LGTM on the implementation side

However two mandatory changes:

- we need to add a note in the doc about this new security feature and how admin can configure it
- we need to keep it configurable for admin who would like to avoid additional network calls for this kind of feature, I would see this feature as disabled by default

yguedidi:
> we need to add a note in the doc about this new security feature and how admin can configure it

seems not possible in the end, for now, to not deploy the change..

> we need to keep it configurable for admin who would like to avoid additional network calls for this kind of feature, I would see this feature as disabled by default

@nicosomb @j0k3r any opinion here? to me as it's security related I'd tend to prefer having it enabled by default but allow admins to disable it if they really really want

yguedidi:
> we need to keep it configurable for admin who would like to avoid additional network calls for this kind of feature

@Kdecherf FYI I'll wait for #8776 to be merge to introduce the new setting 🙁

yguedidi:
@Kdecherf PR updated! you can check this [force-push diff](https://github.com/wallabag/wallabag/compare/f18dc5524507543d9209153bf71a80968d5bf53f..8c494795f18127d4aa56f1a9639d03ae8667139c) to see all changes since last time you reviewed, or do a full review as it's small enough :)

nicosomb:
I love this feature. Well done!

PR #8775, opened 2026-03-31
Update 3rd party libraries

### LINKED ISSUE #5689: Update about page
The `/about` page is quite old and some things need to be updated: 
* 3rd party libraries: we have to update the list 
* the link on "helping wallabag" tab (an issue lists all our needs) is wrong
* the "who is behind wallabag" tab is old: my website is wrong ... and @Kdecherf is missing.

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fix #5689

### DIFF
--- a/templates/Static/about.html.twig
+++ b/templates/Static/about.html.twig
@@ -54,34 +54,60 @@
                                 <th>{{ 'about.third_party.package'|trans }}</th>
                                 <th>{{ 'about.third_party.license'|trans }}</th>
                             </tr>
-                            <tr><td>behat/transliterator</td><td>Artistic 1.0</td></tr>
-                            <tr><td>CraueConfigBundle</td><td>MIT</td></tr>
+                            <tr><td>babdev/pagerfanta-bundle</td><td>MIT</td></tr>
+                            <tr><td>beberlei/assert</td><td>BSD-2-Clause</td></tr>
+                            <tr><td>behat/transliterator</td><td>Artistic-1.0</td></tr>
+                            <tr><td>clue/ndjson-react</td><td>MIT</td></tr>
+                            <tr><td>clue/stream-filter</td><td>MIT</td></tr>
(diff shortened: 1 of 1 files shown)


j0k3r (approved):

yguedidi (changes requested):
for maintainability I would instead generate this list from composer.json and package.json directly in some ways, so that we don't have to maintain this static list manually.
we can warm up the computed list in the cache in case it's heavy to compute but may be not needed

nicosomb:
Your idea is good (i had it but I thought it was overkill). 

But in fact, I wonder if we have to keep this section ... is that necessary to have it? Is that useful? 

Maybe we can remove it.

yguedidi:
Maybe we can keep it but making it point to the composer.lock and yarn.lock files on github, ideally taking care of the right tag of the current version in the URL?

Kdecherf:
Hello @nicosomb @yguedidi,

I may have an idea in between.

What about replacing this list with a link to a tag-based documentation page on the github repository, for example https://github.com/wallabag/wallabag/blob/master/doc/content/developer/_index.md, replacing _master_ with a release tag, and _index.md with the actual page we would like to put).

Then, the documentation page could be automatically generated when creating a new release, by extracting packages from the lock files.

Any thoughts?

yguedidi:
@Kdecherf first thoughts is that if anyway we have to build an extraction logic, then I'd prefer to have it in the app like now.

Kdecherf:
@yguedidi what would be the value of having this logic within the app instead of the documentation?

yguedidi:
@Kdecherf I'd say consistent with current location, but the biggest value to me would be to not have to manage multiple versions of that doc, as it will come directly from the wallabag instance of the user at the version they use

PR #8778, opened 2026-04-04
Add a new configuration to change the homepage view

### LINKED ISSUE #8347: Changing default view from Unread to All Entries
Asked here https://www.reddit.com/r/wallabag/comments/1lk0nlh/changing_default_view_from_unread_to_all_entries/ and it sounds a good idea

We could have a setting to define what is the view displayed when we want to go back to the homepage (unread / starred / archive / whatever).

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| CHANGELOG.md  | no
| License       | MIT

Fixes #8347

### DIFF
--- a/migrations/Version20260331213851.php
+++ b/migrations/Version20260331213851.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Application\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Wallabag\Doctrine\WallabagMigration;
+
+/**
+ * Add default_homepage column to config table.
+ */
+final class Version20260331213851 extends WallabagMigration
+{
+    public function up(Schema $schema): void
+    {
+        $configTable = $schema->getTable($this->getTable('config'));
+
+        $this->skipIf($configTable->hasColumn('default_homepage'), 'It seems that you already played this migration.');
+
+        $configTable->addColumn('default_homepage', 'string', [
+            'notnull' => true,
+            'default' => 'unread',
+        ]);
+    }
+
+    public function down(Schema $schema): void
+    {
+        $schema->getTable($this->getTable('config'))->dropColumn('default_homepage');
+    }
+}
--- a/migrations/Version20260331213851.php
+++ b/migrations/Version20260331213851.php
@@ -1,11 +1,13 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Application\Migrations;
 
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\Schema;
 use Wallabag\Doctrine\WallabagMigration;
+use Wallabag\Enum\HomepageTarget;
 
 /**
  * Add default_homepage column to config table.
@@ -18,10 +20,21 @@ final class Version20260331213851 extends WallabagMigration
 
         $this->skipIf($configTable->hasColumn('default_homepage'), 'It seems that you already played this migration.');
 
-        $configTable->addColumn('default_homepage', 'string', [
-            'notnull' => true,
-            'default' => 'unread',
-        ]);
+        $table = $this->getTable('config');
+        $platform = $this->connection->getDatabasePlatform();
+
+        if ($platform instanceof SqlitePlatform) {
+            $this->addSql('ALTER TABLE ' . $table . ' ADD COLUMN default_homepage VARCHAR(255) NOT NULL DEFAULT \'' . HomepageTarget::Unread->value . '\'');
+        } else {
+            $this->addSql('ALTER TABLE ' . $table . ' ADD COLUMN default_homepage VARCHAR(255)');
+            $this->addSql('UPDATE ' . $table . ' SET default_homepage = \'' . HomepageTarget::Unread->value . '\'');
+
+            if ($platform instanceof PostgreSQLPlatform) {
+                $this->addSql('ALTER TABLE ' . $table . ' ALTER COLUMN default_homepage SET NOT NULL, ALTER COLUMN default_homepage SET DEFAULT \'' . HomepageTarget::Unread->value . '\'');
+            } elseif ($platform instanceof MySQLPlatform) {
+                $this->addSql('ALTER TABLE ' . $table . ' MODIFY COLUMN default_homepage VARCHAR(255) NOT NULL DEFAULT \'' . HomepageTarget::Unread->value . '\'');
+            }
+        }
     }
 
     public function down(Schema $schema): void
--- a/migrations/Version20260331213851.php
+++ b/migrations/Version20260331213851.php
@@ -2,9 +2,6 @@
 
 namespace Application\Migrations;
 
-use Doctrine\DBAL\Platforms\MySQLPlatform;
-use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
-use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\Schema;
 use Wallabag\Doctrine\WallabagMigration;
 use Wallabag\Enum\HomepageTarget;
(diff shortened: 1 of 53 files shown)


yguedidi (changes requested):

src/Entity/Config.php
```diff
@@ -130,6 +131,9 @@ class Config
     #[ORM\Column(name: 'custom_css', type: 'text', nullable: true)]
     private $customCSS;
 
+    #[ORM\Column(name: 'default_homepage', type: 'string', enumType: HomepageTarget::class, options: ['default' => 'unread'])]
```
yguedidi: you can use the value of the enum case instead of hardcoded value

src/Enum/HomepageTarget.php
```diff
@@ -0,0 +1,17 @@
...
+{
+    case Unread = 'unread';
+    case All = 'all';
+    case Archive = 'archive';
+    case Starred = 'starred';
+    case Tags = 'tags';
+
+    public function label(): string
+    {
+        return 'config.form_settings.default_homepage.' . $this->value;
```
yguedidi: a match case enumerating all options is better IMO:
- PHPStan will help that we really cover all cases
- the full translation key will become searchable

migrations/Version20260331213851.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function up(Schema $schema): void
+    {
+        $configTable = $schema->getTable($this->getTable('config'));
+
+        $this->skipIf($configTable->hasColumn('default_homepage'), 'It seems that you already played this migration.');
+
+        $configTable->addColumn('default_homepage', 'string', [
+            'notnull' => true,
+            'default' => 'unread',
```
yguedidi: as we use not null, I feel it's better to not rely on the SQL default. that's why I suggested the 3 queries technique :)

yguedidi (changes requested):

src/Controller/EntryController.php
```diff
@@ -47,6 +48,27 @@ public function __construct(
...
+            $request->attributes->set('_route_params', ['page' => 1]);
+        }
+
+        return match ($defaultHomepage) {
+            UserConfig::DEFAULT_HOMEPAGE_ALL => $this->showAllAction($request, 1),
+            UserConfig::DEFAULT_HOMEPAGE_ARCHIVE => $this->showArchiveAction($request, 1),
+            UserConfig::DEFAULT_HOMEPAGE_STARRED => $this->showStarredAction($request, 1),
+            UserConfig::DEFAULT_HOMEPAGE_TAGS => $this->redirectToRoute('tag'),
+            default => $this->showUnreadAction($request, 1),
+        };
```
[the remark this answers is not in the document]
Kdecherf: I second that @yguedidi, a redirection would be more consistent I think
cc @nicosomb

translations/messages.fr.yml
```diff
@@ -30,7 +30,7 @@ menu:
         about: À propos
         search: Recherche
         save_link: Sauvegarder un nouvel article
-        back_to_unread: Retour aux articles non lus
+        back_to_homepage: Retour à l’accueil
```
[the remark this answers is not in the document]
Kdecherf: I think we should stick with the normal quote instead of this one

yguedidi (changes requested):
@nicosomb looks like my last review round was marked as resolved but it seems there is no code change, maybe you forgot to push? 😅

migrations/Version20260331213851.php
```diff
@@ -0,0 +1,44 @@
...
+        if ($platform instanceof SqlitePlatform) {
+            $this->addSql('ALTER TABLE ' . $table . ' ADD COLUMN default_homepage VARCHAR(255) NOT NULL DEFAULT \'' . HomepageTarget::Unread->value . '\'');
+        } else {
+            $this->addSql('ALTER TABLE ' . $table . ' ADD COLUMN default_homepage VARCHAR(255)');
+            $this->addSql('UPDATE ' . $table . ' SET default_homepage = \'' . HomepageTarget::Unread->value . '\'');
+
+            if ($platform instanceof PostgreSQLPlatform) {
+                $this->addSql('ALTER TABLE ' . $table . ' ALTER COLUMN default_homepage SET NOT NULL, ALTER COLUMN default_homepage SET DEFAULT \'' . HomepageTarget::Unread->value . '\'');
+            } elseif ($platform instanceof MySQLPlatform) {
+                $this->addSql('ALTER TABLE ' . $table . ' MODIFY COLUMN default_homepage VARCHAR(255) NOT NULL DEFAULT \'' . HomepageTarget::Unread->value . '\'');
```
yguedidi: I would not add a default value at the DB level

src/Entity/Config.php
```diff
@@ -130,6 +131,9 @@ class Config
     #[ORM\Column(name: 'custom_css', type: 'text', nullable: true)]
     private $customCSS;
 
+    #[ORM\Column(name: 'default_homepage', type: 'string', enumType: HomepageTarget::class, options: ['default' => HomepageTarget::Unread->value])]
```
yguedidi: I'd not add the default value at DB level

src/Controller/EntryController.php
```diff
@@ -47,6 +48,21 @@ public function __construct(
...
+    {
+        $defaultHomepage = $this->getUser()->getConfig()->getDefaultHomepage();
+
+        return match ($defaultHomepage) {
+            HomepageTarget::Unread => $this->redirectToRoute('unread'),
+            HomepageTarget::All => $this->redirectToRoute('all'),
+            HomepageTarget::Archive => $this->redirectToRoute('archive'),
+            HomepageTarget::Starred => $this->redirectToRoute('starred'),
+            HomepageTarget::Tags => $this->redirectToRoute('tag'),
+        };
```
yguedidi: better be in the enum as a `route` method that returns the route name (similar to `label`)

src/Helper/Redirect.php
```diff
@@ -43,17 +44,28 @@ public function to($url, $ignoreActionMarkAsRead = false)
...
+    private function generateHomepageUrl(User $user): string
+    {
+        return match ($user->getConfig()->getDefaultHomepage()) {
+            HomepageTarget::Unread => $this->router->generate('unread'),
+            HomepageTarget::All => $this->router->generate('all'),
+            HomepageTarget::Archive => $this->router->generate('archive'),
+            HomepageTarget::Starred => $this->router->generate('starred'),
+            HomepageTarget::Tags => $this->router->generate('tag'),
+        };
+    }
```
yguedidi, line 63: will not be needed anymore with a `route` method in the enum

translations/messages.fr.yml
```diff
@@ -82,6 +82,14 @@ config:
             label: Que faire lorsqu'un article est supprimé, marqué comme lu ou marqué comme favoris ?
             redirect_homepage: Retourner à la page d’accueil
             redirect_current_page: Rester sur la page actuelle
+        default_homepage:
+            label: Page d’accueil par défaut
```
yguedidi: use normal single quote

translations/messages.fr.yml
```diff
@@ -82,6 +82,14 @@ config:
...
+            unread: Non lus
+            all: Tous les articles
+            archive: Archives
+            starred: Favoris
+            tags: Tags
+        help_default_homepage: Choisissez la page affichée à l’ouverture de wallabag.
```
yguedidi: use normal single quote

src/Controller/AbstractController.php
```diff
@@ -3,10 +3,27 @@
...
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Wallabag\Entity\User;
+use Wallabag\Helper\Redirect;
 
 abstract class AbstractController extends BaseAbstractController
 {
+    public static function getSubscribedServices(): array
+    {
+        return array_merge(parent::getSubscribedServices(), [
+            'Wallabag\Helper\Redirect' => Redirect::class,
```
yguedidi: couldn't it just be
```suggestion
            Redirect::class,
```

yguedidi (changes requested):

migrations/Version20260331213851.php
```diff
@@ -0,0 +1,39 @@
...
+final class Version20260331213851 extends WallabagMigration
+{
+    public function up(Schema $schema): void
+    {
+        $configTable = $schema->getTable($this->getTable('config'));
+
+        $this->skipIf($configTable->hasColumn('default_homepage'), 'It seems that you already played this migration.');
+
+        $configTable->addColumn('default_homepage', 'string', [
+            'notnull' => false,
```
yguedidi, line 21: why going back to nullable?
any thing wrong with the pattern:
- add column nullabale no default
- update to set a value
- alter column to notnull

to me, final state is not nullable column, and not nullable php property.
btw, I'm not fan of DB default values as I think it's better to enforce defining the values to set in the PHP code, not relying on a hidden DB internal behavior

src/Entity/Config.php
```diff
@@ -130,6 +131,9 @@ class Config
     #[ORM\Column(name: 'custom_css', type: 'text', nullable: true)]
     private $customCSS;
 
+    #[ORM\Column(name: 'default_homepage', type: 'string', enumType: HomepageTarget::class, nullable: true)]
+    private ?HomepageTarget $defaultHomepage = HomepageTarget::Unread;
```
yguedidi, line 135:
```suggestion
    #[ORM\Column(name: 'default_homepage', type: 'string', enumType: HomepageTarget::class)]
    private HomepageTarget $defaultHomepage = HomepageTarget::Unread;
```

src/Entity/Config.php
```diff
@@ -463,6 +467,18 @@ public function setCustomCSS(?string $customCSS): self
         return $this;
     }
 
+    public function getDefaultHomepage(): HomepageTarget
+    {
+        return $this->defaultHomepage ?? HomepageTarget::Unread;
```
yguedidi, line 472:
```suggestion
        return $this->defaultHomepage;
```

yguedidi (changes requested):
I think you can squash directly after this review round

PR #8804, opened 2026-04-22
Bump to Node 24

j0k3r:
Bump `@symfony/webpack-encore` & `webpack-cli` as well.
Remove `webpack-manifest-plugin` because we don't use directly.
This will also fix some Dependabot alerts.


yguedidi (approved):

yguedidi:
No other places where the Node version 20 is mentioned?

j0k3r:
@yguedidi I didn't find it in the doc not on docker

Kdecherf:
For the record there was one missing reference in `docker/php/Dockerfile`, addressed by #8812

PR #8812, opened 2026-05-02
Update development Dockerfile

Kdecherf:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | 
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

- Update php 8.2 base image to debian trixie
- Update nodejs from 20 to 24
- Update PECL libs redis, imagick and xdebug to latest version

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -1,16 +1,16 @@
-FROM golang as envsubst
+FROM golang AS envsubst
 
-ARG ENVSUBST_VERSION=v1.3.0
+ARG ENVSUBST_VERSION=v1.4.3
 
 # envsubst from gettext can not replace env vars with default values
 # this package is not available for ARM32 and we have to build it from source code
 # flag -ldflags "-s -w" produces a smaller executable
 RUN go install -ldflags "-s -w" -v github.com/a8m/envsubst/cmd/envsubst@${ENVSUBST_VERSION}
 
-FROM php:8.2-fpm-bookworm AS rootless
+FROM php:8.2-fpm-trixie AS rootless
 
 ARG DEBIAN_FRONTEND=noninteractive
-ARG NODE_VERSION=20
+ARG NODE_VERSION=24
 
 RUN apt-get update \
     && apt-get install -y \
@@ -18,8 +18,7 @@ RUN apt-get update \
       curl \
       gnupg \
       lsb-release \
-      openssl \
-      software-properties-common
+      openssl
 
 RUN curl -fsSL https://deb.nodesource.com/setup_${NODE_VERSION}.x | bash -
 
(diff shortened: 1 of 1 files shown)


docker/php/Dockerfile
```diff
@@ -1,25 +1,24 @@
...
-ARG ENVSUBST_VERSION=v1.3.0
+ARG ENVSUBST_VERSION=v1.4.3
 
 # envsubst from gettext can not replace env vars with default values
 # this package is not available for ARM32 and we have to build it from source code
 # flag -ldflags "-s -w" produces a smaller executable
 RUN go install -ldflags "-s -w" -v github.com/a8m/envsubst/cmd/envsubst@${ENVSUBST_VERSION}
 
-FROM php:8.2-fpm-bookworm AS rootless
+FROM php:8.4-fpm-trixie AS rootless
```
yguedidi: I'd not update this. I think the dev env should be aligned with the minimum supported version, and as of now it's 8.2.
We can of course maybe reopen the discussion to bump the minimum version to 8.4 😁

yguedidi (changes requested):

docker/php/Dockerfile
Kdecherf: I have mixed feelings on this, I'm more on the side of keeping the dev docker image aligned with the prod container for the php version.

If your concern is the early detection of compatibility issues, the CI workflow should be enough IMO.

yguedidi: Initially I had that concern about compatibility but then indeed I thought the CI will catch issues.
Still to me it feels more "safe" to contribute on the lowest supported version, that avoids enjoying new PHP features then have to fix issues from the CI, with a potential frustration.
IMO, unless we bump the PHP minimum version, I'd prefer to keep the 8.2 for dev.
But it's a preference, will not block the change if others are OK. ping @nicosomb @j0k3r

j0k3r: I agree with @yguedidi about working on the lowest version (even if I use 8.5 locally 🫣)

Kdecherf: Understood, I'm change that, thanks for your input

Kdecherf: 507deb44d

j0k3r (approved):

docker/php/Dockerfile
```diff
@@ -59,9 +58,9 @@ RUN docker-php-ext-install -j "$(nproc)" \
         tidy \
         zip
 
-RUN pecl install redis-6.1.0 \
-    && pecl install imagick-3.7.0 \
-    && pecl install xdebug-3.4.1 \
+RUN pecl install redis-6.3.0 \
+    && pecl install imagick-3.8.1 \
+    && pecl install xdebug-3.5.1 \
```
yguedidi, line 63: have you checked those new versions are OK with php 8.2? if yes then fine for me

yguedidi (changes requested):

docker/php/Dockerfile
Kdecherf: Yes, build and run are fine with this version

yguedidi (approved):

Issue #8752, opened 2026-03-19
Soft deletion policy for Entry


casimir:
**Is your feature request related to a problem? Please describe.**

This issue is a follow-up to #8247. Deleted entries are deleted for good and that creates several issues like 3rd party applications sync (#2906), no trash bin feature (#6635), prevent accidental deletions (#6856).

**Describe the solution you'd like**

Several solutions have already been discussed but it always converged on the use of a deleted timestamp. Let's say it will be `Entry.deleted_at`, a nullable datetime field.

We already talked about this high level plan:
 
1. Add the new field (easy part) and update all the SELECT query  to use it (hard part)
2. Expose the new field in the API
3. Update the deletion code to soft delete (set `deleted_at`)
4. Add a console command to purge deleted entries older than a given date (new configuration attribute?)
5. Implement new features like the recycle bin.

As an aside, I am only interested in the 3rd party apps optimizations here (1-4). 

This split is still subject to discussion, it's merely a report of what has been said in #8247. What are your thoughts? How should I proceed to propose PRs?

j0k3r:
This is a good candidate for the 2.7.0.

You can start working on points 1, 2 and 3. You can do points 4 and 5 later (separately, if you like).

About the API, entries flagged as _deleted_ shouldn't show up in any list requests unless maybe we introduce a `include_deleted` boolean to retrieve them. But for BC, we should exclude them by default.

What should be handle as well is when someone submit an url which is already in a deleted entry. The unique condition will fail and the url addition will fail.
How the API / UI should behave?
- Should we return the `id` of the deleted entry in the API?
- Should we just display a message saying the url is already in the database but on a deleted entry?
- Should we propose the ability to restore it (before having the trash feature)?

yguedidi:
> What should be handle as well is when someone submit an url which is already in a deleted entry.

I would suggest that, if I try to think about the users intention when adding a URL, they would expect it to be last added in unread.
Do I'd suggest to silently undelete the entry and update its creation date so it appears first.
What do you think?

j0k3r:
Yep good idea. But maybe that behavior should at least be written somewhere in the doc.

casimir:
I tried to fit everything into a more formal spec (and I like diagrams). I think I can start working on the _Read_ part. WDYT?

---

The main goal is to introduce the notion of soft-deletion for entries. The obvious is to add `deleted_at` on `Entry`.

## Changes on state management

[...]

PR #8818, opened 2026-05-05
Entry soft-deletion (part 1&2)

casimir:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

First parts of #8752. I originally intended to follow the 3 PR plan from the issue but as implementation went it felt wrong to separate the deletion and the restoration. See details in the parent issue.

The gist of the changes are the new methods on the Entry entity: `isDeleted()` and `updateDeleted(bool)`. There is a lot of mechanical changes in the tests to adapt them to the new behavior.
I took into consideration the webapp, the API and the different importers. It should cover all the way to save an entry.

I wasn't sure about what fields to strip on deletion and where. I went for the pragmatic approach: `content` to optimize storage, `pictureUrl` to help me visualize state during my tests, in `updateDeleted(bool)` for a unique logic. That's easy to change. 

Another PR will follow soon with the expiration limit setting and purge command.

### DIFF
--- a/src/Command/CleanDuplicatesCommand.php
+++ b/src/Command/CleanDuplicatesCommand.php
@@ -93,7 +93,7 @@ class CleanDuplicatesCommand extends Command
                 // entry deleted, dispatch event about it!
                 $this->eventDispatcher->dispatch(new EntryDeletedEvent($entryToDelete), EntryDeletedEvent::NAME);
 
-                $this->entityManager->remove($entryToDelete);
+                $entryToDelete->softDelete();
                 $this->entityManager->flush(); // Flushing at the end of the loop would require the instance not being online
             } else {
                 $urls[] = $entry['url'];
--- a/src/Command/CleanDuplicatesCommand.php
+++ b/src/Command/CleanDuplicatesCommand.php
@@ -93,7 +93,7 @@ class CleanDuplicatesCommand extends Command
                 // entry deleted, dispatch event about it!
                 $this->eventDispatcher->dispatch(new EntryDeletedEvent($entryToDelete), EntryDeletedEvent::NAME);
 
-                $entryToDelete->softDelete();
+                $entryToDelete->setContent(null)->setPreviewPicture(null)->updateDeleted(true);
                 $this->entityManager->flush(); // Flushing at the end of the loop would require the instance not being online
             } else {
                 $urls[] = $entry['url'];
--- a/src/Command/CleanDuplicatesCommand.php
+++ b/src/Command/CleanDuplicatesCommand.php
@@ -93,7 +93,7 @@ class CleanDuplicatesCommand extends Command
                 // entry deleted, dispatch event about it!
                 $this->eventDispatcher->dispatch(new EntryDeletedEvent($entryToDelete), EntryDeletedEvent::NAME);
 
-                $entryToDelete->setContent(null)->setPreviewPicture(null)->updateDeleted(true);
+                $entryToDelete->updateDeleted(true);
                 $this->entityManager->flush(); // Flushing at the end of the loop would require the instance not being online
             } else {
                 $urls[] = $entry['url'];
--- a/src/Command/CleanDuplicatesCommand.php
+++ b/src/Command/CleanDuplicatesCommand.php
@@ -93,7 +93,7 @@ class CleanDuplicatesCommand extends Command
                 // entry deleted, dispatch event about it!
                 $this->eventDispatcher->dispatch(new EntryDeletedEvent($entryToDelete), EntryDeletedEvent::NAME);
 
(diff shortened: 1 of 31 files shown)


src/Entity/Entry.php
```diff
@@ -311,6 +315,27 @@ public function getUrl()
...
+    {
+        return $this->deletedAt;
+    }
+
+    public function isDeleted(): bool
+    {
+        return null !== $this->deletedAt;
+    }
+
+    public function updateDeleted(bool $isDeleted): static
```
j0k3r, line 328: I would rather use `delete()` and `restore()` (if allowed)

src/Helper/ContentProxy.php
```diff
@@ -42,6 +42,8 @@ public function __construct(
      */
     public function updateEntry(Entry $entry, $url, array $content = [], $disableContentUpdate = false): void
     {
+        $entry->updateDeleted(false);
```
j0k3r, line 45:
```suggestion
        // no mater previous state, if an entry got updated, we restore it
        $entry->updateDeleted(false);
```

src/Entity/Entry.php
```diff
@@ -311,6 +315,27 @@ public function getUrl()
...
+    {
+        $this->deletedAt = $isDeleted ? new \DateTimeImmutable() : null;
+        if ($isDeleted) {
+            $this->content = null;
+            $this->previewPicture = null;
```
j0k3r, line 333: Otherwise a public sharing content will still be available.
Maybe you can also add a condition in the `shareEntryAction` Controller
```suggestion
            $this->previewPicture = null;
            $this->uid = null;
```

src/Controller/Api/EntryRestController.php
```diff
@@ -327,6 +327,7 @@ public function getEntriesAction(Request $request, EntryRepository $entryReposit
         $domainName = (null === $request->query->get('domain_name')) ? '' : (string) $request->query->get('domain_name');
         $httpStatus = (!\array_key_exists((int) $request->query->get('http_status'), Response::$statusTexts)) ? null : (int) $request->query->get('http_status');
         $hasAnnotations = (null === $request->query->get('annotations')) ? null : (bool) $request->query->get('annotations');
+        $includeDeleted = (null === $request->query->get('include_deleted')) ? false : (bool) $request->query->get('include_deleted');
```
j0k3r, line 330: Better use `getBoolean` (to be validated)
```suggestion
        $includeDeleted = (null === $request->query->getBoolean('include_deleted')) ? false : (bool) $request->query->getBoolean('include_deleted');
```

src/Command/CleanDuplicatesCommand.php
```diff
@@ -93,7 +93,7 @@ private function cleanDuplicates(User $user): void
                 // entry deleted, dispatch event about it!
                 $this->eventDispatcher->dispatch(new EntryDeletedEvent($entryToDelete), EntryDeletedEvent::NAME);
 
-                $this->entityManager->remove($entryToDelete);
+                $entryToDelete->updateDeleted(true);
```
j0k3r, line 96: As you are retrieving all urls (deleted or not) you might re-deleted a deleted entry. Avoid that work and filter them on line 80

j0k3r (commented):
Also, the `wallabag:entry:reload` command get all entries (deleted or not) and then call the `updateEntry` method which will undeleted all deleted entries.
Maybe the command should only retrieve undeleted entries instead.

Kdecherf:
Hello @casimir,

Would you be able to rebase your PR, or would you like some help doing so?

src/Command/CleanDuplicatesCommand.php
casimir: Fixed in 24c4f7001641b73d578a7763d6790ca2e2e15d78

src/Entity/Entry.php
casimir: The action is already guarded by the URL, a `null` `uid` doesn't match and ends with a 404.

casimir: Just for the peace of mind, I added a test too, see 2a0a9cbcddf24ecbfbe7f39a64dad0f13a2edaa3.

src/Entity/Entry.php
casimir: Done.

casimir:
@Kdecherf I should be able to manage, I'll ping you if not.

@j0k3r I addressed all your feedback. I'll hold some time before rebasing so you can validate that all you concerns are fixed.

casimir:
Rebase done.

PR #8883, opened 2026-06-15
Fix flaky MondeDiplo test

j0k3r:
Looks like when we fetch `monde-diplomatique.fr`, sometimes the request fails. When it fails, no title are extracted so we try to find a default title (`ContentProxy`->`setDefaultEntryTitle`) with became `57476` instead of `Quand Manille manœuvre`. See https://3v4l.org/tiUMt

To avoid that random error, we now mock the HTTP call to the website.
In addition, add a unit test login data are properly passed.


nicosomb:
Thank you @j0k3r !

nicosomb (approved):

PR #8921, opened 2026-07-02
Fix API origin_url unset

### LINKED ISSUE #8920: Cannot unset field via API
The REST API does not properly process requests to unset a field.

For example: if the `origin_url` field is set to "example.com" and we would like to remove this value, the following should unset the field:

```sh
curl "$HOST/api/entries/$ID?access_token=$TOKEN&detail=metadata" \
     -X PATCH -d '{"origin_url": ""}' \
     -H 'Content-Type: application/json' | jq '.origin_url'
```
But this request has no effect. It returns status 200, but the `origin_url` field is still returned as "example.com" and remains unchanged in the database.

wallabag 2.6.14

nicosomb:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Should fix https://github.com/wallabag/wallabag/issues/8920

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -753,8 +753,8 @@ class EntryRestController extends WallabagRestController
             $tagsAssigner->assignTagsToEntry($entry, $data['tags']);
         }
 
-        if (!empty($data['origin_url'])) {
-            $entry->setOriginUrl($data['origin_url']);
+        if (\array_key_exists('origin_url', $data)) {
+            $entry->setOriginUrl('' === $data['origin_url'] ? null : $data['origin_url']);
         }
 
         if (null !== $data['isPublic']) {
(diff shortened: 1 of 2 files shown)


src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -1407,7 +1407,12 @@ private function retrieveValueFromRequest(Request $request)
             'picture' => $request->request->get('preview_picture'),
             'publishedAt' => $request->request->get('published_at'),
             'authors' => $request->request->get('authors', ''),
-            'origin_url' => $request->request->get('origin_url', ''),
         ];
+
+        if ($request->request->has('origin_url')) {
```
yguedidi, line 1412: I think this should be done for all optional parameters, to prevent such bug report for another field later.

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -1407,7 +1407,12 @@ private function retrieveValueFromRequest(Request $request)
...
+            $data['origin_url'] = $request->request->get('origin_url');
```
yguedidi, line 1413: I think retrieveValueFromRequest should clean the values directly, so that:
- not having the key meand leave as is
- have the key with a value means change to that value
- have the key with null means clear the value (so convert input empty string to null)
I think this can apply to other fields too, so that in places using retrieveValueFromRequest we know that if the key exists and it's value === null then we have to clear the field

yguedidi (changes requested):

PR #8927, opened 2026-07-09
Add parameter `detail` to the entry endpoint

### LINKED ISSUE #8901: Implement metadata-only `detail` for /api/entries/ID endpoints
The `/api/entries` endpoint takes an optional `detail`=`metadata` parameter, which omits the `content` field in the response data. This dramatically reduces data transfer and significantly improves performance.

The `detail` parameter is not currently respected by the `/api/entries/ID` endpoint(s) for individual entries. These always return full content. It would be excellent if the `detail` parameter worked here too.

Ideally, the second request below would return metadata only (no `content` data) like the first:

```sh
# /api/entries (GET)
# This already respects detail parameter
curl "$HOST/api/entries?access_token=$TOKEN&detail=metadata"

# /api/entries/ID (PATCH)
# This does not respect the detail parameter...but really should!
curl "$HOST/api/entries/$ID?access_token=$TOKEN&detail=metadata" \
     -X PATCH -d '{"archive": 1}' \
     -H 'Content-Type: application/json'
```
From related GitHub issues (#2585, #2817), it looks like when this feature was first being discussed, its usefulness in PATCH and POST requests on individual entries was recognized. But when the feature was added in #3960, it seems only GET requests to `/api/entries` received support.

Could the  `/api/entries/ID` endpoints support this too?

j0k3r:
Like it was done for the listing, add the parameter `detail` to the entry endpoint.

Remove the `ParamConverter` in the API `getEntryAction` so that we can actually only fetch what the client request (instead of fetching the `content` and then remove it later).

Fix #8901

### DIFF
--- a/src/Controller/Api/EntryRestController.php
+++ b/src/Controller/Api/EntryRestController.php
@@ -394,6 +394,17 @@ class EntryRestController extends WallabagRestController
      *             pattern="\w+",
      *         )
      *     ),
+     *     @OA\Parameter(
+     *         name="detail",
+     *         in="query",
+     *         description="include content field if 'full'. 'full' by default for backward compatibility.",
+     *         required=false,
+     *         @OA\Schema(
+     *             type="string",
+     *             enum={"metadata", "full"},
+     *             default="full"
+     *         )
+     *     ),
      *     @OA\Response(
      *         response="200",
      *         description="Returned when successful"
@@ -404,8 +415,18 @@ class EntryRestController extends WallabagRestController
      */
     #[Route(path: '/api/entries/{entry}.{_format}', name: 'api_get_entry', methods: ['GET'], defaults: ['_format' => 'json'])]
     #[IsGranted('VIEW', subject: 'entry')]
-    public function getEntryAction(Entry $entry)
+    public function getEntryAction(Request $request, Entry $entry)
     {
+        $detail = strtolower($request->query->get('detail', 'full'));
+
+        if (!\in_array($detail, ['full', 'metadata'], true)) {
+            throw new BadRequestHttpException('Detail "' . $detail . '" parameter is wrong, allowed: full or metadata');
+        }
+
+        if ('metadata' === $detail) {
+            $entry->setContent(null);
+        }
+
         return $this->sendResponse($entry);
     }
 
--- a/src/Entity/Entry.php
+++ b/src/Entity/Entry.php
@@ -430,7 +430,7 @@ class Entry
     /**
      * Set content.
      *
-     * @param string $content
+     * @param string|null $content
      *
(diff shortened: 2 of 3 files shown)


src/Controller/Api/EntryRestController.php
```diff
@@ -403,9 +414,22 @@ public function getEntriesAction(Request $request, EntryRepository $entryReposit
      * @return JsonResponse
      */
     #[Route(path: '/api/entries/{entry}.{_format}', name: 'api_get_entry', methods: ['GET'], defaults: ['_format' => 'json'])]
-    #[IsGranted('VIEW', subject: 'entry')]
```
yguedidi, line 406: This change looks wrong to me, I'd keep the IsGranted call so that security is checked first.
I think it should be possible to customize the used ParamConverter with the right attribute to use a specific function, the one you introduced, to fetch the entry. Maybe with the use of an Expression to be able to access the query parameter

yguedidi (changes requested):

src/Controller/Api/EntryRestController.php
j0k3r: Looks like `expr` doesn't have access to `request.query` (only to request attributes) so it's only doable with a custom `ParamConverter` service.

I'm balanced between the current working simple solution and the more engineered one with more files, but only to ensure security is checked first.

yguedidi: @j0k3r just curious, is there really a performance impact in fetching the full entry then not returning the content in the response? I'd guess no as most of the DB cost would be in finding the entry.
Then I'd question if we really need a custom entry fetching that will move the security check later.

To be creative, an alternative solution would be to introduce an `/api/entries/{entry}/metadata.{_format}` endpoint instead of the metadata boolean 😅

src/Repository/EntryRepository.php
```diff
@@ -281,7 +281,8 @@ public function countUntaggedEntriesByUser($userId)
      */
     public function findEntries($userId, $isArchived = null, $isStarred = null, $isPublic = null, $sort = 'created', $order = 'asc', $since = 0, $tags = '', $detail = 'full', $domainName = '', $isNotParsed = null, $httpStatus = null, $hasAnnotations = null)
     {
-        if (!\in_array(strtolower($detail), ['full', 'metadata'], true)) {
+        $detail = strtolower($detail);
+        if (!\in_array($detail, ['full', 'metadata'], true)) {
```
yguedidi: looks like a useless diff now

yguedidi (dismissed):
small change requested that can be fixed during the rebase

yguedidi (approved):

PR #8928, opened 2026-07-11
Import annotations from wallabag JSON exports

### LINKED ISSUE #8643: Importing a Wallabag export does not import annotations
### Environment

* **Version**: 2.6.14
* **Installation**: Docker
* **PHP version**: 
* **OS**:
* **Database**: Postgres
* **Parameters**:

<details>
  <summary>My <code>app/config/parameters.yml</code> is:</summary>

  ```yaml
parameters:
    database_driver: pdo_pgsql
    database_host: db
    database_port: 5432
    database_name: wallabag
    database_user: wallabag
    database_password: SNIP
    database_path: '%kernel.project_dir%/data/db/wallabag.sqlite'
    database_table_prefix: wallabag_
    database_socket: null
    database_charset: utf8
    domain_name: 'SNIP'
    server_name: 'Your wallabag instance'
    mailer_dsn: 'smtp://127.0.0.1'
    locale: en
    secret: SNIP
    twofactor_sender: no-reply@wallabag.org
    fosuser_registration: false
    fosuser_confirmation: true
    fos_oauth_server_access_token_lifetime: 3600
    fos_oauth_server_refresh_token_lifetime: 1209600
    from_email: wallabag@example.com
    rss_limit: 50
    rabbitmq_host: localhost
    rabbitmq_port: 5672
    rabbitmq_user: guest
    rabbitmq_password: guest
    rabbitmq_prefetch_count: 10
    redis_scheme: tcp
    redis_host: redis
    redis_port: 6379
    redis_path: null
    redis_password: null
    sentry_dsn: null
  ```
</details>

1. Annotate some articles on a Wallabag instance (I used wallabag.it)
2. Export all articles as json
3. On a fresh instance, import your JSON file.

Despite the file itself containing populated `annotations` objects, they are not present when being imported. 

It looks like the `annotations` key is not mentioned in the `WallabagV2Import`, `WallabagImport`, or `AbstractImport`, import classes, so I believe the data is just being ignored on import. Forgive me if I missed something, I'm not a PHP dev, otherwise I probably would have submitted a PR instead of a bug report.

TowyTowy:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Wallabag JSON exports include an `annotations` array for every entry, but the wallabag importers only read `url`, `title`, `tags`, etc. and silently dropped the annotations. Re-importing an export therefore lost every highlight/note.

This parses the `annotations` key in `WallabagImport::parseEntry()` and recreates each `Annotation` (text, quote, ranges) attached to the imported entry and the current user, following the same flow as `AnnotationController::postAnnotationAction()`. `Entry::$annotations` is now initialized in the constructor (like `$tags`) so a freshly created entry exposes a proper collection.

Added a unit test (`testImportWithAnnotations`) with a dedicated fixture (80 tests / 405 assertions pass across `tests/unit/Import` + `EntryTest`); the existing wallabag-v2 fixtures (annotations empty) are unaffected. php-cs-fixer clean.

Note: imported annotations currently receive an import-time `created_at` (the entity has no `createdAt` setter); happy to preserve original dates if preferred.

Fixes #8643

---
Disclosure: prepared with AI assistance (Claude); authored/verified by me. Tests and php-cs-fixer run locally and pass.

### DIFF
--- a/src/Import/WallabagImport.php
+++ b/src/Import/WallabagImport.php
@@ -2,6 +2,7 @@
 
 namespace Wallabag\Import;
 
+use Wallabag\Entity\Annotation;
 use Wallabag\Entity\Entry;
 
 abstract class WallabagImport extends AbstractImport
@@ -130,7 +131,37 @@ abstract class WallabagImport extends AbstractImport
         $this->em->persist($entry);
         ++$this->importedEntries;
 
+        if (\array_key_exists('annotations', $importedEntry) && \is_array($importedEntry['annotations'])) {
+            foreach ($importedEntry['annotations'] as $importedAnnotation) {
+                $annotation = $this->parseAnnotation($entry, $importedAnnotation);
+
+                if (null !== $annotation) {
+                    $this->em->persist($annotation);
+                }
+            }
+        }
+
         return $entry;
+    }
+
+    /**
+     * Create an annotation from an imported entry and attach it to the given entry.
+     *
+     * @param array $importedAnnotation Annotation data from the imported file
+     */
+    protected function parseAnnotation(Entry $entry, $importedAnnotation): ?Annotation
+    {
+        if (!\is_array($importedAnnotation) || empty($importedAnnotation['text'])) {
+            return null;
+        }
+
+        $annotation = new Annotation($this->user);
+        $annotation->setEntry($entry);
+        $annotation->setText($importedAnnotation['text']);
+        $annotation->setQuote($importedAnnotation['quote'] ?? '');
+        $annotation->setRanges(\is_array($importedAnnotation['ranges'] ?? null) ? $importedAnnotation['ranges'] : []);
+
+        return $annotation;
     }
 
     /**
--- a/src/Import/WallabagImport.php
+++ b/src/Import/WallabagImport.php
@@ -151,7 +151,7 @@ abstract class WallabagImport extends AbstractImport
      */
     protected function parseAnnotation(Entry $entry, $importedAnnotation): ?Annotation
     {
-        if (!\is_array($importedAnnotation) || empty($importedAnnotation['text'])) {
+        if (!\is_array($importedAnnotation) || !\array_key_exists('text', $importedAnnotation)) {
             return null;
         }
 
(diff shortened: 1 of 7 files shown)


TowyTowy:
Fixed the PHPStan failure: typing `Entry::$annotations` was needed once the constructor initializes it as an `ArrayCollection` (with `inferPrivatePropertyTypeFromConstructor`, the untyped property was inferred as an empty collection). Added `@var Collection<Annotation>` plus a `Collection` type on the property, mirroring `$tags`, and aligned `getAnnotations()`'s return docblock to `Collection<Annotation>`. `bin/phpstan analyse` and the import unit tests pass locally.

src/Import/WallabagImport.php
```diff
@@ -130,9 +131,39 @@ public function parseEntry(array $importedEntry)
...
     }
 
+    /**
+     * Create an annotation from an imported entry and attach it to the given entry.
+     *
+     * @param array $importedAnnotation Annotation data from the imported file
+     */
+    protected function parseAnnotation(Entry $entry, $importedAnnotation): ?Annotation
+    {
+        if (!\is_array($importedAnnotation) || empty($importedAnnotation['text'])) {
```
j0k3r: A note can be empty (if we only want to highlight some text)
```suggestion
        if (!\is_array($importedAnnotation) || !\array_key_exists('text', $importedAnnotation)) {
```

src/Import/WallabagImport.php
```diff
@@ -130,9 +131,39 @@ public function parseEntry(array $importedEntry)
...
+            return null;
+        }
+
+        $annotation = new Annotation($this->user);
+        $annotation->setEntry($entry);
+        $annotation->setText($importedAnnotation['text']);
+        $annotation->setQuote($importedAnnotation['quote'] ?? '');
+        $annotation->setRanges(\is_array($importedAnnotation['ranges'] ?? null) ? $importedAnnotation['ranges'] : []);
+
+        return $annotation;
```
j0k3r, line 172: Also, maybe you can force the created/updated at so it won't be lost after import

TowyTowy:
Good catch on the empty note, applied your suggestion.

For the dates: `EntityTimestampsTrait` was overwriting `updated_at` on every insert, so I split the callback — `prePersist` now only fills the dates when they aren't already set (it already did that for `created_at`), and `preUpdate` still always bumps `updated_at`. Nothing else in the codebase sets `updated_at` before persisting, so existing behaviour shouldn't change.

One thing I ran into: the annotation dates aren't in the JSON export at all — they have no serialization group, so `entries_for_user` filters them out and there was nothing to restore on import. I added the groups so the round-trip works, but that does change the export/API payload for annotations, so say the word if you'd rather have that as a separate PR and I'll pull it back out.

PR #8939, opened 2026-07-16
Migrate to ESM (type: module)

Rotzbua:
Convert the project to native ES modules by setting `"type": "module"` in package.json.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,7 @@
         "Routing": true
     },
     "rules": {
-        "import/no-extraneous-dependencies": ["error", {"devDependencies": true, "optionalDependencies": true, "peerDependencies": true}]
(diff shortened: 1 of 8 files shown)


yguedidi:
@Rotzbua why would this be needed? wallabag is an app, not a shared module

Rotzbua:
Setting `"type": "module"` in package.json is not about whether Wallabag is a shared module vs an app. It is how nodejs interpret the .js files: ES modules (ESM) or CommonJS.

Rotzbua:
esm is the new default. 
But this requires more dependency and code changes than expected.
Webpack, eslint and stylelint.... so just one migration with #8943

PR #8960, opened 2026-07-21
Add optional CAPTCHA protection to user creation

### LINKED ISSUE #8902: Add optional CAPTCHA protection to public registration and administrator user creation

A public wallabag installation can receive automated registration attempts. An administrator may also create users through the same application surface. Operators currently have to rely entirely on a reverse proxy, a WAF, or a custom patch when they want an application-level challenge. Those solutions are useful, but not consistently available and can be difficult to operate for small self-hosted instances.

The feature must not make a CAPTCHA service mandatory: many deployments are private, already protected, or deliberately avoid third-party trackers.

**Describe the solution you'd like**

Add an opt-in CAPTCHA integration based on the maintained `gregwar/captcha` package. It must be disabled by default, require no remote service, and be configurable through normal wallabag configuration.

When enabled, validate the challenge during public registration and when an administrator creates a user. A failed or expired challenge must leave the form usable, display a translated validation error, and never create a partial user. The implementation should preserve the existing flow when disabled.

**Describe alternatives you've considered**

Rate limiting, a reverse proxy, or a WAF remain valid deployment defences, but they cannot provide a portable application setting for every self-hosted installation. A mandatory hosted CAPTCHA provider would add privacy, network, and availability dependencies.

**Additional context**

Suggested acceptance criteria:

- default configuration changes no existing registration behaviour;
- CAPTCHA is required only when explicitly enabled;
- public and administrator creation forms are both covered;
- refresh/expiry and invalid-answer paths are tested;
- documentation explains configuration, accessibility considerations, and
  operational limitations.

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | yes
| CHANGELOG.md  | no
| License       | MIT

Fixes #8902

This adds optional, local CAPTCHA protection for public registration and administrator HTML user creation. It stays disabled by default through `WALLABAG_CAPTCHA_ENABLED=0`.

- Integrates `gregwar/captcha-bundle` with session-whitelisted, reloadable image challenges.
- Invalidates previous answers on refresh or submission and prevents partial user persistence.
- Keeps API registration unchanged and documents accessibility and abuse-prevention limitations.

This PR is easiest to review one commit at a time.


yguedidi:
Opened as an early preview, it still needs some UI improvements, see below

Image

Image

but an early quick review would be appreciable 🙂

yguedidi:
Fixed it by simply removing the field label "Security check", I think it's acceptable with our current UI.

image
image

j0k3r (approved):
The commit split is weird because you add the `CaptchaType` in a commit and then later you add the bundle 🤷‍♂️

yguedidi:
@j0k3r weird, because I see the first commit is the one introducing the bundle 😅

j0k3r:
Ok I read it in the wrong direction 🙃

yguedidi:
@j0k3r it happens haha 😅 
Thanks for the approval! Waiting for another approve from @nicosomb or @Kdecherf before merge

nicosomb (approved):

PR #8966, opened 2026-07-23
Add an opt-in hostname deny-list for outgoing fetches

### LINKED ISSUE #8904: Add an opt-in exact hostname deny-list for every content fetch path

An operator can have a legitimate policy reason to prevent wallabag from contacting a small set of domains: internal policy, known abusive endpoints, or an incompatible external service. Existing post-fetch origin handling cannot enforce this because it runs after a connection has already happened. Network proxies can help, but introduce external infrastructure and may not cover every HTTP client used by the application.

**Describe the solution you'd like**

Add an opt-in setting such as `WALLABAG_FETCH_BLOCKED_HOSTS`. It is empty by default and accepts a documented list of canonical hostname rules.

Normalize each hostname once. A rule without a leading dot uses exact matching: `example.com` blocks only `example.com`. A rule with a single leading dot explicitly covers a hostname subtree: `.example.com` blocks `example.com` and all its descendants, such as `www.example.com` and `deep.api.example.com`, without matching unrelated names such as `badexample.com`.

Do not support wildcards, regular expressions, arbitrary suffix matching, or implicit subdomain matching for exact rules.

Enforce this admission policy before a first request and after every redirect. Apply it consistently to entry fetching, refetching, imports, feed ingestion, email ingestion, background workers, and image downloads. A blocked request should provide a stable user-facing failure while logs retain only diagnostic data appropriate for the deployment's privacy expectations.

Implement the policy at a shared request boundary where possible, with explicit adapters or tests for clients that cannot use that boundary. Document hostname normalization and the behavior for exact rules, leading-dot subtree rules, ports, case, trailing dots, IP literals, and IDNs.

**Describe alternatives you've considered**

A proxy is deployment-specific and can be bypassed by a client not configured to use it. An origin-rewrite rule cannot prevent the original connection.

Wildcards, regular expressions, and implicit suffix matching are intentionally excluded because their security and user expectations are ambiguous. A leading-dot rule is supported as explicit, documented opt-in syntax for operators who intentionally want to deny both a hostname and its complete subtree.

**Additional context**

Suggested acceptance criteria:

- the default empty setting preserves current behavior;
- an exact normalized rule blocks only that hostname before any connection;
- an exact rule does not implicitly block subdomains;
- a leading-dot rule blocks both the named hostname and every descendant;
- a leading-dot rule does not block unrelated suffix lookalikes;
- redirects are checked at every hop;
- all content acquisition paths share the same semantics;
- tests cover case normalization, trailing dots, ports, IDNs, IP literals, redirect targets, exact rules, subtree rules, and blocked versus allowed subdomains.

This is an operator-controlled hostname policy, not complete SSRF mitigation.

yguedidi:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | yes
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Fixes #8904

This adds an opt-in `WALLABAG_FETCH_BLOCKED_HOSTS` admission policy for wallabag's outgoing HTTP fetches. An empty setting preserves current behavior; configured exact hostnames and leading-dot hostname subtrees are normalized and rejected before the first request and before every redirect. Denied requests are logged at the HTTP boundary with the rejected URL and matched hostname.

This is an operational egress policy and resource-use guard, not complete SSRF protection or a general security boundary. It prevents wallabag from spending connections, bandwidth, and worker time on known denied hostnames, but it does not resolve DNS names, classify address ranges, or prevent DNS rebinding. Deployments requiring SSRF isolation still need network-level egress controls.

## Request path coverage

A repository-wide audit found that all application-owned outbound HTTP fetch paths converge on the decorated Symfony client boundary:

- UI and API entry creation and reloads, CLI reload and import commands, regular imports, Redis workers, and RabbitMQ workers -> ContentProxy -> Graby -> PSR-18 bridge -> WallabagClient -> decorated shared client.
- Restricted-site login submissions -> HttpBrowser -> browser scoped client -> decorated shared client. Site-config `request_html` calls -> request_html scoped client -> decorated shared client. Article retries return through the protected WallabagClient.
- Image localization -> DownloadImages -> download_images scoped client -> decorated shared client.
- Pocket OAuth and import API calls -> pocket scoped client -> decorated shared client. Imported article fetching then rejoins the ContentProxy and Graby path.
- Default HttpClientInterface consumers and direct PSR-18 users -> aliases and adapters backed by the decorated shared client.

Compiled-container integration coverage resolves the default, image, browser, request_html, Pocket, Wallabag, and PSR-18 services and proves that each rejects a configured blocked host before transport dispatch. Redirect coverage proves that an allowed initial request cannot continue to a blocked target, including browser and image paths. ContentProxy keeps the configured fetching error when a fetch throws, so API entry creation remains successful and marks the entry unparsed; Pocket and restricted-authentication failure behavior also remains unchanged.

## Main review focus

The part that most needs explicit maintainer agreement is the HTTP client decorator, especially its redirect handling. Symfony exposes no public hook that can reject a redirect target before dispatch, so enforcing the policy at every hop requires disabling native redirects when the policy is active and reproducing their behavior through a per-response AsyncResponse callback.

The redirect state is isolated per response to preserve lazy execution, streaming, and parallel request consumption while retaining Symfony's method, body, header, credential, timeout, and redirect-limit behavior. This is deliberately more involved than the initial hostname check, but a synchronous-only alternative was almost as complex while failing to honor Symfony's asynchronous and parallelism contract. Please focus review on whether wallabag should own this Symfony-compatible redirect state machine and whether its compatibility coverage is sufficient.

Invalid rules fail during service creation.

This PR is easiest to review one commit at a time.


yguedidi:
@nicosomb @j0k3r @Kdecherf please check the big picture of this implementation 🙂 
I still need to do an in deep code and test review, and play with it for real, but I want to check first if you don't have any blocker regarding the implementation using a Symfony HttpClient decorator, even with the complexity of the redirect handling.

src/Helper/DownloadImages.php
j0k3r, line 1: Shouldn't this be done outside of that PR?

src/HttpClient/HostnameDenyListHttpClient.php
```diff
@@ -0,0 +1,201 @@
...
+        return $initialHost === parse_url($redirectUrl, \PHP_URL_HOST)
+            ? $headersWithCredentials
+            : $headersWithoutCredentials;
+    }
+
+    private function assertUrlIsAllowed(string $url): void
+    {
+        $hostname = parse_url($url, \PHP_URL_HOST);
+
+        if (!\is_string($hostname) || null === $hostname = $this->denyList->getBlockedHostname($hostname)) {
```
j0k3r: Why did you reassign `$hostname`? Maybe `$blockedHostname` is better here

```suggestion
        if (!\is_string($hostname) || null === $blockedHostname = $this->denyList->getBlockedHostname($hostname)) {
```

j0k3r (commented):
It looks like a machine gun designed to kill mosquitoes, but at least it will kill them.

I guess that the complex redirect / async mechanism is to handle every redirections used by the client (no matter it's for image or content given to Graby)?

yguedidi:
> I guess that the complex redirect / async mechanism is to handle every redirections used by the client (no matter it's for image or content given to Graby)?

@j0k3r yes! and also to align with Symfony expectations to preserve the async / parallelism feature

src/Helper/DownloadImages.php
yguedidi: I think the change is valuable because of the new introduced tests in https://github.com/wallabag/wallabag/pull/8966/changes/38ae979c8922fb89f36185cd653700c08724b516
so I'd say OK to keep it here

yguedidi:
@nicosomb @j0k3r @Kdecherf ready for a review, I reviewed it myself carefully, looks good to me! I also played with it locally and it seems to work fine.
I'll not be upset if you manage to find something I didn't found 😁 as it's quite an important security/performance/resource saving feature, better make it perfect than accepting only on my sole review 😉

Issue #8940, opened 2026-07-19
ePub export produces an invalid ePub if there's an ampersand in the article title


jtompkins:
### Environment

* **Version**: Latest docker image
* **Installation**: Coolify
* **OS**: Ubuntu 24.04.4 LTS
* **Database**: sqlite

### What steps will reproduce the bug?

1. Add an article with a bar ampersand (&) to your Wallabag instance. [Here's the article I'm using](https://seanguynes.com/2026/06/22/dnd-reading-streams-of-silver/).
2. Export that article as an ePUB.
3. Run `epubcheck` against the output:

```
FATAL(RSC-016): OEBPS/book.opf(-1,-1): Fatal Error while parsing file:
  The entity name must immediately follow the '&' in the entity reference.
FATAL(RSC-016): OEBPS/book.opf(10,23): Fatal Error while parsing file:
  The entity name must immediately follow the '&' in the entity reference.
ERROR(OPF-030): OEBPS/book.opf(-1,-1): The unique-identifier "BookId" was not found.

Check finished with errors
Messages: 2 fatals / 1 error / 0 warnings / 0 infos
```
I have also attached the invalid ePUB file from the URL above.

[invalid epub.epub.zip](https://github.com/user-attachments/files/30159880/invalid.epub.epub.zip)

PR #8968, opened 2026-07-24
Escape dynamic data in the ePub export

j0k3r:
To avoid having bad surprise.

I started to escape only title & url but the content can also have `&` or `<` and then break the html as well.

Fix https://github.com/wallabag/wallabag/issues/8940

### DIFF
--- a/src/Wallabag/CoreBundle/Helper/EntriesExport.php
+++ b/src/Wallabag/CoreBundle/Helper/EntriesExport.php
@@ -157,7 +157,7 @@ class EntriesExport
          * Book metadata
          */
 
-        $book->setTitle($this->title);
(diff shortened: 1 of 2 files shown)


nicosomb (approved):

Kdecherf:
@j0k3r shouldn't we handle this in PHPePub library instead?

yguedidi (approved):
Would have liked to use Twig here with an *.xml.twig template, but could be the scope for a separate refactoring

j0k3r:
> @j0k3r shouldn't we handle this in PHPePub library instead?

Maybe this can be done for title & url, but it's hard to handle the whole chapter escape.

Issue #8935, opened 2026-07-14
Entry created from Miniflux via API with full content has empty published_by


Crocmagnon:
### Environment

* **Version**: 2.6.14
* **Installation**: Docker image
* **PHP version**: Unknown
* **OS**: NixOS
* **Database**:  PostgreSQL 17.10 (Debian 17.10-1.pgdg12+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 12.2.0-14+deb12u1) 12.2.0, 64-bit
* **Configuration**: Likely irrelevant

### What steps will reproduce the bug?

* Configure the wallabag integration in Miniflux (leave "Send only URL instead of full content" unchecked)
* Subscribe to `https://dev.jimgrey.net/feed/` in Miniflux
* Save an article to wallabag via the Miniflux integration (e.g. with https://dev.jimgrey.net/2026/07/01/engineering-the-operating-model/)

Result in postgres:

Article saved manually via wallabag (OK)
```
wallabag=> select id, title, published_by from wallabag_entry where id=1666;
  id  |              title              | published_by 
------+---------------------------------+--------------
 1666 | Engineering the operating model | N;
(1 row)
```

Article saved via Miniflux with "Send only URL" checked (OK)
```
wallabag=> select id, title, published_by from wallabag_entry where id=1667;
  id  |              title              | published_by 
------+---------------------------------+--------------
 1667 | Engineering the operating model | N;
(1 row)
```

Article saved via Miniflux with "Send only URL" unchecked (NOK)
```
wallabag=> select id, title, published_by from wallabag_entry where id=1668;
  id  |              title              |   published_by    
------+---------------------------------+-------------------
 1668 | Engineering the operating model | a:1:{i:0;s:0:"";}
(1 row)
```

I'd expect to never have empty publisher names. These should be filtered out and if the remaining array is empty then store NULL.

Here's the relevant excerpt from Miniflux, where the payload is constructed:

https://github.com/miniflux/v2/blob/79d920bc1ac900834322fcfe13b6e228228c2fe0/internal/integration/wallabag/wallabag.go#L65-L74

PR #8970, opened 2026-07-24
Avoid empty array to be saved in `published_by` field

j0k3r:
Looks like a kind of empty `authors` given from the API can lead to an array of an empty value to be saved in the database. Also, the default value for the `authors` field in the API can generate that problem too.

Fix https://github.com/wallabag/wallabag/issues/8935

### DIFF
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -1406,7 +1406,7 @@ class EntryRestController extends WallabagRestController
             'language' => $request->request->get('language'),
             'picture' => $request->request->get('preview_picture'),
             'publishedAt' => $request->request->get('published_at'),
-            'authors' => $request->request->get('authors', ''),
+            'authors' => $request->request->get('authors', []),
             'origin_url' => $request->request->get('origin_url', ''),
         ];
     }
--- a/src/Wallabag/ApiBundle/Controller/EntryRestController.php
+++ b/src/Wallabag/ApiBundle/Controller/EntryRestController.php
@@ -1406,7 +1406,7 @@ class EntryRestController extends WallabagRestController
             'language' => $request->request->get('language'),
             'picture' => $request->request->get('preview_picture'),
             'publishedAt' => $request->request->get('published_at'),
-            'authors' => $request->request->get('authors', []),
+            'authors' => $request->request->get('authors'),
             'origin_url' => $request->request->get('origin_url', ''),
         ];
     }
--- a/src/Wallabag/CoreBundle/Entity/Entry.php
+++ b/src/Wallabag/CoreBundle/Entity/Entry.php
@@ -908,7 +908,7 @@ class Entry
     }
 
     /**
-     * @return array
+     * @return array|null
      */
     public function getPublishedBy()
     {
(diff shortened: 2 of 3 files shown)


nicosomb (dismissed):

src/Wallabag/ApiBundle/Controller/EntryRestController.php
```diff
@@ -1406,7 +1406,7 @@ private function retrieveValueFromRequest(Request $request)
             'language' => $request->request->get('language'),
             'picture' => $request->request->get('preview_picture'),
             'publishedAt' => $request->request->get('published_at'),
-            'authors' => $request->request->get('authors', ''),
+            'authors' => $request->request->get('authors', []),
```
yguedidi: shouldn't it be null as the property supports null?

src/Wallabag/CoreBundle/Entity/Entry.php
```diff
@@ -908,21 +908,31 @@ public function setPublishedAt(\DateTime $publishedAt)
...
-        $this->publishedBy = $publishedBy;
+        if (\is_array($publishedBy)) {
+            $publishedBy = array_values(array_filter(
+                array_map(
+                    static fn ($author) => is_scalar($author) ? trim((string) $author) : '',
+                    $publishedBy
+                ),
+                static fn ($author) => '' !== $author
+            ));
+        }
```
yguedidi, line 933: looks weird to e to have this logic in the entity, shouldn't it be done before we set the value?

yguedidi (changes requested):

yguedidi (changes requested):
shouldn't a migration be added to clean the DB from bad values?

src/Wallabag/CoreBundle/Entity/Entry.php
j0k3r: As `setPublishedBy` is used in 3 differents places, I thought it should be the best place

j0k3r:
> shouldn't a migration be added to clean the DB from bad values?

I target the 2.6 branch so I can't add a migration here.
Otherwise I must target the master branch.

yguedidi:
> I target the 2.6 branch so I can't add a migration here.

@j0k3r why it's not possible to add a migration on 2.6?

nicosomb:
> > I target the 2.6 branch so I can't add a migration here.
> 
> @j0k3r why it's not possible to add a migration on 2.6?

because it will be a patch version. Migrations will be on minor or major releases.

yguedidi:
> because it will be a patch version. Migrations will be on minor or major releases.

I'm not aware of this rule. why is not possible to have migrations on patch versions? I mean, a migration can not be a new feature nor break changes 🙂

src/Wallabag/CoreBundle/Entity/Entry.php
yguedidi: my point is more that none of the existing setters have a logic inside, so I'd preserve that, especially that they are used by Doctrine.
I'd suggest maybe to add a specialized setter, maybe something like `setRawPublishedAt(...)` or whatever you'll find better. that one then will not be tied to Doctrine, so it can have logic 😉

yguedidi (changes requested):
added few answers to comments

Kdecherf:
@yguedidi @j0k3r generally speaking we avoid pushing database migrations in patch versions because it may require restoring from a backup in case of an upgrade issue.

In this case it should be possible to provide a straightforward `down()` path for a rollback, maybe we can go ahead and include it.

PR #8992, opened 2026-08-11
ci: harden the release packaging path and the workflow actions

Totara-thib:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Documentation | no
| Translation   | no
| CHANGELOG.md  | no
| License       | MIT

Three commits, focused on the workflow that builds the tarball users download.

**Reference every action by commit hash.** The release packaging workflow uploads the tarball to the release with a write token, and all actions were referenced by tags. A tag is a movable pointer, whoever controls an action repository can point it at different code after review, which is how the tj-actions/changed-files incident (CVE-2025-30066) leaked CI secrets at scale, and that action is in use here. A commit hash cannot be retargeted. Each pin keeps the version as a comment for review, versions stay exactly where they were (composer-install sits on its 4.0.0 release since the v4 branch floats ahead of it), and dependabot keeps bumping hash pins the same way it bumps tags today.

**Build the release package without the yarn cache.** The packaging job restored a cache populated by earlier branch builds while producing the artifact that ships to users. Anything that poisons that cache reaches the release tarball. Release builds are rare, the extra install time costs less than the exposure.

**Declare token permissions.** Only the release packaging workflow keeps `contents: write` for the upload, the test, lint and git checks drop to read only.

One heads-up: if the org ever restricts allowed actions with tag patterns like `owner/action@v2`, those stop matching hash refs and workflows fail at startup, the fix is `owner/action@*` there. Nothing to do if no such restriction is configured.

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -19,10 +19,10 @@ jobs:
 
     steps:
       - name: "Checkout"
-        uses: "actions/checkout@v7"
+        uses: "actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1" # v7.0.1
 
       - name: "Install PHP"
-        uses: "shivammathur/setup-php@v2"
+        uses: "shivammathur/setup-php@f3e473d116dcccaddc5834248c87452386958240" # 2.37.2
         with:
           coverage: "none"
           php-version: "8.2"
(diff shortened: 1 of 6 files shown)


Kdecherf (approved):
Thanks @Totara-thib for your contribution 👍 

Note: I manually reviewed the hashes

Totara-thib:
> Thanks @Totara-thib for your contribution 👍
> 
> Note: I manually reviewed the hashes

Thanks for the merge, and extra thanks for verifying the hashes by hand, that is exactly the right reflex for a PR like this one.

If you want to keep the workflows from drifting back, the [Plumber](https://github.com/getplumber/plumber) workflow action runs these same checks continuously and can gate PRs. I work on it, so that is the disclosure. Happy to drop an example config here or open a small PR, and no hard feelings if you would rather leave it at the hardening

